PR #90, opened 2023-07-30
replace scoped style with node specific selectors

### LINKED ISSUE #86: dashboard breaks all nodes using an editor
When installed, any nodes using the RED.editor breaks

https://github.com/flowforge/flowforge-nr-dashboard/assets/44235289/a722b691-a36f-4bfe-8e38-2da632b94f68

---

Dashboard2 v0.0.4
Node-RED 3.1.0-beta.4
Chrome latest
Node 16

Steve-Mcl:
fixes #86

remove `scoped` from `<style>` and replace it with more specific specifiers to avoid affecting all `form-rows`

### DIFF
--- a/nodes/widgets/ui_button.js
+++ b/nodes/widgets/ui_button.js
@@ -14,10 +14,6 @@ module.exports = function(RED) {
                 var error = null
                 var event = msg.event
 
-                if (event) {
-                    console.log('event triggered by UI button click')
-                }
-
                 // retrieve the payload we're sending from this button
                 var payloadType = config.payloadType;
                 var payload = config.payload;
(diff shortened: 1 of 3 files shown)


joepavitt:
Does it make more sense to wrap all of the rows in a single div? Then had a single class to that parent div instead?

Or does Node-RED need individual form rows?

Steve-Mcl:
> Or does Node-RED need individual form rows?

I thought so (maybe they used to be?) but a quick check in the browsers dev tools tells me there is no reason we couldnt just wrap the `form-row`s in a `form-rows` div - who knew 🤷

Steve-Mcl:
@joepavitt wrapped in a div

Before:

image


After:

image


Still working:
chrome_PInyPhdTys


Pushed in

joepavitt (approved):

PR #93, opened 2023-07-30
Full Linting

Steve-Mcl:
## Description

* Adds linting to js, html and vue files (backend, node-red editor files and vue frontend linting)
* fixes 99% of lint issues - some imports and their ordering remains as was as this seems to break things
* fixes crash when doing full deploy &closing server (opportunistic bug fix)
* ensures server and socket.io callback before calling node.done()
* bumps version up to 0.0.6


nodes/config/ui_base.js
```diff
@@ -90,13 +122,12 @@ module.exports = function(RED) {
...
-     * @param {*} n 
+     * @param {*} n
      */
-    function UIBaseNode(n) {
-
+    function UIBaseNode (n) {
         const node = this
-
-        RED.nodes.createNode(node, n);
+        console.log('Creating UI Base Node', n)
```
Steve-Mcl, line 129:
```suggestion
```

remove temp debug logging

joepavitt (approved):

PR #97, opened 2023-08-02
Ensure items are closed and cleaned up

Steve-Mcl:
1. Ensure close is called for all types - not just widgets
2. inform each items parent of removal
   1. this ends up at the ui-base where de-registration occurs
3. close the socket.io server when there is nothing left to service

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -79,7 +79,6 @@ module.exports = function (RED) {
      * Close the SocketIO Server
      */
     function close (node, done) {
-        console.log('📈 ui-base: close')
 
         if (!ui.ioServer) {
             done()
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -36,7 +36,6 @@ module.exports = function (RED) {
      * Initialise the Express Server and SocketIO Server in Singleton Pattern
      */
     function init (node, config) {
-        console.log('📈 ui-base: init', config)
         // eventually check if we have routes used, so we can support multiple base UIs
         if (!ui.app) {
             ui.app = RED.httpNode || RED.httpAdmin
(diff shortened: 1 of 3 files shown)


nodes/config/ui_base.js
```diff
@@ -172,11 +194,13 @@ module.exports = function (RED) {
         }, 300)
 
         // Make sure we clean up after ourselves
-        node.on('close', async (done) => {
-            // clear event handlers on base node
-            ui.ioServer?.off('connection', onConnection)
```
joepavitt, line 177: why has this been removed? We want to remove the `ioServer` event handlers when a widget is deleted, which, if I can see correctly, you're no longer doing?

nodes/config/ui_base.js
```diff
@@ -36,6 +36,7 @@ module.exports = function (RED) {
      * Initialise the Express Server and SocketIO Server in Singleton Pattern
      */
     function init (node, config) {
+        console.log('📈 ui-base: init', config)
```
joepavitt, line 39:
```suggestion
```

nodes/config/ui_base.js
```diff
@@ -68,45 +69,64 @@ module.exports = function (RED) {
...
-        const httpServerClose = (_done) => {
-            if (ui.httpServer) {
-                ui.httpServer.close(() => {
-                    node.log('server shut down')
-                    _done()
-                })
-            } else {
-                _done()
-            }
+        console.log('📈 ui-base: close')
```
joepavitt, line 83:
```suggestion
```

nodes/config/ui_base.js
Steve-Mcl: as discussed - resolved in [e0e33b6](https://github.com/flowforge/flowforge-nr-dashboard/pull/97/commits/e0e33b68b474778909a503221a72150159a2e2f9)

joepavitt (approved):

PR #105, opened 2023-08-04
Event emitter improvements

### LINKED ISSUE #96: EventEmitter Memory Leak
### Description

```sh
(node:93636) MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
11 disconnect listeners added to [Socket]. Use emitter.setMaxListeners() to increase limit
```

### Epic/Story

_No response_

### LINKED ISSUE #102: widget-action and other events are increasing with each deploy
Due to the organisation of code, events like `widget-action` cannot be unhooked and as a result, each time a node is deployed, new events are hooked up causing duplicate, triplicate messages being posted to node-red

Steve-Mcl:
closes #96 
closes #102 

## Description

* uses common handlers for the 3 widget socket.io channels - instead of - an event hander per widget per handler
* ensures clean up socket io handlers on disconnect / close
* ensures emit config is requested when a node is added or removed in all deploy modes
* updates docs
* improved jsdoc type hints in ui-base
* removes the need for tracking load/change/action events
* only emits config to "own nodes connections"
   * prior, all connections (regardless of which base they belong to) were issued a config when any 1 base was updated. Now, only sockets connected to `this` base are emitted new config

## Related Issue(s)

#96 
#102 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -24,16 +24,13 @@ module.exports = function (RED) {
         /** @type { Server } */
         ioServer: null,
         /** @type {Object.<string, Socket>} */
-        connections: {},
-        events: {
-            load: {},
-            change: {},
-            action: {}
-        }
+        connections: {}
     }
 
-    /*
+    /**
      * Initialise the Express Server and SocketIO Server in Singleton Pattern
+     * @param {Object} node - Node-RED Node
+     * @param {Object} config - Node-RED Node Config
      */
     function init (node, config) {
         // eventually check if we have routes used, so we can support multiple base UIs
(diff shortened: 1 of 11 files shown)


nodes/config/ui_base.js
```diff
@@ -327,144 +517,37 @@ module.exports = function (RED) {
...
+                changes = true
+            }
+            if (changes) {
+                node.requestEmitConfig()
             }
         }
+
+        // Finally, queue up a config emit to the UI.
+        // NOTE: this is a cautionary measure only - typically the registration of nodes will queue up a config emit
+        // but in the event no widgets are registered, we still need still emit a config
```
Steve-Mcl, line 549:
```suggestion
        // but in cases where the dashboard has no widgets registered, we still need to emit a config
```

joepavitt:
In testing found a couple of issues. If you wire a text input into a text node:

Screenshot 2023-08-05 at 10 09 12

Then the full `msg` object is displayed, not just the payload as should be the default:

Screenshot 2023-08-05 at 10 09 25

This is odd however, because the Text widget does specify:

```vue
value: function () {
    return this.messages[this.id]?.payload
}
```

If we then refresh the view, the value in the text input, changes to an object, not the value input in the first place:

Screenshot 2023-08-05 at 10 10 32

joepavitt:
Seems like a dodgy nesting issue, this is the output if we add:

```js
console.log(this.messages[this.id])
```

to the computed `value` of the Text widget.

Screenshot 2023-08-05 at 10 13 06

joepavitt:
> Seems like a dodgy nesting issue

This should be fixed in https://github.com/flowforge/flowforge-nr-dashboard/pull/105/commits/89afb65f9a0f7779b39f165ec98e125df596a09a

joepavitt (approved):

Steve-Mcl:
Thanks for the catch Joe.

Issue #128, opened 2023-08-16
Radio Button Group


joepavitt:
### Description

Should be able to render a Radio Button Group as a standalone widget, _and_ as part of a `ui-form` widget.

### Properties

- Options

### Events

- **on-change**: this widget would need to send a message back to Node-RED when it's value is changed/updated

### Controls

- **enabled**: this widget can be enabled/disabled

### Existing Examples

https://vuetifyjs.com/en/components/radio-buttons/

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #131, opened 2023-08-16
Widget: Radio Group

joepavitt:
## Description

Adds a new `radio-group` node. This follows a similar pattern to the dropdown widget, in that it presents a table of options to be defined.

You can also control the number of `columns` to render the group across.

## Related Issue(s)

#128 

Note this does not include a radio group as an option inside the existing `ui-form` as proposed in #128 because the existing form definition table is fairly constrained to having singular values per "item" in the form, whereas we need to be able to define multiple options for a single component/entry in the form.

## Checklist


joepavitt:
Config in Node-RED:

Screenshot 2023-08-16 at 14 23 07

Renders:

Screenshot 2023-08-16 at 14 23 11

nodes/widgets/ui_radio_group.html
```diff
@@ -0,0 +1,161 @@
...
+            outputs: 1,
+            icon: 'font-awesome/fa-dot-circle-o',
+            paletteLabel: 'radio group',
+            label: function () { return this.name || (~this.label.indexOf('{' + '{') ? null : this.label) || 'radio group' },
+            labelStyle: function () { return this.name ? 'node_label_italic' : '' },
+            oneditprepare: function () {
+                if (this.multiple === undefined) {
+                    $('#node-input-multiple').prop('checked', false)
+                }
+                // $("#node-input-size").elementSizer({
```
ZJvandeWeg, line 50: ✂️

PR #139, opened 2023-08-17
Introduce publish package pipeline

ppawlowski:
## Description

Introduce package build and publish pipeline. At this stage, publishing is made only to the internal NPM registry.

## Related Issue(s)

https://github.com/flowforge/flowforge-nr-dashboard/issues/133

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Once this is merged we will need to also update the release action to publish full releases to the private repo as well

hardillb:
@joepavitt do we expect these tests to fail at the moment?

joepavitt:
Nope

ppawlowski:
> Once this is merged we will need to also update the release action to publish full releases to the private repo as well
> 
> https://github.com/flowforge/flowforge-nr-dashboard/blob/main/.github/workflows/publish-release.yml

Change available [here](https://github.com/flowforge/flowforge-nr-dashboard/pull/140).

joepavitt:
Not sure why this has triggered those tests to fail, need to investigate

joepavitt:
Tests fixed in #146 but I also would like to rename this a little better, as we do [already have](https://github.com/flowforge/flowforge-nr-dashboard/blob/main/.github/workflows/publish-release.yml) a `Publish & Release` workflow for publishing the to public repository.

If I understand correctly, a more appropriate name for this workflow would be `publish-ff-private` or something like that? Do we also need to update the `flowforge` usage here too, since the rebrand?

ppawlowski:
The name of the workflow has been updated.

> Do we also need to update the `flowforge` usage here too, since the rebrand?

`flowfroge` reference in the workflow is the name of the GitHub organisation. We will update it everywhere once we rename it.

hardillb:
Apart from that dashboard content appears to have snuck in here as well, is this good to merge now?

ppawlowski:
Maybe let's merge #146 first.

joepavitt:
My reasoning for wanting to split the workflows is because of this:

Screenshot 2023-08-29 at 12 34 29

We now have these workflows labelled accordingly running for every PR, but the labelling doesn't reflect what is actually happening.

Here, it's just running the `build and test`. In another instance, where we are pushing/merging to `main`, then that labelling is appropriate.

joepavitt (approved):

Issue #137, opened 2023-08-17
Configure Default Page


joepavitt:
### Description

When you have multiple pages defined, the router seems to just default to the page created the soonest. We should provide an option to define the "Default" page that would load when a user navigates to the `base` url, without specifying a page.

This option likely makes sense to live at the Dashboard 2.0 side menu (defined in `ui_base.html`), as by toggling on 'default' for one page, we need to affect all other pages, which I think is tricky to conduct from within a single widget's configuration menu.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Hitting circular dependency issues where I want the `ui-base` to provide a selection of a `ui-page` (to define the default page), however, `ui-base` is required by a `ui-page` in order to define where that `ui-page` exists.

gemini86:
This may not be the appropriate place for this type of comment, but it's my opinion that this feature should follow dashboard 1.0's implementation where the page order is determined by the order they listed in the sidebar, and the first parge is always loaded by default (/dashboard). Mentioning just in case this does get worked on in the future. The current work-around is to move all groups and widgets you want on the default page to the one that loads by default (in my case, the most recent page created) and then rename that page.

joepavitt:
Thanks @gemini86 - think you're presenting a very reasonable solution there. Prevents the circular dependency problem we have too.

PR #143, opened 2023-08-21
feat: add default path to group config

idolieb26:
## Description

Added default path on group

## Related Issue(s)

[#137](https://github.com/flowforge/flowforge-nr-dashboard/issues/137)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_group.html
+++ b/nodes/config/ui_group.html
@@ -6,11 +6,7 @@
                 value: 'Group Name',
                 required: true
             },
-            page: { type: 'ui-page', required: false },
-            path: {
-                value: '/',
-                required: false
-            },
+            page: { type: 'ui-page', required: true },
             width: { value: 6 },
             height: { value: 1 },
             order: { value: 0 },
@@ -38,10 +34,6 @@
     <div class="form-row">
         <label for="node-config-input-page"><i class="fa fa-bookmark"></i> Page</label>
         <input type="text" id="node-config-input-page">
-    </div>
-    <div class="form-row">
-        <label for="node-config-input-path"><i class="fa fa-bookmark"></i> Default Path</label>
-        <input type="text" id="node-config-input-path">
     </div>
     <div class="form-row">
         <label for="node-config-input-size"><i class="fa fa-arrows-h"></i> Size</label>
--- a/nodes/config/ui_group.html
+++ b/nodes/config/ui_group.html
@@ -6,7 +6,11 @@
                 value: 'Group Name',
                 required: true
             },
-            page: { type: 'ui-page', required: true },
+            page: { type: 'ui-page', required: false },
+            path: {
+                value: '/',
+                required: false
+            },
             width: { value: 6 },
             height: { value: 1 },
             order: { value: 0 },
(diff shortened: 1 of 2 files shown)


nodes/config/ui_group.html
```diff
@@ -6,7 +6,11 @@
                 value: 'Group Name',
                 required: true
             },
-            page: { type: 'ui-page', required: true },
+            page: { type: 'ui-page', required: false },
+            path: {
```
joepavitt, line 10: Thanks for the suggestion @idolieb26 - not sure I follow why we'd need a default path on a _group_ though. A single page contains many groups of widgets. All of them render at the same path (defined by the Page).

https://github.com/flowforge/flowforge-nr-dashboard/issues/137 is referring to adding this option at a `ui_base` level, and would allow you to select any existing `ui-page` elements.

nodes/config/ui_group.html
```diff
@@ -6,7 +6,11 @@
                 value: 'Group Name',
                 required: true
             },
-            page: { type: 'ui-page', required: true },
+            page: { type: 'ui-page', required: false },
```
joepavitt, line 9: Any group will require a `page` on which to live, so we need to have this as `required: true`

nodes/config/ui_group.html
idolieb26: @joepavitt But there is already a `path` field in `ui-base`. So  I just need to select any existing `ui-page` elemensts, right?

joepavitt: That's correct, so a default page being added to a `ui-base` would be the way to go. We have a four-tier hierarchy:

- Base: defines the `base` URL, e.g. `/dashboard/`
- Page: individual tabs/pages, all hosted at the same base, will extend the base URL, e.g. `/dashboard/page1`
- Group: All exist within a given page
- Widget: All exist within a given group

So if someone navigates to `/dashboard` or any subpath that isn't a valid page, we need to redirect them to _somewhere_. That somewhere is what we need to define as the `default`.

Two options exist for where this could live, at the `ui_base` level, in it's own configuration menu, or we have a global "Dashboard 2.0" side menu which controls global config options, where it could also live. This side menu is a WIP though in the branch I'm working on for https://github.com/flowforge/flowforge-nr-dashboard/issues/136

idolieb26: @joepavitt I added `default` to `ui-base`, but there is no way to implement redirect logic in our dashboard, I guess.

joepavitt: We have a VueJS app for the Dashboard. Node-RED is sending websocket events to the VueJS app. That app has a `router.js` which is empty by default, but then populated when we receive a `ui-config` socket event in `App.vue` which details the pages/routes.

idolieb26: @joepavitt Yeah, that is what I guessed so. But the VueJS app is not in the `flowforge-nr-dashboard` project, right?

joepavitt: Indeed it is - under `/ui`

joepavitt: If you're running this locally, you'll need to run an `npm run build` from the cmd line at the project level, that builds an instance of the VueJS app, which Node-RED can then pick up. 

It's a pain to have to do a manual build step after each change, it's on our todo list to sort to also work with a dev server.

Issue #76, opened 2023-07-28
Responsive Data Tables


joepavitt:
### Description

Proposed here: https://github.com/flowforge/flowforge-nr-dashboard/issues/16#issuecomment-1608288550

### Properties

- columns/rows
- Style should be responsive and switch to "card" views on mobile/narrow layouts

### Events

- **on-row-click**: this widget could have an action triggered when selecting a given row of data?

### Existing Examples

Vuetify table: https://vuetifyjs.com/en/components/tables/

This does not include an auto-wrapping to a "card" view as requested though.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

thebaldgeek:
I think the responsiveness of the table gadget has regressed in v 0.10.2?

I now have to stretch my browser over the full width of a 4k + 2k monitor to see the last few columns (in the red square).
The lack of a horizontal scroll bar makes it impossible to see the data on anything less than 6k of screen.

I chose to switch to a monospace font (Lucinda Console) with the v1 table to help the readability. I see that the v2 table does not have the option to set the font type at all?

Being able to set the column types is critical, link, html etc (and I see a request in for that), but having only about 5% responsiveness is making it tricky to use. (It does expand and contract around the 5k monitor size, but no less).

![image](https://github.com/FlowFuse/node-red-dashboard/assets/7385116/fc65632c-3d19-4854-9275-c288dc64db09)

Here is the exact same table and data on the v1 dash on the same center monitor and Chrome browser.

![image](https://github.com/FlowFuse/node-red-dashboard/assets/7385116/2583e85e-f5e8-4ae6-8ee3-53bbc827b348)

joepavitt:
New issue opened to track that particular issue, thanks @thebaldgeek

joepavitt:
Challenge is is how to detect when to go into responsive mode. It's difficult to judge based on overflow/overlap to then switch, as no JS events are fired in that occasion, and I don't really want to be watching every single cell of a table for overflow manually.

The "easy" here is a screen width option, but similar, if it's just two columns on a mobile, then, in theory, that's fine to render as a table?

I'll move forward with building out the card/mobile view now, but the logic for when to switch to that view needs to be refined/decided - @thebaldgeek thoughts very welcome here.

joepavitt:
[...]

PR #168, opened 2023-09-02
Widget: Data Table

joepavitt:
## Description

Precedes #76 - realise I hadn't added the table widget at all.

- Will display a holding "No Data" message until it receives a payload.
- On receipt of a `msg`:
    - If "Auto Calculate Columns" is active, it will read the keys of the data objects in the received array and auto create columns based on _all_ keys found in the data.
    - If it isn't active, then it will use any explicitly defined "Columns" in the column-editor within Node-RED, where each column requires a Label & Key from which to read in the object.
- Adds pagination option with a "Max Rows" option

## Checklist


joepavitt:
https://github.com/flowforge/flowforge-nr-dashboard/assets/99246719/5f6973cf-8455-4271-8b56-e462a34cf10e

joepavitt:
Also know that @Steve-Mcl would advocate here for a "rows" value on the widget that is a typed input, defaulting to `msg.payload`, but that allows us to easily define which input property for a message we use for the rows, likely a nice +1 feature after this one though as it's not essential.

joepavitt:
Could also add alignment (left/center/right) for each column as a future feature too.

joepavitt:
Would also appreciate @robmarcer's opinion here as I noticed you'd be rendering tables a few times with Dashboard 1.0 (and I assume using `ui-template`?) Any opinions around structure/UX are most welcome

Steve-Mcl:
> and I assume using `ui-template`?

Possibly, or the UI table node [`node-red-node-ui-table`](https://flows.nodered.org/node/node-red-node-ui-table) which is essentially a wrapper around the extremely capable/flexible **tabulator**

robmarcer:
I've been using https://flows.nodered.org/node/node-red-node-ui-table

Having the ability to make the table continue on until the end of the data set would be really useful, so a user can use their browser's main window to scroll. You'd ideally still be able to put content beneath the table. If that's possible in node-red-node-ui-table I have not yet worked out how to do it.

I could spend a few hours digging into the features I've not used to give more detailed ideas, I suspect there's lots I've not even looked at yet. 

Some features I would see value in:

- Easy to define 'rollups' (total, min. max etc).
- Easy to add URLs to link to other locations in the dashboard to create drill-downs.
- Ability to crud individual lines of the data rather than replace the whole set each time something needs to change.

I'm pretty sure node-red-node-ui-table does all of this with the exception of allowing the table length to extend the dashboard to show all data.

joepavitt:
Fun demo linking the table to an API and having a button to generate 10 random jokes:

https://github.com/flowforge/flowforge-nr-dashboard/assets/99246719/1719dbfc-c529-463a-bbd3-3e8e8d2e3fc9

PR #191, opened 2023-09-12
Add support for settings-provided middleware

### LINKED ISSUE #190: Add support for custom http middleware to be added in front of all dashboard routes
### Description

The existing dashboard allows a custom middleware to be added in front of the dashboard via the `ui.middleware` setting.

We need a similar mechanism in the new dashboard.

The new dashboard does not currently make use of any runtime settings.

For backwards compatibility reasons, I propose an approach that

1. Uses Dashboard2.0 settings object if present
2. Otherwise, fallback to Dashboard1.0 settings object

In other words, if `dashboard` is present in the settings file, use that. Otherwise, look for the `ui` settings object and use that.

Initially, this should only apply to the `middleware` setting. The other setting, `path` can be used to customise the path the dashboard is served from. If you have both 1.0 and 2.0 installed, then you don't want them sharing a path. We can figure out ways to customise the path in a future iteration.

knolleary:
Closes #190 

## Description

This adds support for providing custom middleware in front of the dashboard (and its websocket) routes.

By default it looks for:
  - `dashboard.middleware`
  - `dashboard.ioMiddleware`

If the top-level `dashboard` settings object is not present, it will look for the existing dashboard settings object - `ui.*`.

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -31,6 +31,7 @@ export default {
           text: 'User Guides',
           collapsed: false,
           items: [
+            { text: 'Settings', link: '/user/settings' },
             { text: 'Dynamic Properties', link: '/user/dynamic-properties' }
           ]
         },
--- a/docs/user/settings.md
+++ b/docs/user/settings.md
@@ -1,0 +1,78 @@
+# Settings
+
+Whilst most dashboard configuration is done within the Node-RED editor, some settings
+can be provided in the [Node-RED runtime settings file](https://nodered.org/docs/user-guide/runtime/settings-file).
+
+The `settings.js` file is typically located in the `~/.node-red` directory. If you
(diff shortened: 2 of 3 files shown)


joepavitt:
Thanks Nick, could I trouble you to add some content into `/docs/user`, maybe a new `settings.md` or `authentication.md` and include a basic example of what this settings file/content would look like please? I know we know for FF, but want to make sure general docs are up to date too.

Steve-Mcl:
great stuff. If by any chance this is released today/tomorrow, I can do that write up after all (and include some of the new goodies we added)

knolleary:
Working on the docs now - worth adding 'requires changes' review as that'll cut through my notifications a bit better :)

knolleary:
@joepavitt docs added

joepavitt (approved):

PR #192, opened 2023-09-12
Widget: UI Notification

### LINKED ISSUE #149: Add the Notification node (ui-toast) that is present in Dashboard 1
### Description

D1 includes a Notification node (ui_toast). It appears to be missing from D2

I am no FlowForge team member

joepavitt:
## Description

ui-notification

Add `ui-notification` which is a mirror of `ui-toast` from Dashboard 1.0.

Some decisions made that differ from it's implementation in D1.0:
- Refinement of the `layout` property to `position`
    - Additionally, I've removed the Dialog options here, `position` should be just that, it felt confusing that "Dialog" variants were options here. Instead, we will develop an additional `ui-dialog` widget.
- Added a progress bar that displays time left until the notification will close (can be turned off)
- Optional "Close/Clear" button, can be configured to be hidden and to control what text displays. 

## Related Issue(s)

Closes #149 

## Checklist


flinthamm:
First off, thank you for this amazing project and I look forward to ongoing developments. 
Sorry to ask such a trivial question but could someone give an example of how to display a ui-notification as depicted in the screenshot i.e. with a bold title and message directly underneath? I found examples of dashboard 1, where it appeared sending a msg.topic would set the title and the msg.payload the message but this did not work (no title, just message). I've tried sending basic html formatting e.g.
```
<p><string>Generated Notification</strong><br />This is custom HTML...</p>
<p><string>Generated Notification</strong></p><p>This is custom HTML...</p>
```
Only the title is displayed but perhaps I need to introduce inline CSS?

joepavitt:
@flinthamm you will need to check the "Accept raw HTML/JavaScript input in msg.payload to format popup."

flinthamm:
Thank you and apologies @joepavitt, I should have made this clearer, that this option has already been checked. It does display the bolded text (not raw html), just not the message afterwards, presumably an HTML or CSS issue, so it would be great to see an example of the data behind the example toast that was given? I assume only the msg.payload data is used and the msg.topic or similar cannot be used to set a title as this is not mentioned in the docs?

joepavitt:
Ah, understood - yes, that should be made clearer. The boldened text is achieved with a `<h3></h3>` in the `msg.payload, e.g:

```html
<h3>Test Notification</h3><p>Some content</p>
```

in your `msg.payload` will do the trick

joepavitt:
Example added: https://dashboard.flowfuse.com/nodes/widgets/ui-notification.html#example

flinthamm:
Wow @joepavitt amazing that worked and stupid really, I should have thought to try headings! We live and learn and who knows, might help someone else. Look forward to the ongoing dev. of this brilliant project and thanks again.

PR #268, opened 2023-10-12
Injection of x/y-AxisProperty into ChartJS config from NR config

### LINKED ISSUE #230: Expose ChartJS "parsing" option to define nested data values for x and y axes
### Description

ChartsJS has a `parsing` [option](https://www.chartjs.org/docs/latest/general/data-structures.html#object-using-custom-properties) that, if we expose would meant that Node-RED users no longer need to transform their data to `msg.payload.x/y` format, and instead, can pass their raw data into the chart directly.

e.g.

```js
onst cfg = {
  type: 'bar',
  data: {
    datasets: [{
      data: [{id: 'Sales', nested: {value: 1500}}, {id: 'Purchases', nested: {value: 500}}]
    }]
  },
  options: {
    parsing: {
      xAxisKey: 'id',
      yAxisKey: 'nested.value'
    }
  }
}
```

joepavitt:
## Description

https://github.com/FlowFuse/node-red-dashboard/assets/99246719/822d72f0-56c7-452e-aee9-c86a72513dd6

## Related Issue(s)

Closes #230 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
All of these examples now show scatter/line/bar charts all being driven by different input types (e.g. single numbers, objects and arrays of data), as well as a combination of how `category` (series for a line/scatter plot) is defined (sometimes `msg.topic`, sometimes a value in the data set itself, sometimes a fixed string for the whole Chart:

https://github.com/FlowFuse/node-red-dashboard/assets/99246719/93f0cecb-2d4c-4e4a-b3bd-8133d853752f

MarianRaphael:
Overall, it works!

Minor hints and comments:
The label "property" was a little bit confusing, I am used to fields with the label (and functionality) "msg." In the "Series" field you have the selection between both; for the x/y values it's fixed.
Screenshot 2023-10-19 at 07 39 02

Sometimes while opening the Dashboard and Error occurs (could be just my local installation):

Screenshot 2023-10-19 at 08 10 22

MarianRaphael (approved):

joepavitt:
Thanks Marian.

Regarding the "property", having `msg.` here would be inaccurate, as its looking for the property/key/field on an object in `msg.payload` or `msg.payload[n]`

I could just leave them as standard string input, and not have the label? But want to try and hint to use that it _is_ different to the other input box

joepavitt:
The reason for going typed inputs, with type `property` was because, for the Series, you can use the `msg.payload[n].<series>` or `msg.<series>`

joepavitt:
Maybe `key` would be a better label?

MarianRaphael:
> Maybe key would be a better label?

Yes

Issue #229, opened 2023-10-03
Migration: Chart Input/Output Formats


MarianRaphael:
One issue during the migration of one of my dashboards was that the output and expected input from Dashboard 2.0's chart differ. Previously, it was possible to store the output of the chart and insert the stored values upon instance refresh. This is now different, which makes it hard to create a persistent dashboard. An easier migration would be possible if the old chart input format were also accepted.

## Requirements

- [ ] Dashboard 2.0 should accept the old chart input format alongside the new one.
- [ ] Support for Year-Month-Date on X-axis
- [ ] Output of the chart node should be compatible with input

joepavitt:
Not sure I know the problem here? Can you be more specific please?

They should be identical?

joepavitt:
> Output of the chart node should be compatible with input

Just checked this:

<img width="553" alt="Screenshot 2023-10-03 at 10 18 30" src="https://github.com/FlowFuse/node-red-dashboard/assets/99246719/0e1b5baa-8775-4075-a429-0706f84326c9">

Output is identical to input:

<img width="324" alt="Screenshot 2023-10-03 at 10 18 41" src="https://github.com/FlowFuse/node-red-dashboard/assets/99246719/7313307d-15a4-4bfd-9176-d515a401bf53">

joepavitt:
> Support for Year-Month-Date on X-axis

This is also supported already:

<img width="971" alt="Screenshot 2023-10-03 at 10 22 10" src="https://github.com/FlowFuse/node-red-dashboard/assets/99246719/6bb87081-06b1-453d-bdb7-c817f84a8760">

<img width="542" alt="Screenshot 2023-10-03 at 10 22 26" src="https://github.com/FlowFuse/node-red-dashboard/assets/99246719/ddb072a7-cd7f-4e8c-ac75-543f7ca360c5">

<img width="545" alt="Screenshot 2023-10-03 at 10 22 20" src="https://github.com/FlowFuse/node-red-dashboard/assets/99246719/e9907917-5be3-4bbc-9a6d-f5779f99384d">

MarianRaphael:
Here is the example, a new dashboard with old dashboard data: 
https://product-metrics-dev.flowforge.cloud/?#flow/927154a090f3f116
![Screenshot 2023-10-03 at 11 45 18](https://github.com/FlowFuse/node-red-dashboard/assets/73583313/2e333ce3-4268-4b42-a411-5ca23a2a609f)

[...]

Issue #212, opened 2023-09-21
Add "migration" guide from D1.0 > D2.0


joepavitt:
### Description

We'll have (hopefully) a lot of users transitioning from Dashboard 1.0 to 2.0, and should have an explicit page on the documentation to detail where some things have moved from/to.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
First draft pushed in #283 

Closing for now and further work will be done with individual node/widget updates.

PR #283, opened 2023-10-20
D1.0 > 2.0 Migration Guide

joepavitt:
## Description

First draft of the Migration Guide, it's fairly bland and detailed, but wanted to get the ball rolling, and we can improve on it over time.

The main missing piece at the moment is documentation of migrating events (e.g. injection of `msg.<something>` into a widget) in order to control behaviour, and how that may differ in some cases (e.g. `ui-chart` doesn't support the standard data format for Dashboard 1.0's `ui_chart` - https://github.com/FlowFuse/node-red-dashboard/issues/229).

## Related Issue(s)

#212


joepavitt:
Will merge as is, put it into production docs, and can then improve based on feedback after that.

PR #298, opened 2023-10-24
Update to new npm org

knolleary:
## Description

Update the package name to the new `@flowfuse/node-red-dashboard` org.

Once this has been merged, I will manually publish 0.7.0 to the new package name. From then on, new releases can be created in the usual way and will be published to the new org.


joepavitt (approved):

joepavitt:
@knolleary all good to go

PR #321, opened 2023-11-07
Widget: UI Event

### LINKED ISSUE #274: Widget: UI Event
### Description

It has been discussed on many occasions that it would be beneficial to have the context of `user` data available in Node-RED when a Dashboard is rendered, i.e. if a page is rendered, having context on which user is viewing that page in both NR and Dashboard.

Given that we do have FF & basic auth available to Dashboard through the NR middleware, this is data we should be able to have an expose.

Similarly, we've had a request whereby we want to have different data show for different users, as you would expect on any standard software product when a user logs in.

Underpinning this, I believe there is an architecture within the flows that could be accomplished with a `ui-event` node. This node would send a payload whenever a set of given events takes place, and could include things like `pageview`, with accompanying data included in it's transmitted payload that details the authenticated user.

The most significant complexity for the above use cases is server-side state management (SSSM), which would essentially need to be turned off, and moved to a Database (most likely user managed, rather than something we offer as a service with Dashboard). With SSSM till on, a user viewing a page would trigger a database lookup, sending that data to the chart, and be made visible to the given user, but because that data would be retained server-side, any user viewing that dashboard would now see that user's data.

Instead, we would need to look at SSSM on a connection-by-connection basis - which could get very messy, or no SSSR at all, which feels cleaner given that state would need to be database managed anyway.

So a side-issue of this implementation may need to be the ability to turn off SSSR for a given widget/group/page/ui

joepavitt:
## Description

- Adds new "Event" widget
- The widget tracks `$pageview` and `$pageleave` events.
- Scope to grow these events is fairly easy too, just pushing in this as the foundation of "first" events.

## Related Issue(s)

Closes #274 

## Checklist

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -83,6 +83,7 @@ export default ({ mode }) => {
                 { text: 'ui-button', link: '/nodes/widgets/ui-button' },
                 { text: 'ui-chart', link: '/nodes/widgets/ui-chart' },
                 { text: 'ui-dropdown', link: '/nodes/widgets/ui-dropdown' },
+                { text: 'ui-event', link: '/nodes/widgets/ui-event' },
                 { text: 'ui-form', link: '/nodes/widgets/ui-form' },
                 { text: 'ui-markdown', link: '/nodes/widgets/ui-markdown' },
                 { text: 'ui-notification', link: '/nodes/widgets/ui-notification' },
--- a/docs/components/EventsList.vue
+++ b/docs/components/EventsList.vue
@@ -1,0 +1,30 @@
+<template>
+    <table style="width: 100%;">
+        <thead>
+            <tr>
+                <th>Event</th>
+                <th>Payload</th>
+                <th>Description</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr v-for="e in page.frontmatter?.events" :key="e.event">
+                <td><code>{{ e.event }}</code></td>
+                <td><code>{{ e.payload }}</code></td>
+                <td v-html="e.description"></td>
+            </tr>
+        </tbody>
+    </table>
+</template>
+
+<script setup>
+import { useData } from 'vitepress'
+
+const { page } = useData()
+</script>
+
+<script>
+export default {
+    name: 'EventsListTable'
+}
+</script>
--- a/docs/components/WidgetCard.vue
+++ b/docs/components/WidgetCard.vue
@@ -1,6 +1,7 @@
 <template>
     <a class="widget-card" :href="url">
         <img
+            v-if="widget.image"
             :src="widget.image"
             :alt="widget.description"
         >
--- a/docs/nodes/widgets.md
+++ b/docs/nodes/widgets.md
@@ -74,10 +74,17 @@
         description: 'Adds a clickable button to your dashboard.'
     }]
 
+    const events = [{
+        name: 'Event',
+        widget: 'ui-event',
+        description: 'Monitors for events in the Dashboard and emits accordingly.'
+    }]
+
     const widgets = ref({
         general,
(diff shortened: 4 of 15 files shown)


joepavitt:
Issues opened to capture future ideas:
- https://github.com/FlowFuse/node-red-dashboard/issues/323

MarianRaphael:
Hi @joepavitt on October 27, we discussed UI Event and UI Control and decided to include the functionality in UI Control as in D1.0. What has changed since then?

joepavitt:
That is a conversation I forgot that we had. 

I dove into `ui-control` a bit more in D1.0, and it only emits socket connection events, no user interaction type events.

Architecturally, the only justification I can think of for keeping them the same is that Dashboard 1.0 does it. Whilst, for migration, that would make things easier, we are in a position where we should be making moves to clean architecture too.

Having a separate `ui-control` that _controls_ and a `ui-event` which monitors foe activity and events on the Dashboard is a cleaner separation imo. Having a `ui-control` that emits events too feels unintuitive given its naming

MarianRaphael:
Yes, it is exactly about the discussion you describe. Nick noted that UI control already offers some functionality in this regard. The original debate centered on whether to use CE or EE for UI events. We discussed all the arguments you described and came to the decision to include it in ui-control as in D1.0, if your analysis shows that the functionalities proposed for CE are possible.

In my view, there has been no change in substance since that discussion. UI socket events should serve their intended purpose for the functionalities available in the CE.

MarianRaphael:
@joepavitt  I think this will evolve into a longer discussion thread here. I would suggest reverting the merge and proposing a meeting when I return.

joepavitt:
> The original debate centered on whether to use CE or EE for UI events

We concluded that `ui-events` and `ui-control` were both CE. 

As I discussed on that call, I am _categorically_ against `ui-event` being an EE node only. It is incredibly easy just to reproduce such a node in a template or third party node, and is not sufficient enough value/IP to warrant an EE feature. The "EE" value we had discussed was around the inclusion of `user` data with these events, which would only be possible with SSO/Auth layer as middleware in Node-RED. That, I agree with.

I want to emphasise that what `ui-event` is doing as a result of this PR does not overlap with _any_ functionality that `ui-control` offers. 

`ui-control` in D1.0 offers the following events:

- socket connected
- socket disconnected

`ui-event` here in D2.0, so far, includes the following events:

- page view
- page leave

They are different events. Now, my plan _was_ to migrate over the socket events into `ui-event` too, but as a backward compatibility/migration middle ground, we could support. _both_. `ui-control` can still emit these events _exactly_ as they are done in Dashboard 1.0, in order to support migrations. We can then also introduce similar events into `ui-event` that use our new data structure to be consistent with other events being emitted from this node.

MarianRaphael:
Just for clarification in preparation for the meeting: In D1.0, the UI control reports any "view change" and also every page view. See my  screenshot. Why is there no overlap?
Screenshot 2023-11-08 at 10 28 05

joepavitt:
Ah, missed that, that is not in the documentation

Steve-Mcl:
FWIW, I personally agree with the direction this PR takes DB2. 

Basic functionality should definitely be CE even if that is something new DB2 has over DB1. If not, I fear we risk alienating folk and foster ill feeling or worse, bad press.

As for control/events: the cleaner separation of control and events is definitely a worthwhile change IMO.  Dashboard 1 design is terribly unintuitive and inflexible in places. I feel we should not be blindly duplicating functionality when it doesn't make sense.

If there is a vote, please count this statement towards the current direction Joe has taken. (ignore otherwise 😉)

joepavitt:
Marian and I have just spoken - agreement in separation of `ui-event` and `ui-control`

PR #325, opened 2023-11-08
Revert "Widget: UI Event"

MarianRaphael:
Reverts FlowFuse/node-red-dashboard#321

I think this will evolve into a longer discussion thread here. I would suggest reverting the merge and proposing a meeting when I return.


joepavitt:
Thread of discussion taking place here: https://github.com/FlowFuse/node-red-dashboard/pull/321#issuecomment-1801353575

joepavitt (approved):

PR #336, opened 2023-11-13
Convert project to build ES Modules using Vite

Pezmc:
## Description

As part of the preparation for supporting third party widgets (https://github.com/FlowFuse/node-red-dashboard/issues/307), this PR re-works Node-Red Dashboard 2.0 to be built using Vite and rollup instead of Webpack.

The third party widgets will also be expected to be built as ES Modules with UMD enabled and we'll be suggesting Vite is an easy low configuration way for other developers to achieve this in.

As an aside, Vite also gives notable speed improvements and significantly simplifies the asset pipeline.

One drawback, Node-Red does not support loading nodes using the file extension `.cjs` yet (see https://github.com/node-red/node-red/issues/4431 and https://github.com/node-red/node-red/pull/4433), so, in the short term, all dashboard UI files must use the `mjs` file extension so they are treated as modules and this project will continue to be configured as CommonJS.

I also took a moment to clean up a few linting disables, which hid real errors from being caught.

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/307

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Backport needed? -> add the `backport` label
 - [ ] Includes a DB migration? -> add the `area:migration` label

Issue #307, opened 2023-10-26
Why difference between core and custom ui nodes?


bartbutenaers:
### Description

Hi @joepavitt,

The old dashboard originally contained a limited set of core ui nodes.  A few years later, fortunately there arrived a mechanism to contribute custom ui nodes.  That was a great step forward.  But the setup of the custom ui nodes was very different from the core nodes.

The new dashboard works very similar.  There is a set of core ui nodes, and the setup of the custom ui nodes is different.  Or at least for me (as a VueJs noob) they look different.

I was wondering why not all ui nodes are implemented the same way, i.e. like custom ui nodes.  The only difference would be that the core nodes would be installed out of the box, in contradiction to the custom nodes which still need to be installed seperately.  Although it would be ok for me if I had to install all core nodes seperately via the palette.  But that is another discussion...

Because if the core ui nodes would have the same setup as the custom ui nodes, that would have some advantages for me as a ui node developer:
1. The contribution documentation for core nodes and custom nodes would be identical.
2. It would be easy to look at the code of core nodes to learn from it, in order to be able to create my own custom node.  Now when I want to get started with this, it is very hard to determine if I can reuse some core stuff or not.
3. A custom node could become easily afterwards a core node, if there would be some need for that.  E.g. for maintenance issues or whatever...

Perhaps this technically very difficult, or it has some disadvtages I am not aware of.  Don't know.  Just thinking out loud...

Thanks for reading!
Bart

### Have you provided an initial effort estimate for this issue?

I am no FlowFuse team member

joepavitt:
This is absolutely what I wish to strive for, as a first iteration I've copied the pattern from Dashboard 1.0, where all 3rd party widgets are integrated into the client-side Dashboard via `ui-template`.

It all essentially boils down to needing to have a better way of completing this line: https://github.com/FlowFuse/node-red-dashboard/blob/89fb0644baab1e257c90c3fe8b663c23684a15b0/ui/src/App.vue#L64

**Node-RED:**
We have a fixed set of config options, all get registered to the `RED` object, so we have complete freedom on how this is handled.

**Client Side:**
This is where the difficulty lies. The main Vue app maps a `component` against each of the widget configurations received from the `ui-config` SocketIO event. Because the client-side code for any of the third-party widgets is not stored locally in the directory for Dashboard 2.0, we have no way of mapping the `component` at this point, hence the need to have a different structure.

[...]

PR #351, opened 2023-11-16
Third Party Widget Support

Pezmc:
## Description

Adds support for third party widgets with full Vue support in NR dashboard as per the example widget implemented in https://github.com/FlowFuse/node-red-dashboard-example-node/pull/3.

This works by loading the third party Vue components over HTTP and using UMD to bind those components to `window`, as well are using the same instance of vue in both places (via `window.Vue`)

In practice that looks like this*:
<sub>*Where all of the visible widgets were in fact loaded over HTTP and not in the main bundle</sub>

Screenshot 2023-11-17 at 13 17 57

The **documentation will need to be updated** to reflect this new way to build third party nodes and their corresponding widgets, developers should follow the example (and Vite config) in https://github.com/FlowFuse/node-red-dashboard-example-node to get started.

## Related Issue(s)

For https://github.com/FlowFuse/node-red-dashboard/issues/307

Pairs with: https://github.com/FlowFuse/node-red-dashboard-example-node/pull/3

Follow up to: https://github.com/FlowFuse/node-red-dashboard/pull/336

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Pezmc:
Tests are broken, I believe due to unrelated changes, but handling back to @joepavitt for any finishing toushes.

joepavitt:
FYI @bartbutenaers and @ jtalbourdet just getting this on your radar as I know you're both actively looking at third-party widget support.

This will be an overhaul of the way we integrate (apologies its come so quickly after the first pass), but given the feedback we had in https://github.com/FlowFuse/node-red-dashboard/issues/307 felt it was required.

This new approach will utilise `vite` to run a build (via `npm run build`) in the third-party widget, package everything up into a single `umd.js` file, which we can then inject directly into Dashboard 2.0

With this, you can define any number of Vue Components, dependencies, etc. and it will all get packaged up. You will also be able to write full vue components as you _should_ be able to, and will be consistent with core.

A couple of use cases I still need to iron up:

- How can we utilise the `useDataTracker()` functionality we have in core widgets as we can't currently import that file into the third-party components.
- Consequently, how do we access the main Vue `$sockets` connection & `$state` store access for data retention

I will then be getting documentation fully up to date, including the updated `example` repository which will bring a much better workflow for development.

joepavitt:
Example updates to reflect all (I think) cases of customisation/integration that would be required:

https://github.com/FlowFuse/node-red-dashboard/assets/99246719/8eb74a36-717e-47c3-8de1-5b0c3326f593

Only outstanding item remains to be accessing the Dashboard datastore from within the `ui-example.js` file on the Node-RED side, as this is a `.js` file within Dashboard core that isn't exposed outside of that core package. One entry could be via the `group` that we retrieve, and then exposing the `datastore` at the UI Base level.

joepavitt:
This is all now fully functional - just need to update the documentation

joepavitt (approved):

Issue #358, opened 2023-11-21
Chart: Append/Replace Data Option


joepavitt:
### Description

By default, we assume that any newly received data into `ui-chart` should be appended to the existing data on the chart, however, in the case of showing API data for example, we may actually just want to replace the entire dataset with the newly received data.

Whilst this can be achieved by sending a `[]` msg to the chart to clear data, a lower-code approach would be to provide a configuration option on the chart itself which provided the option to _"Append"_ or "Replace", maybe even being able to define this as a `TypedInput`, so that we can control it on a msg by msg basis.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

colinl:
I think it may be more important that it is available as a msg property than as a config item.  It is common to preload the chart with historical data and then add further data in real time.

joepavitt:
Great point Colin. Node-RED's TypedInputs would allow for this. Example below of how we do this for defining the `topic`:

<img width="452" alt="Screenshot 2023-11-21 at 09 52 26" src="https://github.com/FlowFuse/node-red-dashboard/assets/99246719/e48483de-8f93-4bc0-bf76-a2af908bbf14">

joepavitt:
This has now been requested by multiple people, bumping up the priority

colin-grierson:
My thoughts as a very new Node-red dashboard user
We should have consistency 
- The chart and table nodes are inconsistent. Given a set of data, chart updates/appends while table replaces existing data
- I think updating/appending is natural when each message contains one value, and replacing existing data is natural when a set of values are sent 
Having an explicit control in the configuration object is a great idea - it makes everything obvious
Allowing control via message properties is also a great idea
Ideally all controls should be able to be overridden via message properties
Having controls that are only in the message properties is not so good as they are not obvious and it can be hard to find the option exists
Why is the editor help so much different to the documentation at https://dashboard.flowfuse.com/nodes/XXXXXXXXX.html?
- If it adds too much work to make this happen, how about a link in the editor help to go to the much more complete documentation at flowfuse.com?

I really appreciate what you guys are doing
Regards, Colin

joepavitt:
> The chart and table nodes are inconsistent. Given a set of data, chart updates/appends while table replaces existing data

Very good point - we could/should also have this feature added to table for consistency.

> Why is the editor help so much different to the documentation at

Two reasons:

[...]

PR #364, opened 2023-11-22
Docs: Add payload examples for Bar Chart

### LINKED ISSUE #355: Documentation for chart bar graph
I'm trying to pass a set of data points to a chart to be displayed as a bar graph. I have found chart documentation at https://dashboard.flowfuse.com/nodes/widgets/ui-chart.html - this gives instructions and examples for how to pass data for line and scatter graphs - and a picture of a bar graph :-(
Where can I find instructions and examples for how to pass data to a bar graph?
Apologies if this is not the correct place for this question - please direct me to the correct forum if so
Thanks for your help, Colin

joepavitt:
## Description

Adds documented examples of payloads to send to a bar chart

Closes #355

### DIFF
--- a/docs/nodes/widgets/ui-chart.md
+++ b/docs/nodes/widgets/ui-chart.md
@@ -290,3 +290,93 @@ Scatter Plots accept a variety of payload formats. The following are all valid:
 
 ![Example of a Bar Graph](/images/node-examples/ui-chart-bar.png "Example of a Bar Graph"){data-zoomable}
 *Example of a rendered bar graph.*
+
+Bar charts can only be configured with a "Categorical" x-axis type, the series that each data point will group into is then defined by the "Series" property, with the "y" proeprty defining the value of each bar.
+
+#### Payloads
+
+The payload will heavily depend on the "Series" value used, as such, we provide examples of each as follows:
+
+##### Series: `msg.<property>`
+
+Each message send to the chart will assign to the category depending on `msg.<property>`. In Dashboard 1.0 this was the default behaviour, where `msg.topic` would control the series of the data, and this is still supported in Dashboard 2.0.
+
+```js
+const msg = {
+    "topic": "bar-1",
+    "payload": 1
+}
+```
+
+would set the value of the `bar-1` series to `1`.
+
+##### Series: `string`
+
+Each message received will always bind to the same bar, this restricts the chart to only being able to show a single column/category of data.
+
+```js
+const msg = {
+    "payload": 12
+}
+```
+
+would set the value of the single series to `12`.
+
+##### Series: `key:<key>`
+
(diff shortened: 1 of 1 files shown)


colin-grierson:
Joe, many thanks for your quick response.
Alas, maybe I don't yet have the background knowledge, or am too old school but I can't get it to work.
I want to output the graph in one message (I can make it work with multiple messages)
My payload is structured as in your example (I think):
msg : Object
  object
  payload: array[10]
    0: object
      category: "SASMAR"
      value: 94351308
    1: object
      category: "SASOWNER"
      value: 54322300
    2: object
    3: object
    4: object
    5: object
    6: object
    7: object
    8: object
    9: object
      category: "SASLJ"
      value: 1197152
_msgid: "7631345d0d5c80a5"
In the chart node I have type: Bar. What should I set the Properties Series and Y to?
I have tried quite a number... my best result is with Series: msg. = key:category  and Y: key: = value   
This results in two bars. The first blank, the second for SASLJ and 1197152 - my last data element.
How do I make this work?

In the documentation you talk about setting the property 'Series', and give example data, but do not say how 'Y' is set. I think it would be much easier to follow if we had: Series, Y, data, Result for the various representative cases.

Thanks again, Colin

joepavitt:
> Series: msg. = key:category and Y: key: = value

I would have expected this to work. I will investigate.

joepavitt:
Can I check which version of Dashboard you're using? Just tried this with the following dataset:

```js
[
    {
        "category": "SASMAR",
        "value": 94351308
    },
    {
        "category": "SASOWNER",
        "value": 54322300
    },
    {
        "category": "SASLJ",
        "value": 1197152
    }
]
```

and with the following chart config:

Screenshot 2023-11-23 at 08 54 06

and it rendered:

Screenshot 2023-11-23 at 08 54 20

joepavitt:
I have also just updated the documentation further to be more explicit about the options available.

colin-grierson:
Node-RED version: v3.1.0
Node.js  version: v18.17.1
flowfuse/node-red-dashboard 0.7.2
Found it - imagine a shameface icon..
Series msg.key:category
image
Series key:category
image

Sigh, thanks again for your help

colin-grierson:
Comment/nice to have...
When we are adding a set of data points to a chart, as in my Top 10 Users example, we most likely want to replace the existing chart than add to it. Of course this can be done by sending a separate message containing a null array, but this is a minor hassle - and easy to miss if most of the time the series are the same and only the values change.
It would be nice to have an option on the chart - maybe 'Clear existing data?' Y/N

joepavitt:
> It would be nice to have an option on the chart - maybe 'Clear existing data?' Y/N

You're not alone in your request: https://github.com/FlowFuse/node-red-dashboard/issues/358

joepavitt:
> Found it - imagine a shameface icon..

Maybe I can include more screenshots of configurations in the docs to make this clearer?

PR #370, opened 2023-11-23
Set default of ui-text to match the labelling of Vuetify components

### LINKED ISSUE #347: Switch label uses the wrong colour (arguably)
The Switch widget label displays in a paler colour than, for example, the Text widget label. I think the defaults should be the same.
For example:

image

joepavitt:
## Description

Issue was opened raising the inconsistency between `ui-text` and any other widgets we have that utilise Vuetify. Whilst `ui-text` does allow for customisation to intentionally make this different, we now set the default styling to match.

## Related Issue(s)

Closes #347 

## Checklist

### DIFF
--- a/nodes/widgets/ui_text.html
+++ b/nodes/widgets/ui_text.html
@@ -34,6 +34,10 @@
                 name: 'Gill Sans'
             },
             {
+                value: 'Helvetica, sans-serif',
+                name: 'Helvetica'
+            },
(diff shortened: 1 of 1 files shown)


joepavitt:
Example of the `ui-text` alongside a few `ui-slider` widgets:

Screenshot 2023-11-23 at 14 10 05

PR #438, opened 2023-12-13
(Draft) Plugins API

joepavitt:
## Description

- Adds in API Hooks for Node-RED plugins to extend Dashboard 2.0 server-side functionality
- Adds a new `/_setup` endpoint that is called when Dashboard client first loads. This allows the passing of SocketIO config (which was previously hard-coded to assume `/dashboard` base endpoint.

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -123,7 +123,7 @@ export default ({ mode }) => {
               ]
             },
             {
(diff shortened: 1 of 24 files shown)


joepavitt:
Tests will not pass due to https://github.com/node-red/node-red-node-test-helper/pull/73 - running the fix locally, the tests pass though

PR #472, opened 2024-01-04
UI Template - Add "Page-Scoped" and "UI-Scoped" options for widget placement

### LINKED ISSUE #440: UI Template - Allow for Page/UI Scoped Templates
### Description

Currently, templates are always bound to a `ui-group` in their config. We have the option to introduce `Teleports` into Dashboard, which would allow us to inject content into the App Bar (for example). This currently only then works for whichever page the `ui-template` is bound to (via it's group), it also then renders the widget at the group level still too.

### LINKED ISSUE #445: ui-template can end up with conflicting group dashboard and page entries
In the flows file a ui-template node has entries for group, dashboard (ui-base) and page.  It is possible for these to get out of sync.  For example, see these two template records from the same flows file:
```
{
        "id": "1c35624c74ddbdcf",
        "type": "ui-template",
        "z": "997da33a0beedade",
        "group": "7545fdb22e4a388c",
        "dashboard": "ID-BASE-1",
        "page": "",
....
```
and
```
{
        "id": "de6cfe9d86ed5b93",
        "type": "ui-template",
        "z": "997da33a0beedade",
        "group": "7545fdb22e4a388c",
        "dashboard": "64d151d89e8704cf",
        "page": "0a534af73f5f11fb",
....
```
Note that they have the same group id, but different dashboards.  In the second one the page does belong to that dashboard, but the group does not belong to that page.

This gives rise to a flow file that shows multiple dashboard entries 
image
Both ui-base config nodes show as in use so it is not possible to delete one of them. The only recovery method that I have found is to delete the template node, or hack the flows file to assign the template to the correct ui-base.  The additional ui-base then becomes unused and can be deleted.

To replicate the problem, import a flow containing a template which has different group, page and dashboard ids to the flow being imported into.  Then open the template node and change it to use a group from the original flows and deploy.  This edits the group entry in the template but leaves the original page and dashboard.

joepavitt:
## Description

- Adds a loop on each page layout to render page-scoped widgets
- Adds 2 x additional options to the `ui-template` "scope" property, which allow for a template to render at the `page` and `group` level.
- Adds skeleton styling for the `#app-bar-actions` such that if multiple template\s define content, they position and layout nicely.
- Also ensures that `page`, `ui` and/or `group` are cleared appropriately when not required by the `ui-template` (Fixes #445)

## Related Issue(s)

- Closes #440
- Fixes https://github.com/FlowFuse/node-red-dashboard/issues/445

## Checklist

### DIFF
--- a/nodes/widgets/locales/en-US/ui_template.json
+++ b/nodes/widgets/locales/en-US/ui_template.json
@@ -2,7 +2,9 @@
     "ui-template": {
         "label": {
             "scope": "Type",
-            "local": "Widget",
+            "widget-group": "Widget (Group-Scoped)",
+            "widget-page": "Widget (Page-Scoped)",
+            "widget-ui": "Widget (UI-Scoped)",
             "site-style": "CSS (All Pages)",
             "page-style": "CSS (Single Page)",
             "group": "Group",
--- a/nodes/widgets/ui_template.html
+++ b/nodes/widgets/ui_template.html
@@ -74,8 +74,8 @@
             color: RED._('@flowfuse/node-red-dashboard/ui-base:ui-base.colors.dark'),
             defaults: {
                 group: { type: 'ui-group', required: true }, // for when template is scoped to 'local' (default)
-                dashboard: { type: 'ui-base', required: false }, // for when template is scoped to 'site'
                 page: { type: 'ui-page', required: false }, // for when template is scoped to 'page'
+                ui: { type: 'ui-base', required: false }, // for when template is scoped to 'site'
                 name: { value: '' },
                 order: { value: 0 },
                 width: {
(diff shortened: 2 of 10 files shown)


gemini86:
After updating to 0.11.3, I'm seeing that there is not documentation on what the widget scoping does or how to configure it, which is problematic. This is considered a breaking change that does not default to previously expected behaviour. One of my pages is showing widgets from other pages also included now. The only way to correct this is to change the scope setting and then select the group scope option again and redeploy. The biggest issue though is lack of documentation on what the different scopes do (in a node-red fashion)

joepavitt:
Hi @gemini86 - thanks for raising this. Odd behaviour as I didn't actually change the value at all for the default option, only the label shown in the dropdown, so it _shouldn't_ make a difference.

I will investigate when I'm at my desk tomorrow.

joepavitt:
Documentation has been updated, and a fix for others to avoid this problem should go live today.

PR #486, opened 2024-01-12
Update .gitignore for better management of auto-generated NR files

joepavitt:
## Description

Auto-generated files were being updated in the `user-dir` for Node-RED when running the E2E tests, we don't want auto-generated files in the git repo, they'll be creating at the point the tests run anyway.


joepavitt:
E2E Tests failed for unknown reasons - all pass locally, and verified the failing tests perform as expected when using the UI manually

PR #565, opened 2024-02-10
Update UISwitch.vue

egs2003:
Toggle by pressing the label. This makes it easier to operate on small screens

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-switch/UISwitch.vue
+++ b/ui/src/widgets/ui-switch/UISwitch.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="nrdb-switch" :class="{'nrdb-nolabel': !props.label, [className]: !!className}">
+    <div class="nrdb-switch" :class="{'nrdb-nolabel': !props.label, [className]: !!className}" @click="toggle">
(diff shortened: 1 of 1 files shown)


joepavitt (changes requested):
Thanks for the PR! Great idea for UX improvment.

Whilst this does then enable the label to be clicked for switching, it consequently breaks the ability to click the toggle, as it switches once for clicking the toggle, _and_ toggles back again for clicking the parent.

A better place to put the `@click=""` would be on the `<label />` itself

colinl:
Am I reading this request correctly?  That it would enable clicking the label on a switch to have the same effect as clicking the switch area itself?  I can see problems with this, for example it would presumably mean that clicking anywhere to the left of the switch in this example would toggle the switch, which would not be expected. Also it might be too easy to accidentally toggle the switch while attempting to scroll the display with a touch screen.
image

joepavitt:
You are reading this correctly @colinl 

I can understand from both sides, and am not sure which way to go right now. Dashboard 1.0 was switch-only

PR #598, opened 2024-02-22
UI Gauge: Better zero check & unit label alignment

### LINKED ISSUE #583: ui_gauge does not respect 0 as valid numeric value
### Description

Seems that 0 treated as falsy due the missing validator function. Test with negative scales to reveal the issue.

### Epic/Story

_No response_

joepavitt:
## Description

- Improve the value check when rendering the value text got a `ui-gauge`.
- Fix horizontal alignment of the "units" label, when no icon is provided

## Related Issue(s)

Closes #583

### DIFF
--- a/ui/src/widgets/ui-gauge/UIGauge.vue
+++ b/ui/src/widgets/ui-gauge/UIGauge.vue
@@ -41,7 +41,7 @@
             </g>
         </svg>
         <div ref="value" class="nrdb-ui-gauge-value" :class="'nrdb-ui-' + props.gtype">
-            <span>{{ props.prefix }}{{ value !== undefined ? value : props.min }}{{ props.suffix }}</span>
+            <span>{{ props.prefix }}{{ value ?? props.min }}{{ props.suffix }}</span>
             <label v-if="props.icon || props.units"><v-icon v-if="props.icon" :icon="`mdi-${icon}`" />{{ props.units }}</label>
         </div>
     </div>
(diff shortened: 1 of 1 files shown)


ui/src/widgets/ui-gauge/UIGauge.vue
```diff
@@ -41,7 +41,7 @@
             </g>
         </svg>
         <div ref="value" class="nrdb-ui-gauge-value" :class="'nrdb-ui-' + props.gtype">
-            <span>{{ props.prefix }}{{ value || props.min }}{{ props.suffix }}</span>
+            <span>{{ props.prefix }}{{ value !== undefined ? value : props.min }}{{ props.suffix }}</span>
```
Steve-Mcl, line 44: Joe, 

it looks like you are handling `0` here right?

I think this should use [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) here.  What if the value is `null`?

```suggestion
            <span>{{ props.prefix }}{{ value ?? props.min }}{{ props.suffix }}</span>
```

* if the value is `0`, then `0` will be used
* if the value is `undefined`, then `props.min` will be used
* if the value is `null`, then `props.min` will be used
* if the value is `-1`, then `-1` will be used

However, if there is other code that absolutely prevents `null` making its way to this point, then your solution is fine too.

Steve-Mcl (approved):

Steve-Mcl:
Approved pending tests

Issue #526, opened 2024-01-26
Allow Dashboard to be run from reverse proxy with path change


hardillb:
### Description

Currently the dashboard uses hard coded full paths (stating with `/dashboard`) for assets and the socket.io connection.

Making them all relative paths will allow the dashboard to be proxied onto to a different path.

Making the paths relative in the `dist/index.html` moves things forward but the socket.io connection is still trying to open a full path.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Not sure this is a "small" sizing? If I recall, there were a lot of headaches when trying to do this before, but I'll see what we can do. Have any quick config I can run locally to give me a test environment for this?

I suspect there is overlap in a solution with https://github.com/FlowFuse/node-red-dashboard/issues/437

hardillb:
yep, I'll update when done with call

hardillb:
create the following file called `nginx.conf`

```
worker_processes 1;

events {
  worker_connections 1024;
}

error_log /dev/stderr;

http {

  server {
    listen 80;
    access_log /dev/stdout;

    root /var/www/html;

    location /test/ {
	    rewrite /test/(.*) /dashboard/$1 break;
	    proxy_pass http://172.17.0.1:1880/dashboard/;
    }
  }	
}
```

Then run:

```
docker run --rm -v `pwd`/nginx.conf:/etc/nginx/nginx.conf -p 8080:80 nginx
```

You may need to change `172.17.0.1` to host.docker.internal` on a mac.

This expects Node-RED to be running on port 1880 with a dashboard deployed and you access it via `http://localhost:8080/test/`

L3v3L:
This probably affects Home Assitant users, as all node red endpoints are automatically proxied to /endpoint/ path

robmarcer:
@hardillb can you let me know which customer this is? Either post a Hubspot link or tell me their details and I'll add it.

robmarcer:
This customer requested this feature - https://app-eu1.hubspot.com/contacts/26586079/record/0-1/2454

joepavitt:
- Complexity currently exists that `socketio` is using an absolute path, to use a relative path then means that SocketIO defaults to the base URL being `window.location.host` ([docs](https://socket.io/docs/v3/client-api/#iourl))
[...]

Issue #437, opened 2023-12-13
D2 fails if httpNodeRoot is non-blank


TotallyInformation:
As mentioned in the forum: https://discourse.nodered.org/t/dashboard-2-beta-development/83550/52?u=totallyinformation

A regression seems to have crept in because - despite completely removing all D2 nodes - including config nodes - removing D2, restarting, adding D2, restarting then adding a single text node, I cannot access the D2 dashboard page.

This is because, as always, I have the httpNodeRoot set to "nr" so that Node-RED page URLS are always prefixed - found this problem the hard way in the past whihc is why it is always set on my dev PC. D2 is still making assumptions that ignore httpNodeRoot and I get these errors:

![image](https://github.com/FlowFuse/node-red-dashboard/assets/1591850/3ae3e12f-b4e4-4ace-8de1-32054797473c)

As you can see, it is trying to load resources without the /nr prefix which is incorrect and so cannot find them.

Almost certainly, this is the result of coded URL's that don't use the relative prefix of ./ so that they pick up the current URL prefix. The link in the Node-RED Editor has been corrected but the resource links haven't.

joepavitt:
Thanks for capturing this Julian. Do you have an example of how the httpRootNode is set please?

TotallyInformation:
Sorry for late response. `httpNodeRoot` is set in `settings.js`, it is a standard configuration option for Node-RED. I have it always set to `nr` in my dev setup so that I am always testing it. It means that http-in/-out, ws in/out and uibuilder (with the default config) all have all URL paths prefixed with `/nr/`.

joepavitt:
I suspect this is also causing #563 and issues with running Dashboard on Home Automation too

PR #604, opened 2024-02-23
More robust path replacement to account for proxy & httpNodeRoot configs

### LINKED ISSUE #563: Error running When Dashboard 2.0 on Bosch Ctrlx Core
The pallete installs successfully, but when I add any dashboard element and navigate to the dashboard url it is blank. When I inspect the page it shows the following error. 
index-j47xBCGS.js:1 Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/html". Strict MIME type checking is enforced for module scripts per HTML spec.

joepavitt:
## Description

- Ensure we have a more robust SocketIO path replacement, which was introduced in #595 (for #526) and had broken the work we'd done in #581 (a fix for #437)

## Related Issue(s)

Fixes #563

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -141,11 +141,10 @@ export default {
             Object.values(payload.pages).forEach(page => {
                 // check that the page's bound UI is also in our config
                 if (payload.dashboards[page.ui]) {
-                    const ui = payload.dashboards[page.ui]
-                    // re-write base path in case of proxy
-                    ui.path = this.setup.basePath
+                    // const ui = payload.dashboards[page.ui]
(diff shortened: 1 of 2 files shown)


hardillb (approved):

hardillb:
Looks ok, but you have a test failure. I'll trigger a re-run in case it's a timing thing

joepavitt:
Yeah, it's a very unreliable E2E test set, Cypress doesn't like Websockets... Always run fine when local

hardillb:
@joepavitt I'll let you decide to merge then

PR #607, opened 2024-02-23
onChange: Ensure custom handlers have context of any general changes made to the msg. object

### LINKED ISSUE #588: Switch widget does not send _client property the first time it is clicked after NR restart
A ui-switch with default configuration does not include the `_client` property the first time it is clicked (after widget deploy or restart).  It is present for subsequent clicks.

joepavitt:
## Description

- When a widget was using a custom `onChange` handler (only currently used by `ui-switch`) we were only considering the `value` received server-side by the widget, and not any general changes made to `msg` by the global `onChange` parent handler. 
- Because the default handler is scoped _inside_ the parent, that worked fine when modifying the existing `msg.` object, which modifies the `msg` object to include `_client`.
- What's confused me the most is how it _ever_ works, even second/third time round, as I can't find the path by which that would have happened, can only think there is some odd two-way memory binding with JS through the datastore

## Related Issue(s)

Fixes #588

### DIFF
--- a/docs/contributing/guides/registration.md
+++ b/docs/contributing/guides/registration.md
@@ -123,8 +123,6 @@ onChange: async function (msg, value) {
     const off = RED.util.evaluateNodeProperty(config.offvalue, config.offvalueType, wNode)
     msg.payload = value ? on : off
 
-    wNode._msg = msg
-
     // simulate Node-RED node receiving an input
     wNode.send(msg)
 }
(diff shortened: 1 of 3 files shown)


Steve-Mcl (approved):
Looks sane 👍

joepavitt:
Thanks Steve - sorry, didn't mean to disturb you on your day off - I'd re-assigned to Nick when I realised.

Steve-Mcl:
It's not a problem, there will be a point later in the day where I can't. But for now, small quick wins like these, I'm able to assess from phone and get them out.

PR #633, opened 2024-03-01
UI Control: Ensure better error reporting for ui-control when group/pages not found & update tests

joepavitt:
## Description

- Now that we have the option to modify the navigation style, I've been able to unskip some of the `ui-control` tests that were bugging out due to how the Vuetify navigation drawer behaves. 
- Whilst doing that, I spotted an issue with poor error reporting when showing/hiding groups in `ui-control`, so this applies a fix there too.

### DIFF
--- a/nodes/widgets/ui_control.js
+++ b/nodes/widgets/ui_control.js
@@ -13,8 +13,12 @@ module.exports = function (RED) {
         function updateStore (all, items, msg, prop, value) {
             items.forEach(function (item) {
                 const i = all[item]
-                // update the state store for each page
-                statestore.set(ui, i, msg, prop, value)
+                if (i) {
+                    // update the state store for each page
+                    statestore.set(ui, i, msg, prop, value)
+                } else {
+                    node.error('No item with the name ' + item + ' found')
+                }
             })
         }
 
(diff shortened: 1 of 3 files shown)


Steve-Mcl:
@joepavitt, I assume you meant to set me as reviewer not assignee?

joepavitt:
yep, apologies @Steve-Mcl

nodes/widgets/ui_control.js
```diff
@@ -13,8 +13,12 @@ module.exports = function (RED) {
         function updateStore (all, items, msg, prop, value) {
             items.forEach(function (item) {
                 const i = all[item]
-                // update the state store for each page
-                statestore.set(ui, i, msg, prop, value)
+                if (i) {
+                    // update the state store for each page
+                    statestore.set(ui, i, msg, prop, value)
+                } else {
+                    node.error('No item with the name ' + item + ' found')
```
Steve-Mcl, line 20:
```suggestion
                    node.error('No item with the name ' + item + ' found', msg)
```

If `msg` is not present in `node.error`, it cannot be caught by the `catch` node.

Steve-Mcl (commented):
Are we expecting the user to be able to catch and handle/mute these errors?

joepavitt:
> Are we expecting the user to be able to catch and handle/mute these errors?

I hadn't, but agree with your changes

Steve-Mcl (approved):
Approved pending tests re-run succeeds

PR #640, opened 2024-03-04
Update ui-template.md

Steve-Mcl:
Correction to template H1 (currenlty says "Text ui-template" - should say "Template - ui-template")


joepavitt (approved):

joepavitt:
Thanks Steve - have viewed that page so many times and never spotted that 🤦

PR #651, opened 2024-03-07
UI Gauge - Better dynamic sizing and smaller size rendering

### LINKED ISSUE #557: UI Gauge - Improve sizing/layout at smaller sizes (1x and 2x widths)
### Description

Details: https://discourse.nodered.org/t/dashboard-2-0-is-now-generally-available/84913/86?u=joepavitt

3/4 Gauges have a hard-coded height, which is a problem when the width is smaller than the height. The gauges are also pinned to the bottom of the rendered area, which gives the illusion that the padding between title and gauge is the problem.

### Epic/Story

_No response_

joepavitt:
## Description

- With these changes, `ui-gauge` will no longer have a forced 150/300px height, instead it will always stay bound within the `w/h` limits defined in the NR Editor.

## Related Issue(s)

Closes #557

### DIFF
--- a/nodes/widgets/ui_gauge.html
+++ b/nodes/widgets/ui_gauge.html
@@ -46,7 +46,7 @@
             group: { type: 'ui-group', required: true },
             order: { value: 0 },
             width: {
-                value: 0,
+                value: 3,
                 validate: function (v) {
                     const width = v || 0
                     const currentGroup = $('#node-input-group').val() || this.group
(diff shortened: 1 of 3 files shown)


nodes/widgets/ui_gauge.html
joepavitt, line 1: this sets the default `ui-gauge` height and width to something reaosnable.

knolleary (approved):

PR #655, opened 2024-03-07
UI Group - Component-ise the widget-group rendering to ensure consistency

### LINKED ISSUE #650: Alignment of groups in pages
I have a page laid out as 'Fixed', and 2 groups;
The first group is sized 6 x 6 and which contains gauges.
The second group is 3 x 6 and does not contain any widgets.
Together, they look like this;

However, whenever I add something to the second group, the group's height increases, and is no longer bottom aligned with the first group.
By example, I've added a 3 x 3 text box and some random text.

In my example above, the group was 3 x 6, and the widget was only 3 x 3, so why was it necessary to auto increase the size of the group, surely 3 x 3 would fit with room to spare??

Forum discussion - https://discourse.nodered.org/t/alignment-of-groups-in-pages-dashboard-2/86210?u=paul-reed

joepavitt:
## Description

- Creates a new `widget-group` Vue component
- This new component handles the layout rendering of widgets within a single group. Now centralised given that the different layouts only affect how _groups_ are laid out, not their contents.

## Related Issue(s)

Fixes #650

### DIFF
--- a/ui/src/layouts/Flex.vue
+++ b/ui/src/layouts/Flex.vue
@@ -15,19 +15,7 @@
                         {{ g.name }}
                     </template>
                     <template #text>
-                        <div class="nrdb-group-widgets nrdb-layout-group--grid" :style="`grid-template-columns: repeat(${ g.width }, 1fr); grid-template-rows: repeat(${g.height}, minmax(${rowHeight}px, auto)); `">
-                            <div
-                                v-for="w in widgetsByGroup(g.id)"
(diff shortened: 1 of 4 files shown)


joepavitt:
### Grid:

Screenshot 2024-03-07 at 17 13 54

### Fixed:

Screenshot 2024-03-07 at 17 14 17

### Notebook (where all groups are forced to 6 units wide:

Screenshot 2024-03-07 at 17 14 33

Steve-Mcl:
Approved pending e2e test re-run

PR #666, opened 2024-03-12
UI Form: Ensure isValid updated after clear()

### LINKED ISSUE #654: Widget Form - Is sending message when required field absent
### Description

Hi

In case i have a `Form` that have a text field that is required
image

If the user in the dashboard, perform a `clear` the `submit` button is enabled and allows a message to be sent with payload `empty`

[screen-capture (1).webm](https://github.com/FlowFuse/node-red-dashboard/assets/17281826/fa01147d-e4e8-4119-b50f-d3f07b2171a0)

Would expect that since the `required` field is not present/filled no message would be sent.

### Epic/Story

_No response_

I am no FlowFuse team member

joepavitt:
## Description

- ~Updates `isValid` after a `clear()` button is clicked~
- ~Could have run `form.validate()`, but this then highlighted the invalid fields, which had been raised an an annoying feature previously on first load.~

Updated approach: wrap Vuetify's [`form.reset()`](https://vuetifyjs.com/en/api/v-form/#slots-default)

## Related Issue(s)

Closes #654

### DIFF
--- a/cypress/fixtures/flows/dashboard-forms.json
+++ b/cypress/fixtures/flows/dashboard-forms.json
@@ -87,7 +87,7 @@
         "height": 0,
         "options": [
             {
-                "label": "Name:",
+                "label": "Name",
                 "key": "name",
                 "type": "text",
                 "required": true,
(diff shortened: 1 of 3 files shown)


joepavitt:
Initial approach fails the tests as `isValid` loses two-way binding. Looking for alternative approach.

joepavitt:
E2E Tests incoming first thing tomorrow

PR #687, opened 2024-03-16
UI Table - Add row selection config, fix column arrangement & preserve defined labels

### LINKED ISSUE #397: ui-table 'send data on click'
### Description

v1 ui-table node has an input and output.
The output can be used to 'send data on click' where clicking any row/column will send that data to the nodes output.

v2 ui-table does not have either output or the option to 'send on click' any of the table data.

### Properties

_No response_

### Events

_No response_

### Controls

_No response_

### Existing Examples

Sending data on output is a very important function that is used a great deal to drill into / get more details on any table data.
image

Typical example flow showing the use of this input/output.
image

Clicking on the table column both shows the full data and also does a db lookup on that clicked term for display in a second table.

I am no FlowFuse team member

### LINKED ISSUE #613: TableUI - MouseOver Hightlight and ClickRow sends row out to payload
### Description

Very happy with the new Table functionality where sending an array to it will give a usable table output.
Great work!

A very nice to have would be to highlight a table row on mouse-over. In addition, when a row is selected, it would come in very handy to have the row data be sent as output as a message payload.

I am no FlowFuse team member

### LINKED ISSUE #677: Table-UI FR: make data entries clickable
### Description

As a dashboard user viewing information in a Table-UI node, I would like to be able to select an entry, so further action can be taken on my choice, like pre-fill another form or showing a modal window for deleting that entry.
It would be ideal if also the column that was clicked within that row would be reported.

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

### LINKED ISSUE #678: Table - Labels not being preserved, and column order not being respected
### Current Behavior

This happened after upgrading to v1.4.1, already restarted node-red service a couple of times and force a `Ctrl+F5` and `developer tools, disable cache` to check if it could be some issue related to cache behaviour.

Is not possible to define the `label` for the tables.

Trying to edit the table as:
image

After clicking `Done`, the node does not change status to `blue` indicator, indicating that as changed configuration, and if open again shows
image

And in the page, it shows as:

image

Already tried to place a new table, just in case was an issue with the migration of existing tables nodes, and the issue is the same.

Another thing, related to this...
When editing a Table Node, and disabling the `Auto calculate Columns`, and clicking `add column` the following error appears in the browser console (page of NodeRed Editor):
image

One other issue, may be related to this, is that the column order is also not being respected.

Table configuration:
image

It Showns on Webpage as:
image

### Expected Behavior

Label description should be preserved and shown correctly in the WebPage.

### Steps To Reproduce

_No response_

### Environment

- Dashboard version: 1.4.1
- Node-RED version: 3.1.7
- Node.js version: 20.11.0
- npm version: 10.4.0
- Platform/OS: Ubuntu 22.04.03
- Browser:Microsoft Edge, Version 122.0.2365.80 (Official build) (64-bit)

joepavitt:
## Description

- Adds a new "Selection" configuration on `ui-table`, providing three options:
    - **None:** Behaviour as before, just displays the data and nothing else
    - **Multi Selection:** Shows checkbox against each row, and outputs an array of any items/rows selected
    - **Single Row Selection:** Makes the entire row clickable, and outputs the full row `item` as an object when clicked.
- Updates documentation to include the new configuration option

### Outstanding Items

- [x] E2E Tests

## Related Issue(s)

Closes #397 
Closes #677 (duplicate)
Closes #613 (duplicate)
Closes #678 

## Checklist


Steve-Mcl:
Joe, i will pull this tomorrow and give it a run through (do it some justice)

joepavitt:
What would be the best way to address this, `oneditprepare`?

Steve-Mcl:
> What would be the best way to address this, `oneditprepare`?
> 
> ```js
> if (!this.selectionType) {
>     this.selectionType = 'none'
> }
> ```
> 
> Should do the job, right?

It will however that will cause the node to become dirty even though user has not changed anything.  We try quite hard to avoid that in core Node-RED but not certain there is an absolute need here.

If you want to have inplace upgrades, you have to avoid setting `this.anything` unless user has specifically chosen to change the property in the UI.  That means you need to do 2 things:
1. test for existence of the property in `oneditprepare`, if not present, select the correct item in the UI
2. test for existence of the property in `oneditsave`, if present, then apply the value from UI

(or some combination of the above - i can dig out a working example if you need it)

nodes/widgets/locales/en-US/ui_table.json
```diff
@@ -8,7 +8,13 @@
...
             "key": "Key",
             "label": "Label",
-            "column": "column"
+            "column": "column",
+            "selection": "Selection"
+        },
+        "selection": {
+            "none": "None",
+            "checkbox": "Multi Selection",
+            "click": "Single Row Selection"
```
Steve-Mcl, line 17: I'm not feeling these Joe. 

They are much more about "row events" than "selection"


Perhaps "Row Checkboxes" and "Row Click" are more descriptive?

image

Happy to be overridden of course.

nodes/widgets/ui_table.html
```diff
@@ -130,6 +138,14 @@
         <label for="node-input-maxrows"><i class="fa fa-tag"></i> <span data-i18n="ui-table.label.maxRows"></label>
         <input type="number" id="node-input-maxrows">
     </div>
+    <div class="form-row">
+        <label for="node-input-selectionType"><i class="fa fa-crosshairs"></i> <span data-i18n="ui-table.label.selection"></span></label>
+        <select id="node-input-selectionType">
+            <option value="none" data-i18n="ui-table.selection.none"></option>
+            <option value="checkbox" data-i18n="ui-table.selection.checkbox"></option>
+            <option value="click" data-i18n="ui-table.selection.click"></option>
```
Steve-Mcl, line 146: Should single come before multiple like it does in the docs?

```suggestion
            <option value="none" data-i18n="ui-table.selection.none"></option>
            <option value="click" data-i18n="ui-table.selection.click"></option>
            <option value="checkbox" data-i18n="ui-table.selection.checkbox"></option>
```

docs/nodes/widgets/ui-table.md
```diff
@@ -4,6 +4,7 @@ props:
     Size: Controls the width of the button with respect to the parent group. Maximum value is the width of the group.
     Label: The text shown within the button.
     Max Rows: Defines the maximum number of data-rows to render in the table. Excess rows will be available through pagination control.
+    Selection: Provides three options for table interaction - "None", "Multi Selection" and "Single Selection"
```
Steve-Mcl, line 7:
```suggestion
    Selection: Provides three options for table interaction - "None", "Single Selection" and "Multi Selection"
```

Match docs ordering

nodes/widgets/locales/en-US/ui_table.json
joepavitt: How about "Interaction" as the label, with "Checkboxes" and "Click" as the label options?

joepavitt: Actually, I think "Selection" works better than "Interaction"

docs/nodes/widgets/ui-table.md
Steve-Mcl, line 7:
```suggestion
    Selection: Provides three options for table interaction - "None", "Click" and "Checkbox"
```
update to align with last iteration

Steve-Mcl:
Approved, merge after test pass.

Steve-Mcl:
usual ui-control e2e test fail (unrelated). Merging

PR #690, opened 2024-03-19
UI Chart - Have the default bg color for points as that defined in the NR Editor

joepavitt:
## Description

- Scatter plots were difficult to read as the background color was semi-transparent grey for all points. As such, we're changing the default to be a solid colour instead,

## Related Issue(s)

https://discourse.nodered.org/t/add-axis-labels-to-ui-chart/86097/6

### DIFF
--- a/ui/src/widgets/ui-chart/UIChart.vue
+++ b/ui/src/widgets/ui-chart/UIChart.vue
@@ -240,15 +240,16 @@ export default {
          */
         addToLine (datapoint, label) {
             // consider msg.topic (label) as the label for the series
-            const datalabels = [...new Set(this.chart.data.datasets?.map((set) => {
+            const dataLabels = [...new Set(this.chart.data.datasets?.map((set) => {
                 return set.label
             }))]
-            const index = datalabels?.indexOf(label)
+            const index = dataLabels?.indexOf(label)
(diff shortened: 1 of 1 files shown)


ui/src/widgets/ui-chart/UIChart.vue
```diff
@@ -240,15 +240,16 @@ export default {
...
             }))]
-            const index = datalabels?.indexOf(label)
+            const index = dataLabels?.indexOf(label)
             // the chart is empty, we're adding a new series
             if (index === -1) {
                 const radius = this.props.pointRadius ? this.props.pointRadius : 4
                 this.chart.data.datasets.push({
-                    borderColor: this.props.colors[datalabels.length],
+                    borderColor: this.props.colors[dataLabels.length],
+                    backgroundColor: this.props.colors[dataLabels.length],
```
Steve-Mcl, line 252: Joe, wont setting the background and border colours the same make the borders invisible (or is that the point)?

joepavitt: Yeah, it does render the borders useless, it's a quick iteration, will gauge community feedback and adjust again accordingly. May make it a config at some point.

PR #703, opened 2024-03-22
UI Base - Add a helper function for retrieving Dashboard hosted URL

### LINKED ISSUE #693: Ability to get the dashboard URL on the server-side
### Description

Hi guys,

As discussed on Discourse (see [here](https://discourse.nodered.org/t/red-nodes-eachconfig-is-undefined/86621/5?u=bartbutenaers)), I need the dashboard URL in the _**server-side**_ code of my web-push UI node.

The dashboard D2 sidebar *'open dashboard'* button  does already calculate this URL (see [code](https://github.com/FlowFuse/node-red-dashboard/blob/main/nodes/config/ui_base.html#L324)), but I need some kind of way to get hold of that URL on the server-side.  So I ***don't*** need it in my frontend Vue code!

When a message arrives in the onInput handler on the server-side of my web-push node, I need to be able to enrich the message with the dashboard URL before I forward the message to the (Google, Apple, Microsoft, ...) push service.

Otherwise people need to start hardcoding it in their messages, which is not really user friendly from a maintenance point of view.

Thanks (again)!!!
Bart

joepavitt:
## Description

- Adds a new `base.getBaseURL()` function which will return the relevant URL where our Dashboard is hosted, taking into account the `RED.settings.httpNodeRoot`, and cleaning the concatenated path.

## Related Issue(s)

Closes #693

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -759,6 +759,17 @@ module.exports = function (RED) {
          */
         node.emit = emit
 
+        node.getBaseURL = function () {
+            // get the endpoint for the ui-base
+            const path = n.path || ''
+            // get our HTTP root, defined by NR Settings
+            const base = RED.settings.httpNodeRoot || '/'
+            const basePart = base.endsWith('/') ? base : `${base}/`
+            const dashPart = path.startsWith('/') ? path.slice(1) : path
+            const fullPath = `${basePart}${dashPart}`
+            return fullPath
+        }
+
         /**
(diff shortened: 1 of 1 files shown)


bartbutenaers:
Hi @joepavitt,

I tested your PR and it indeed works:

image

There is a difference with the *"Open dashboard"* button in the sidebar.  That one offers a full url, while the new function in this PR only returns the last part of the path.  I assume you cannot return the full path in this new function, because you don't know which hostname is being used?  But not sure...

Steve-Mcl (approved):

PR #724, opened 2024-04-02
Docs: Improve "Getting Started" docs

joepavitt:
## Description

- Copies over content from https://flowfuse.com/blog/2024/03/dashboard-getting-started/ to improve the underlying documentation with the latest workflow


joepavitt:
Okay, it doesn't copy the whole article, just two of the sections. I can re-write if required.

joepavitt:
Comment from @Paul-Reed:

> Shouldn't all the documentation really be in one place instead of users having to follow links to other sites...

ZJvandeWeg (approved):

PR #725, opened 2024-04-02
Docs: Add link to Discourse Forum

joepavitt:
## Description

Adds link to the "dashboard-2" topic on the Node-RED forums.

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -54,6 +54,7 @@ export default ({ mode }) => {
         { text: 'Docs', link: '/getting-started.html' },
         { text: 'Widgets', link: '/nodes/widgets.html' },
         { text: 'Contributing', link: '/contributing/' },
+        { text: 'Forum', link: 'https://discourse.nodered.org/tag/dashboard-2' },
         { text: 'FlowFuse', link: 'https://flowfuse.com' },
         { text: 'GitHub', link: 'https://github.com/FlowFuse/node-red-dashboard' }
       ],


docs/.vitepress/config.js
```diff
@@ -54,6 +54,7 @@ export default ({ mode }) => {
         { text: 'Docs', link: '/getting-started.html' },
         { text: 'Widgets', link: '/nodes/widgets.html' },
         { text: 'Contributing', link: '/contributing/' },
+        { text: 'Forum', link: 'https://discourse.nodered.org/tag/dashboard-2' },
```
ZJvandeWeg, line 57: Can we also include the discord for chat?

ZJvandeWeg (changes requested):

docs/.vitepress/config.js
Paul-Reed: Shouldn't all the documentation really be in one place instead of users having to follow links to other sites...

joepavitt: The documentation is all available in the docs, but this links to the Forum for discussion of topics, or to the Discord for those that are operating there instead (as proposed by ZJ)

joepavitt: RE: Discord link specifically, I'm a little cautious of over-loading that top-right corner, but can add it if required... doesn't stop this PR though which is constrained just to the "Forum".

Adding a Discord link should be a new PR imo

Paul-Reed:
Sorry Joe, my comment should have gone to #724 and not here (working via
phone!)

ZJvandeWeg (approved):

ZJvandeWeg:
@joepavitt The E2E is failing, can you take a look?

joepavitt:
> The E2E is failing, can you take a look?

Known problem with inconsistencies in the websocket traffic when running in GH

PR #738, opened 2024-04-03
UI Gauge - Add in "tile" gauge type

joepavitt:
## Description
Screenshot 2024-04-03 at 17 38 09

- Add new "tile" type for `ui-gauge` which is a solid block colour (defined by the segments) and renders the label centrally as per common Manufacturing HMIs
- Break `UIGauge` into multiple sub widgets for easier control over functionality
- Add documentation (in-Editor and online)
- Add E2E tests to ensure correct type is rendered

## Related Issue(s)

Closes #737 

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Backport needed? -> add the `backport` label
 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #750, opened 2024-04-06
Fix: Page Title & add E2E test to prevent regression

joepavitt:
## Description

Introduced in #738 

- This fixes the injection of "Test" and adds an E2E to make sure we don't regression this again.


Steve-Mcl (approved):
Sorry I guess I missed that in review too 😞

PR #757, opened 2024-04-07
Switch button group label as text

m-schaeffler:
## Description

The CSS classes of the labels of `ui-switch` and `ui-buttongroup` are changed according to `ui-text`:
- in the template `v-label` is replaced by a name according to the general naming style
- the style definition is amended so that it is similar to `ui-text`

## Related Issue(s)

This gives a consistent UI experience.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts


joepavitt:
Agree with the move, although, rather than having disparate classes defined in each widget, we could have a higher level `nrdb-ui-label` class in `/stylesshets/common.css` and then use that class in the relevant `ui-text`, `ui-switch` and `ui-button-group` locations?

joepavitt (commented):
Agree with the move, although, rather than having disparate classes defined in each widget, we could have a higher level nrdb-ui-label class in /stylesshets/common.css and then use that class in the relevant ui-text, ui-switch and ui-button-group locations?

m-schaeffler:
@joepavitt Yes, I think your Idea is the better solution

m-schaeffler:
please do it in one gig shot.
Remark: this hard coded color is already in several components.

Issue #628, opened 2024-02-29
Notification FR: send a message when the user clicks Close or the Notification Auto-Closes


SynoUser-NL:
As a dashboard designer, I would like to possibility to take further action when the dashboard user clicks the close button or when the notification is closed automatically (by time-out).

SynoUser-NL:
Thank you!

PR #766, opened 2024-04-10
Notification output & output msg when button clicked

bartbutenaers:
## Description

I need an output message as soon as the button on the notification popup dialog is clicked:

But not sure if the implementation of this PR is correct:
1. The payload contains just 'clicked':

   image
   
   Not sure if that should be customizable?
2. There is now 1 output instead of 0.  But for existing nodes in the flow that won't be the case.  Reason is that in the ui_notification.html file, the number of outputs is defined twice:

   ```
            defaults: {
                ....
                outputs: { value: 1 },
                ...
            },
            outputs: 1,
   ```
   Normally you don't specify the `outputs` as a variable in the `defaults` section.  Only for nodes like the Function-node, because there in the config screen the number of outputs can be manually entered (which means that entered value needs to be stored inside the node, via the outputs property in the defaults section).
   
   But because that property is already there, a value of 0 has been stored for it in all existing notification nodes.   As a result existing notification nodes in the flow still won't have an output.  When I remove the outputs variable from the defaults, that does not help because the value keeps being stored for existing nodes.
4. In case the dialog is being closed by a timeout time, I don't send an output message.  Not sure if that is correct behaviour...

## Related Issue(s)

Closes https://github.com/FlowFuse/node-red-dashboard/issues/628

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-notification/UINotification.vue
+++ b/ui/src/widgets/ui-notification/UINotification.vue
@@ -17,7 +17,7 @@
         <template v-if="props.allowDismiss" #actions>
             <v-btn
                 variant="text"
-                @click="close(true)"
+                @click="close('clicked')"
             >
                 {{ props.dismissText || "Close" }}
             </v-btn>
@@ -84,7 +84,7 @@ export default {
 
             this.timeouts.close = setTimeout(() => {
                 // close the notification after time has elapsed
-                this.close(false)
+                this.close('timeout')
             }, time)
 
             // update the progress bar every 100ms
@@ -96,11 +96,9 @@ export default {
                 this.countdown = 100 - (elapsed / parseFloat(this.props.displayTime)) * 100
             }, 100)
         },
-        close (sendMsg) {
+        close (payload) {
             this.show = false
-            if (sendMsg) {
-                this.$socket.emit('widget-change', this.id, 'clicked')
-            }
+            this.$socket.emit('widget-change', this.id, payload)
 
             clearTimeout(this.timeouts.close)
             clearInterval(this.timeouts.step)
--- a/ui/src/widgets/ui-notification/UINotification.vue
+++ b/ui/src/widgets/ui-notification/UINotification.vue
@@ -98,7 +98,7 @@ export default {
         },
         close (payload) {
             this.show = false
-            this.$socket.emit('widget-change', this.id, payload)
+            this.$socket.emit('widget-action', this.id, payload)
 
             clearTimeout(this.timeouts.close)
             clearInterval(this.timeouts.step)
--- a/ui/src/widgets/ui-notification/UINotification.vue
+++ b/ui/src/widgets/ui-notification/UINotification.vue
@@ -98,7 +98,6 @@ export default {
         },
         close (sendMsg) {
             this.show = false
-debugger
             if (sendMsg) {
                 this.$socket.emit('widget-change', this.id, 'clicked')
             }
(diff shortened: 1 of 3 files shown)


ui/src/widgets/ui-notification/UINotification.vue
```diff
@@ -96,8 +96,12 @@ export default {
                 this.countdown = 100 - (elapsed / parseFloat(this.props.displayTime)) * 100
             }, 100)
         },
-        close () {
+        close (sendMsg) {
```
ZJvandeWeg, line 99: What do you think about making it default to true? (Not sure if Vlue works the same as JS, btw). Than the diff on line 20 can be reverted and it seems cleaner.

```suggestion
        close (sendMsg = true) {
```

ui/src/widgets/ui-notification/UINotification.vue
```diff
@@ -96,8 +96,12 @@ export default {
...
             this.show = false
+debugger
```
ZJvandeWeg, line 101:
```suggestion

```

bartbutenaers: Hi @ZJvandeWeg,
I had not cleaned up my code yet, because I assume I will get some other feedback that I need to implement before it is being approved.

ui/src/widgets/ui-notification/UINotification.vue
bartbutenaers: Well to be honest I have never used default parameters yet in javascript, but if you guys like it then it is fine for me.  Personally I like to see explcit values, because then I immediately see what the value is in each call.  But that is just a matter of personal taste...

joepavitt:
Currently, messages sent to the notification are then passed through too - which I don't think should be the case?

joepavitt:
Just found https://github.com/FlowFuse/node-red-dashboard/issues/628 which also asks for an event when the timeout expires, with "clicked" as one type, a natural event to also include here would be "timeout"?

bartbutenaers:
@joepavitt,
I had overlooked that issue.  
Good that you found it, because now the parameter of the `close` function makes more sense.

The code is now updated to send an output message both on timeout or when the button has been clicked:

image

joepavitt (changes requested):

ui/src/widgets/ui-notification/UINotification.vue
```diff
@@ -96,8 +96,11 @@ export default {
                 this.countdown = 100 - (elapsed / parseFloat(this.props.displayTime)) * 100
             }, 100)
         },
-        close () {
+        close (sendMsg) {
             this.show = false
+            if (sendMsg) {
+                this.$socket.emit('widget-change', this.id, 'clicked')
```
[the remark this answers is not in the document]
bartbutenaers: Done

joepavitt (approved):

cruce:
I am using Dashboard 2.0 version 1.13.0. I installed it through Manage Palette. The output to ui-notification happens as soon as the flow executes the node and debug on the output shows the incoming msg.payload. 

Is the output supposed to wait until timeout or button clicked?
I am sure I am missing something obvious, how do I change debug node to get the msg:Object in the debug images I see above?

I had installed node-red on my pi 3B+ using the install script and it installed NodeJS v18.20.4 and npm v10.7.0.
I am going to try and run the script again to have it install NodeJS v20 as I have been having other issues with node-red service crashing and restarting.

bartbutenaers:
@cruce,
What you describe is one of the issues that should be solved by [this](https://github.com/FlowFuse/node-red-dashboard/pull/932) pull request.

cruce:
@bartbutenaers Thanks. I have subscribed to that pull request.

PR #774, opened 2024-04-16
Only apply the padding top/bottom changes for v-text-field to outlined variant

### LINKED ISSUE #773: Overriding padding top/bottom offsets text incorrectly in v-text-field
### Current Behavior

If a user includes their own `v-text-field` with the default `variant, then the spacing between the label and value is negligible. This occurs because we override the default top/bottom padding for the `v-field__input`, in order to account for our slightly taller text fields (48px).

### Expected Behavior

The adjusted height should only impact the variant of text-field we're using in core, the other varians should not have their behaviour overridden

### Steps To Reproduce

See: https://discourse.nodered.org/t/ui-text-input-value-near-to-label/87074/2

### Environment

_No response_

joepavitt:
## Description

More precise CSS selector for text fields

## Related Issue(s)

Closes #773

### DIFF
--- a/ui/src/layouts/grid-groups.css
+++ b/ui/src/layouts/grid-groups.css
@@ -1,6 +1,9 @@
 /** https://github.com/vuejs/vue-loader/releases/tag/v12.2.0 */
 :deep(.v-field__input) {
     --v-field-input-min-height: calc(var(--widget-row-height) - 4px);
+}
+
+:deep(.v-field--variant-outlined) {
     --v-field-input-padding-top: 12px;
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):

Steve-Mcl:
Approved however the e2e error is not the usual ui-control error.

> ui-form
>     blurring a required field runs validation


Happy to merge as is (error may have slipped in another PR as this seems unrelated)

joepavitt:
> Approved however the e2e error is not the usual ui-control error.

It's always one of those two that fails - 🤷

Issue #777, opened 2024-04-17
Button: Add passthrough


m-schaeffler:
### Description

Plaease add such a passthrough option also to the button widget:
![2024-04-17 08_35_41-Node-RED _ Diagramm – Mozilla Firefox](https://github.com/FlowFuse/node-red-dashboard/assets/31366299/ef7f8ad2-a158-48f4-b734-38bd002c5635)

### Have you provided an initial effort estimate for this issue?

I am no FlowFuse team member

colinl:
Is issue #209 actually what you are looking for?

joepavitt:
Closing as duplicate of #209 - will prioritise as now requested by more than one person.

joepavitt:
hmm, actually they're

colinl:
I can't see the point in just passing through.  If that is required then just wire round the button.

m-schaeffler:
yes, emulate a click fulfills my needs

PR #783, opened 2024-04-18
UI Button - Add "Emulate Click" option

### LINKED ISSUE #209: Button widget: Option missing - 'If msg arrives on input, emulate a button click'
As with D1, the Button widget should have an option to emulate a button click if a message is passed in.  It should then send the received message, but with the configured topic and payload.
However, if the message includes msg.enabled then this should not emulate a click.

joepavitt:
## Description

- Adds "emulateClick" option
- Adds E2E tests to cover this
- Adds new option to the `checkOutput` cypress command to cover `not.eq` and others too.

## Related Issue(s)

Closes #209
May cover #777 too

### DIFF
--- a/cypress/fixtures/flows/dashboard-buttons.json
+++ b/cypress/fixtures/flows/dashboard-buttons.json
@@ -1,50 +1,5 @@
 [
     {
-        "id": "node-red-tab-buttons",
-        "type": "tab",
-        "label": "UI Buttons",
-        "disabled": false,
-        "info": "",
-        "env": []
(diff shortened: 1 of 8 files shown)


joepavitt:
> Slight caveat, I believe it is referred to as pass thru on DB1.

The variable in D1 was `passthru`, but all terminology in the editor was "Emulate button click". I also want to be sure it's not referred to as "pass through", because it doesn't, it modifies the `msg` object before sending on from the node

Issue #784, opened 2024-04-18
WebSocket Connection Issues


joepavitt:
### Description

Using this as a centralised store for websocket connection-related problems.

- [ ] https://github.com/FlowFuse/node-red-dashboard/issues/620
- [ ] https://github.com/FlowFuse/node-red-dashboard/issues/444
- [ ] https://github.com/FlowFuse/node-red-dashboard/issues/619

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
When we lost connection between the client and server, we don't make any attempts to re-connect, nor do we inform the user of a lost connection.

omrid01:
Adding a scenario - when the user deploys or restarts the flows on the editor, the dashboard clients realize it and try to re-render, but the 'onSocket' listeners on the server node stop receiving custom client messages. reloading the client browser (F5) fixes the issue.
The temp work around I currently implement, is upon startup, the server node waits 200 ms and then sends a special message to all clients which tells them to reload themselves.

Paul-Reed:
I'm assuming this is the same issue, but thought you may wish to consider PWA when fixing.

Saved as an App (PWA) on a android phone, the data stream is updated every 5 seconds.
At 7 seconds in the video, I restart node-RED which produces a 'connection lost' message as expected.
When it 'reconnects' at 12 seconds, all of the values then show zero, and remain at zero, despite node-RED updating normally (I'm also watching the dashboard on a laptop)

https://github.com/FlowFuse/node-red-dashboard/assets/973580/22ae5cad-54bc-47c0-8913-55b7c4378103

colinl:
@joepavitt do you want ongoing issues over connections in here, or in individual issues?  I would vote for putting everything related here as several of the issues are related.  Except maybe the ui-template issue which it appears is not directly related.

I find that if, on a PC, I open the dashboard and disconnect the network and leave it for 30 mins (not sure what the min to show the problem is) then it never reconnects when I connect the network again, it just sits there with Connection Lost popup.  I see the same on Android sometimes when I re-open the dashboard window after it has been in the background for some time.  I suspect it is the same issue.

joepavitt:
Each in a new issue, them link them here so I can add to the list please.

I need to double check, but I believe that's expected. @Steve-Mcl recommended I add a "give up trying to connect" timeout, and I believe I set it to 30 minutes

joepavitt:
[...]

PR #790, opened 2024-04-19
SocketIO - Improve reconnection behaviour of SocketIO client

joepavitt:
## Description

- This sets up an auto-reconnect every one second when a client loses connection to the Node-RED server.
- We add a `ui-notification` which is called on disconnect and reconnect events to inform the front-end user.
    - Disconnect events are not auto-closed, nor closable, if a disconnect has happened, it is important the user knows.
    - Connection established notifications are closable, and auto-timeout after 3 seconds

## Related Issue(s)

Likely to fix the issues linked in #784

### DIFF
--- a/ui/src/main.mjs
+++ b/ui/src/main.mjs
@@ -86,8 +86,10 @@ fetch('_setup')
 
             // handle final disconnection
             socket.on('disconnect', (reason) => {
-                disconnected = true
-                disconnectedAt = new Date()
+                if (!disconnected) {
+                    disconnectedAt = new Date()
+                    disconnected = true
+                }
                 // tell the user we're trying to connect
                 Alerts.emit('Connection Lost', 'Attempting to reconnect to server...', 'red', {
                     displayTime: 0,
--- a/ui/src/main.mjs
+++ b/ui/src/main.mjs
@@ -74,17 +74,57 @@ fetch('_setup')
 
         store.commit('setup/set', setup)
 
-        const socket = io(setup.socketio)
-
         let disconnected = false
         let disconnectedAt = null
 
         const MAX_TIMER = 300000 // 5 minutes
 
+        let socket = null
+
+        function connect () {
+            socket = io(setup.socketio)
+
+            // handle final disconnection
+            socket.on('disconnect', (reason) => {
+                disconnected = true
+                disconnectedAt = new Date()
+                // tell the user we're trying to connect
+                Alerts.emit('Connection Lost', 'Attempting to reconnect to server...', 'red', {
+                    displayTime: 0,
+                    allowDismiss: false,
+                    showCountdown: false
+                })
+                socket.destroy()
+                // attempt to reconnect
+                reconnect()
+            })
+
+            socket.on('connect', () => {
+                console.log('SIO connected')
+                // if we've just disconnected (i.e. aren't connecting for the first time)
+                if (disconnected) {
+                    // send a notification/alert to the user to let them know the connection is live again
+                    Alerts.emit('Connected', 'Connection re-established.', '#1BC318', {
+                        displayTime: 3,
+                        allowDismiss: true,
+                        showCountdown: true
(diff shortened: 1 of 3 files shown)


joepavitt:
fyi @colinl

ui/src/main.mjs
```diff
@@ -73,13 +76,43 @@ fetch('_setup')
 
         const socket = io(setup.socketio)
 
+        let disconnected = false
+
+        function reconnect (interval = 1000) {
```
Steve-Mcl, line 81: is 1s not a bit aggressive Joe? And indefinitely? I think there should be some eventual circuit breaker where it stops trying and puts the ball into the users hands (i.e. click a retry button)

Thinking about the issue we had recently where an editor was left open and it kept trying to reach a deleted/suspended instance indefinitely (didnt it cause an alert?)

In these situations, it is V common to see an exponential back-off and a circuit breaker.

colinl:
Are you sure it does not reconnect? I am using an event node and it generates an event when the connection is re established.

joepavitt:
> Are you sure it does not reconnect? I am using an event node and it generates an event when the connection is re established.

That seemed to be the problem I could see, the connection times out. When I added this back in, the connection worked and events were being received as expected.

ui/src/main.mjs
joepavitt: It's a fair point, happy to go smaller and exponentially increase until limit is reached

colinl:
Have you looked at how D1 does it, or is that not Relevant?

ui/src/main.mjs
joepavitt: Have updated

joepavitt:
@Steve-Mcl fyi - ready for re-review, have overhauled the reconnect logic to `socket.destroy()` on error and disconnect, then implementing/relying on our own `reconnect()` logic.

We need to `destroy()` otherwise we end up up new event handlers every time we try to reconnect

ui/src/main.mjs
```diff
@@ -71,20 +74,69 @@ fetch('_setup')
...
+
+        let socket = null
+
+        function connect () {
+            socket = io(setup.socketio)
+
+            // handle final disconnection
+            socket.on('disconnect', (reason) => {
+                disconnected = true
+                disconnectedAt = new Date()
```
Steve-Mcl, line 90:
```suggestion
                if (!disconnected) {
                    disconnectedAt = new Date()
                    disconnected = true
                }
```

Recommendation: prevent `disconnectedAt` being updated if already disconnected. It is an unlikely scenario but in the event of a bug, we wont update the disconnection timestamp and the circuit breaker logic will work as designed.

ui/src/main.mjs
```diff
@@ -71,20 +74,69 @@ fetch('_setup')
...
+                // if we've just disconnected (i.e. aren't connecting for the first time)
+                if (disconnected) {
+                    // send a notification/alert to the user to let them know the connection is live again
+                    Alerts.emit('Connected', 'Connection re-established.', '#1BC318', {
+                        displayTime: 3,
+                        allowDismiss: true,
+                        showCountdown: true
+                    })
+                }
+                disconnected = false
```
Steve-Mcl, line 113: Recommendation: Here, we should probably call `clearTimeout(...)` to halt any ongoing timeout. Not mandatory since you do test the `disconnected` boolean. (good practice)

joepavitt: I shall make it so, thanks Steve

joepavitt:
Good to review again @Steve-Mcl

Steve-Mcl (approved):

Steve-Mcl:
ui test fail is the oft unreliable  `widgets/control.spec.js` test. Happy to merge @joepavitt ?

joepavitt:
Yeah, my attempt at making them more _reliable_ seems to have not worked too well

PR #808, opened 2024-04-26
UI Gauge - Fix the re-ordering of the gauge segments in the NR Editor

### LINKED ISSUE #558: UI Gauge - cannot re-order segments
Just noticed that the limit segments cannot be re-ordered by dragging the hamburgers.
I assume that it should be able to do so?

move

DB v1.1.0

joepavitt:
## Description

Fixes the typo from the copy/pasted code (from `ui-dropdown`) which meant the sorting selector wasn't properly enabled.

## Related Issue(s)

Close #558

### DIFF
--- a/nodes/widgets/ui_gauge.html
+++ b/nodes/widgets/ui_gauge.html
@@ -155,7 +155,7 @@
 
             $('#node-input-segments-container').sortable({
                 axis: 'y',
-                handle: '.node-input-option-handle',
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):

Steve-Mcl:
e2e fail is the usual ` widgets/control.spec.js ` - others pass ok.

PR #825, opened 2024-05-07
Add @flowfuse/flow-renderer & worldmap example

joepavitt:
## Description

- Adds a "worldmap" example to "UI Template Examples", based on recent support ticket request
- Uses the new flow-renderer to include our first explicit example to the docs with downloadable flows


docs/user/template-examples.md
```diff
@@ -2,6 +2,17 @@
 description: Get inspired with a variety of UI template examples in Node-RED Dashboard 2.0, enhancing your dashboard design process.
 ---
 
+
+<script setup>
```
ZJvandeWeg, line 6: I'll merge, but there might be a nice trick with putting the mapping from worldmap to ExampleFlowWorldmap in the frontmatter instead?

ZJvandeWeg (approved):

docs/user/template-examples.md
joepavitt: This is frontmatter essentially, but allows me to write JS, instead of formatted yml

PR #861, opened 2024-05-15
UI Slider - Dynamic Properties

### LINKED ISSUE #860: Dynamic Properties: UI Slider
```[tasklist]
### Properties
- [x] Label
- [x] Thumb Label
- [x] Range
- [x] Payload
```

joepavitt:
## Description

- Adds dynamic properties for `label`, `thumbLabel`, `min`, `step`, `max`
- Updates documentation
- Adds E2E test to cover `label`, `min` and `max` update, testing the step size

## Related Issue(s)

Closes #860


Steve-Mcl (changes requested):
### comparisons to `undefined`
I recommend the use of `typeof x !== 'undefined'` - it is safer and is best practice

### Fallbacks dont consider user might want empty string labels
The code uses this pattern  `return this.dynamic.xxx || this.props.xxx`. This will not permit a user to set an empty string e.g. a label of `""` is falsey and will default to `props.label`

### question
The old Node-RED pattern for dynamic variables is to be provided invisibly in a `msg` object had unwritten rules in that if the property has a value, the `msg` property is discarded (does not override the entered value)
The newer Node-RED pattern is to provide a visible means for setting dynamic properties via the typedInput 
Neither of these patterns are observed in this PR - can you please explain?

nodes/widgets/locales/en-US/ui_slider.html
```diff
@@ -8,6 +8,26 @@
     <p>The slider can be reversed by setting the min value larger than the max value. e.g. min 100, max 0.</p>
     <h3>Dynamic Properties (Inputs)</h3>
     <p>Any of the following can be appended to a <code>msg.</code> in order to override or set properties on this node at runtime.</p>
```
Steve-Mcl, line 10:
```suggestion
    <p>Any of the following can be appended to a <code>msg.ui_update.</code> in order to override or set properties on this node at runtime.</p>
```

ui/src/widgets/ui-slider/UISlider.vue
```diff
@@ -63,6 +86,27 @@ export default {
...
+            const updates = msg.ui_update
+            if (!updates) {
+                return
+            }
+            if (updates.label) {
+                this.dynamic.label = updates.label
+            }
+            if (updates.thumbLabel) {
+                this.dynamic.thumbLabel = updates.thumbLabel
+            }
```
Steve-Mcl, line 100: checking `truthy` will prevent anyone clearing the dynamic value or setting an empty string.

I suggest a set of basic rules are adhered to across the board in all UI nodes so that there is one rule for users to get used to. 

Common tasks users do in dashboard:
1. Set an override - OK - possible
2. Set an empty string (e.g. no label) -  Not possible in this implementation (sending `""` would be ignored)
3. Clear an override - Not possible in this implementation

I would suggest, much like context, we differentiate missing property from `null`/`undefined` etc

The rule could be:
1. Exclude the property to skip updaing it (e.g. `ui_control: {}`)
2. Include the property to update it  (e.g. `ui_control: { label: xxx }`)
    1. `null` means remove any override
    2. anything else means use the override 

e.g. Take the case of setting a `label` 
* user sends `ui_control: { min: 7 }` - `label` is missing, dont update the dynamic label value!
* user sends `ui_control: { label: null }` - `label` is clearly present, use `null` to remove the dynamic property (and thus fall back to pre-set value)
* user sends `ui_control: { label: "" }` - `min` is clearly present, use its value (even if it is an empty string) to update the dynamic property

Steve-Mcl: in https://github.com/FlowFuse/node-red-dashboard/pull/864#discussion_r1618610128 you use !== undefined. Which means it would work as I described here:



> The rule could be:
> 
> 1. Exclude the property to skip updaing it (e.g. `ui_control: {}`)
> 2. Include the property to update it  (e.g. `ui_control: { label: xxx }`)
>    
>    1. `null` means remove any override
>    2. anything else means use the override

So my suggestion becomes:


```suggestion
            if (typeof updates.label !== 'undefined') {
                this.dynamic.label = updates.label
            }
            if (typeof updates.thumbLabel !== 'undefined') {
                this.dynamic.thumbLabel = updates.thumbLabel
            }
```

This way, since `null` already means "dont user the dynamic override", we can permit the user to send `""` for a deliberately empty label and `null` to undo the dynamic override.

ui/src/widgets/ui-slider/UISlider.vue
```diff
@@ -63,6 +86,27 @@ export default {
...
+            if (updates.min !== undefined) {
+                this.dynamic.min = updates.min
+            }
+            if (updates.max !== undefined) {
+                this.dynamic.max = updates.max
+            }
+            if (updates.step !== undefined) {
+                this.dynamic.step = updates.step
+            }
```
Steve-Mcl, line 109:
```suggestion
            if (typeof updates.min !== 'undefined') {
                this.dynamic.min = updates.min
            }
            if (typeof updates.max !== 'undefined') {
                this.dynamic.max = updates.max
            }
            if (typeof updates.step !== 'undefined') {
                this.dynamic.step = updates.step
            }
```

at minimum, best practice, though for number fields like these, I'd be tempted to ensure the type is `'number'` or to attempt to parse the value to a number (to avoid the inevitable forum posts about sending "12" and it didnt work as expected)

nodes/widgets/ui_slider.js
```diff
@@ -14,11 +16,35 @@ module.exports = function (RED) {
...
+                /**
+                 * Dynamic Properties
+                 * */
+                const updates = msg.ui_update
+                if (updates) {
+                    if (typeof (updates.label) !== 'undefined') {
+                        // dynamically set "label" property
+                        statestore.set(group.getBase(), node, msg, 'label', updates.label)
+                    }
+                    if (updates.thumbLabel) {
```
joepavitt, line 36:
```suggestion
                    if (typeof (updates.thumbLabel) !== 'undefined') {
```

nodes/widgets/ui_slider.js
```diff
@@ -14,11 +16,35 @@ module.exports = function (RED) {
         const evts = {
             onChange: true,
             beforeSend: function (msg) {
-                if (!node.pt) {
-                    node.state[0] = msg.payload
-                    node.status({ shape: 'dot', fill: 'grey', text: node.state[0] + ' | ' + node.state[1] })
-                } else if (node._wireCount === 0) {
-                    node.status({ shape: 'dot', fill: 'grey', text: msg.payload })
+                if (msg.payload !== undefined) {
```
Steve-Mcl, line 19:
```suggestion
                if (typeof msg.payload !== 'undefined') {
```

ui/src/widgets/ui-slider/UISlider.vue
```diff
@@ -21,28 +21,51 @@ export default {
...
+            return this.dynamic.max !== null ? this.dynamic.max : this.props.max
         }
     },
     watch: {
         storeValue: function (val, oldVal) {
             if (this.value === val) {
                 return // no change
             }
-            this.value = val
+            if (val !== undefined) {
```
Steve-Mcl, line 61:
```suggestion
            if (typeof val !== 'undefined') {
```

Steve-Mcl (approved):
approved with couple of best practice change requests

PR #864, opened 2024-05-15
UI Button - Add dynamic properties for label, icon & iconPosition

### LINKED ISSUE #862: Dynamic Properties: UI Button
```[tasklist]
### Properties
- [x] Label
- [x] Icon
- [x] Icon Position
```

joepavitt:
## Description

- Adds `label`, `icon` and `iconPosition` dynamic properties
- Adds E2E coverage
- Updates documentation

## Related Issue(s)

Close #862

### DIFF
--- a/nodes/widgets/locales/en-US/ui_button.html
+++ b/nodes/widgets/locales/en-US/ui_button.html
@@ -20,6 +20,10 @@
     <dl class="message-properties">
         <dt class="optional">label <span class="property-type">string</span></dt>
         <dd>Override the label displayed on the button</dd>
+        <dt class="optional">icon <span class="property-type">string</span></dt>
+        <dd>Override the icon defined in the initial configuration</dd>
+        <dt class="optional">iconPosition <span class="property-type">'left' | 'right'</span></dt>
+        <dd>Change which side of the label the icon renders</dd>
         <dt class="optional">class <span class="property-type">string</span></dt>
         <dd>Add a CSS class, or more, to the Button at runtime.</dd>
     </dl>
--- a/nodes/widgets/locales/en-US/ui_button.html
+++ b/nodes/widgets/locales/en-US/ui_button.html
@@ -16,7 +16,7 @@
         <dd>If enabled, any received message will trigger a button click, emitting the relevant payload and topic</dd>
     </dl>
     <h3>Dynamic Properties (Inputs)</h3>
-    <p>Any of the following can be appended to a <code>msg.</code> in order to override or set properties on this node at runtime.</p>
+    <p>Any of the following can be appended to a <code>msg.ui_update</code> in order to override or set properties on this node at runtime.</p>
     <dl class="message-properties">
         <dt class="optional">label <span class="property-type">string</span></dt>
         <dd>Override the label displayed on the button</dd>
--- a/nodes/widgets/locales/en-US/ui_button.html
+++ b/nodes/widgets/locales/en-US/ui_button.html
@@ -18,6 +18,8 @@
     <h3>Dynamic Properties (Inputs)</h3>
     <p>Any of the following can be appended to a <code>msg.</code> in order to override or set properties on this node at runtime.</p>
     <dl class="message-properties">
+        <dt class="optional">label <span class="property-type">string</span></dt>
+        <dd>Override the label displayed on the button</dd>
         <dt class="optional">class <span class="property-type">string</span></dt>
         <dd>Add a CSS class, or more, to the Button at runtime.</dd>
     </dl>
--- a/ui/src/widgets/ui-button/UIButton.vue
+++ b/ui/src/widgets/ui-button/UIButton.vue
@@ -79,14 +79,18 @@ export default {
             return 'mdi-' + icon.replace(/^mdi-/, '')
         },
         onDynamicProperties (msg) {
-            if (msg.label !== undefined) {
-                this.dynamic.label = msg.label
+            const updates = msg.ui_update
+            if (!updates) {
+                return
             }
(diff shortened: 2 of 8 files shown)


Steve-Mcl (changes requested):
### comparisons to `undefined`
I recommend the use of `typeof x !== 'undefined'` - it is safer and is best practice

### Fallbacks dont consider user might want empty string labels
The code uses this pattern  `return this.dynamic.xxx || this.props.xxx` and `if (msg.xxx) { /* dynamic xxx, update the store */ }`. 
These will not permit a user to set an empty string e.g. a label of `""` is falsey and will default to `props.label`

### question
The old Node-RED pattern for dynamic variables is to provide them invisibly in  `msg` object properties. The unwritten rule was that that if the property has a value, the magic `msg` property is discarded (i.e. does not override the entered value)

The newer Node-RED pattern is to provide a visible means for setting dynamic properties via the `typedInput` control 

Neither of these patterns are observed in this PR - can you please explain?

joepavitt:
> The old Node-RED pattern for dynamic variables is to provide them invisibly in msg object properties. The unwritten rule was that that if the property has a value, the magic msg property is discarded (i.e. does not override the entered value)

I wasn't familiar with that, as far as I'm concerned, if a user has injected an override, it should override.

joepavitt:
> Comparisons to undefined
I recommend the use of typeof x !== 'undefined' - it is safer and is best practice

> Fallbacks dont consider user might want empty string labels
The code uses this pattern return this.dynamic.xxx || this.props.xxx and if (msg.xxx) { /* dynamic xxx, update the store */ }.
These will not permit a user to set an empty string e.g. a label of "" is falsey and will default to props.label

Addressed in 1feb92a8

Steve-Mcl:
> > The old Node-RED pattern for dynamic variables is to provide them invisibly in msg object properties. The unwritten rule was that that if the property has a value, the magic msg property is discarded (i.e. does not override the entered value)
> 
> I wasn't familiar with that, as far as I'm concerned, if a user has injected an override, it should override.

https://discourse.nodered.org/t/dynamic-input-for-range-node/3091/4?u=steve-mcl 


There are better explanations that the linked one above but I am fairly certain the reason was that it prevents stray properties a msg has picked up overriding "hard coded values a user has entered".  

As an alternative option, you may wish to consider a pattern that is less likely to fall foul of the above. Obviously that would be to use the typedInputs as per intended design but I understand the reluctance. Perhaps a more specific/less-common `msg` property specifically for property overrides could be employed by the `ui` nodes?  Something like `msg.ui.override.xxx` or `msg.uiProps.xxx` - something that is not so easy to be picked up by a message like `label`.  Alternatively, support a distinct but well formed message type with a specific topic "set-label" and `payload` containing the dynamic value.  This way, the widgets value part (what the payload normally used for) is NOT REQUIRED (i.e no need to delete msg.payload for only updating the label) 

These are just a couple of suggestions, it is, of course, your choice. Happy to discuss merits of alternative approaches that keep it simple but avoid pitfalls discussed above.

nodes/widgets/locales/en-US/ui_button.html
```diff
@@ -18,6 +18,12 @@ <h3>Properties</h3>
     <h3>Dynamic Properties (Inputs)</h3>
     <p>Any of the following can be appended to a <code>msg.</code> in order to override or set properties on this node at runtime.</p>
     <dl class="message-properties">
+        <dt class="optional">label <span class="property-type">string</span></dt>
+        <dd>Override the label displayed on the button</dd>
+        <dt class="optional">icon <span class="property-type">string</span></dt>
+        <dd>Override the icon defined in the initial configuration</dd>
+        <dt class="optional">iconPosition <span class="property-type">'left' | 'right'</span></dt>
+        <dd>Change which side of the label the icon renders</dd>
```
Steve-Mcl, line 26: No mention of `msg.ui_control` - users will likely stumble here Joe.

In https://github.com/FlowFuse/node-red-dashboard/pull/861 you have added a comment above the dynamic properties to clarify this - see here: https://github.com/FlowFuse/node-red-dashboard/pull/861#discussion_r1618566423

nodes/widgets/locales/en-US/ui_button.html
```diff
@@ -18,6 +18,12 @@ <h3>Properties</h3>
     <h3>Dynamic Properties (Inputs)</h3>
     <p>Any of the following can be appended to a <code>msg.</code> in order to override or set properties on this node at runtime.</p>
```
Steve-Mcl, line 19:
```suggestion
    <p>Any of the following can be appended to a <code>msg.ui_update</code> in order to override or set properties on this node at runtime.</p>
```

ui/src/widgets/ui-button/UIButton.vue
```diff
@@ -59,6 +77,24 @@ export default {
...
+            }
+            if (updates.label !== undefined) {
+                this.dynamic.label = updates.label
+            }
+            if (updates.icon !== undefined) {
+                this.dynamic.icon = updates.icon
+            }
+            if (updates.iconPosition !== undefined) {
+                this.dynamic.iconPosition = updates.iconPosition
+            }
```
Steve-Mcl, line 94:
```suggestion
            if (typeof updates.label !== 'undefined') {
                this.dynamic.label = updates.label
            }
            if (typeof updates.icon !== 'undefined') {
                this.dynamic.icon = updates.icon
            }
            if (typeof updates.iconPosition !== 'undefined') {
                this.dynamic.iconPosition = updates.iconPosition
            }
```

best practice.

Steve-Mcl (approved):

PR #873, opened 2024-05-17
Adds in richer configuration of UI Table columns

joepavitt:
## Description

- Adds new configuration options on a column:
    - Width
    - Align
    - Type
- Updated the UI Data Table vue component to support the following cell types:
    - Text
    - HTML
    - Link
    - Progress
    - Color
    - Tick/Cross
    - Sparkline - Trend
    - Sparkline - Bar
    - Row Number
- Adds in-Editor documentation for UI Table
- Updates the online documentation for UI Table, including examples

### Example

Screenshot 2024-05-17 at 18 27 01

## Related Issue(s)

Closes #392

PR #875, opened 2024-05-19
Support nested properties in new UI Table

joepavitt:
## Description

Implements feature request to support nested properties for #873 

Did this as a standalone PR as it required some re-wiring of how we pass properties into the `UITableCell` component, so wanted to make it easier to review


Steve-Mcl (approved):

Steve-Mcl:
I believe the e2e table test failures are a result of the ongoing work in #873 - feel free to merge if this is correct.

Issue #589, opened 2024-02-18
Informational question about (web) push notifications


bartbutenaers:
Hi,

At the time being, I needed push notification to my old AngularJs dashboard.  So I developed the [node-red-contrib-ui-web-push](https://github.com/bartbutenaers/node-red-contrib-ui-web-push) node, that allowed me to ***web push*** notifications to my Android smartphone.  And that still works very well.  When I developed that node, Apple didn't want to support web push notifications but seems that they meanwhile have allowed it.  So it has now become a technology that is supported on quite a number of platforms.

Note that I have never published that node on npm, since it is rather time consuming to troubleshoot/debug the client side code.  And I don't have enough free time to do that for everybody who has problems with this node on his phone...

Before I can switch my personal dashboard to D2, I need to have an alternative.  So - if I ever find time - I could migrate my ui node to D2.  But I was wondering if there are perhaps already plans for web push support?  I am not familiar with Vue, so I have no idea at all whether something like that is perhaps already supported by Vue out-of-the-box?  Or if you can use normal (non-web) push notifications with Vue, like with all other apps?  Because in that case it would be silly if I spend my precious time on migrating my old node.

Hopefully somebody can illuminate me...

P.S. This discussion is only about full integration of (web) push notifications in dashboard D2, ***not*** about external push technologies (like Telegram, ...).   

Thanks!!
Bart

joepavitt:
This isn't something I've had to cover before, so isn't my area of expertise unfortunately, although worth stressing, Vue stuff doesn't have to be Vue-specific, you can still write raw JavaScript content and it'll be fine. 

@totallyinformation may have covered similar in UI Builder?

joepavitt:
https://web.dev/articles/push-notifications-overview 

This gives a great overview, and does reinforce that we need a supporting "push service" to get the full pipeline working.

bartbutenaers:
Currently my [node-red-dashboard-2-ui-web-push](https://github.com/bartbutenaers/node-red-dashboard-2-ui-web-push) is nearly finished.   Just need to find time to fix the last bits and pieces.  But most of it already works fine.

[...]

PR #881, opened 2024-05-21
PWA Update to Allow Customizing Service Worker, Fix for PWA Blank Screen on Re-authentication

### LINKED ISSUE #817: PWA feature - problem with Google Oauth authentication (and probably other similar systems)
### Current Behavior

~~I am not 100% certain that the below description is completely correct, but I am fairly sure it is close.~~
I use Cloudflare ZeroTrust with Google Oauth authentication to access the editor at nodered.mydomain.com, and nodered.mydomain.com/dashboard to access the dashboard from Android.

Before the PWA capability, if re-authentication was required, reloading the dashboard page (by swiping down in Android) attempted to reload the dashboard and I was redirected to the Google login screen by Cloudflare so I could login and then the dashboard loaded.
Now, with the PWA version, if the authentication has expired then it appears that I just get a blank screen and there does not seem to be a way of forcing a reload so I am stuck.

To recover the situation I have to browse to the node-red root url (not the dashboard) and then I get the login screen before the editor opens. Then if I go back to the dashboard application it opens ok.  Currently there is no way of adding a shortcut to the dashboard to the home screen in Android, other than via a PWA application.

This has been discussed in PR #708 and in forum post [Add PWA Functionality by cgjgh · Pull Request #708 · FlowFuse/node-red-dashboard · GitHub](https://github.com/FlowFuse/node-red-dashboard/pull/708#issuecomment-2082915609) where @TotallyInformation suggested:
> It sounds like the client code needs to have something that picks up a 401 response code from a client request and triggers a standard redirect to an endpoint defined in the D2 config.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- Dashboard version: 1.8.1
- Node-RED version: 4.0.0-Beta.2
- Node.js version:
- npm version:
- Platform/OS: Server on Ubuntu or Pi OS
- Browser: Chrome and Brave on Android phone

cgjgh:
## Description

- Changed service worker generation strategy to allow customizing service worker for features such as push notifications as referenced in this [issue](https://github.com/FlowFuse/node-red-dashboard/issues/589#issuecomment-2079974279)
- Updated PWA implementation in accordance with latest Vite-PWA documentation and recommended project scaffolding (includes Vue library updates)
- Added PWA related notifications on _new update available_ and _PWA assets download completion_
- Fix for blank screen [issue](https://github.com/FlowFuse/node-red-dashboard/issues/817#issuecomment-2120562888) that occurred in PWA when re-authentication was necessary, due to a combination of caching and an absence of a subsequent action in error handling of D2 setup request.
## Related Issue(s)

Resolves #817 
Related to #589

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


manifest.json
cgjgh, line 1: Aesthetic refactoring

package.json
cgjgh, line 1: - Updated Vue and Vite-PWA libraries in accordance with Vite-PWA recommendations
- Added [Workbox](https://developer.chrome.com/docs/workbox/) dev libraries required for custom service worker generation strategy
- Added [Sharp](https://www.npmjs.com/package/sharp) image generation dev library overrides required for Vite-PWA PWA asset generation

ui/index.html
cgjgh, line 1: - Cleaned up previously commented code
- Removed unnecessary mask icon; only needed in manifest
- Add Cross-Origin safe manifest inclusion

ui/src/App.vue
cgjgh, line 1: - Updated Vite-PWA implementation

ui/src/components/PWABadge.vue
cgjgh, line 1: ### PWA Component

#### Includes all PWA related front-end logic and UI
- Periodic sync check every hour
- PWA related notifications: `New content available, click on reload button to update.`  and `App is offline ready` on post-install

#### Possible TODO:
Use styling from `common.css` instead of scoped styling

ui/src/main.mjs
cgjgh, line 1: - Removed old PWA implementation
- Enabled `no-cors` mode to fetch request, necessary because in certain cases fetch request wouldn't complete also seemingly bypassing error handling which is necessary to fix blank screen issue - might replace with CORS pre-flighting in the future to preserve CORS security
- Added subsequent action in error handling to fix blank screen issue, left comments in for future reference till any potential additional related issues have been worked out

ui/src/sw.js
cgjgh, line 1: - Added service worker base code - any additional features should be added below as needed

vite.config.js
cgjgh, line 1: ### Updated Vite-PWA Implementation

- Changed service worker generation strategy to allow custom service worker
- Changed update mode from `Automatic` to `Prompt` for better transparency when receiving new builds

cgjgh (commented):
@joepavitt Attempted to make the code changes clear; please inform me if additional explanations are required.

joepavitt:
Thanks for the detail @cgjgh - I've assigned @cstns to review this, please leave the merge to me though as we're due to do a release or two over the next couple of days and I need to time the merge accordingly.

cgjgh:
@joepavitt Ok, sounds good.

ui/src/main.mjs
```diff
@@ -72,7 +66,7 @@ const vuetify = createVuetify({
 // if our scoket disconnects, we should inform the user when it reconnects
 
 // GET our SocketIO Config from Node-RED & any other bits plugins have added to the _setup endpoint
-fetch('_setup')
+fetch('_setup', { mode: 'no-cors' })
```
cstns, line 69: I'm not convinced by the no-cors mode because it has multiple security implications. I was able to bypass the blank screen without cors disabled, albeit I only tested on chrome w linux.

The proper solution would be to pass the specific access control origin headers. I'm not sure if we have access to that information when this call is being made. @joepavitt could weigh in on this

ui/src/main.mjs
cstns: cors errors are supposed to evade script error handling by design. It's a pain to work with, but would allow attackers to learn details network topology, so they're treated as network errors:

"Fetch will return a [network error](https://fetch.spec.whatwg.org/#concept-network-error) if the request is not made to a same-origin URL."

ui/src/components/PWABadge.vue
```diff
@@ -0,0 +1,118 @@
...
+                Reload
+            </button>
+            <button type="button" @click="close">
+                Close
+            </button>
+        </div>
+    </div>
+</template>
+
+<style scoped>
```
cstns, line 87: a lang=scss attribute should be added here

ui/src/components/PWABadge.vue
```diff
@@ -0,0 +1,118 @@
+<script setup>
```
cstns, line 1: The only thing that would need changing here is to convert this template to an options format instead of a composition api as that is the general format agreed upon in this repository. 

I'd be more than happy to lend a hand if needed!

cstns (changes requested):
Great work on this PR! apart from the cors related concerns and minor changes we're there looks good to me!

ui/src/main.mjs
joepavitt: Yeah, `no-cors` _shouldn't_ be necessary here because the `_setup` call is always called to the same host?

cstns: After a chat with @joepavitt , it is my understanding that this _setup call will always be same origin as the NR instance serves the dashboard via an express server.

Cors might become an issue after adding the ability to host the dashboard on a separate domain than the NR instance, but we can mark that as a problem for future selves because it will bring up another host related of problems

cgjgh: Does the same hold true if your auth service is on same host but different port and/or subdomain?

Quote from #817: 

> I use Authentik, a self-hosted auth provider with Google OAUTH as a social login option in front of NodeRed. In Edge on Windows I tried to replicate this issue by removing session cookies from Dashboard in an attempt to emulate session expiration. It did indeed act in the manner you described: blank screen on Dashboard (reloading dashboard didn't work), but navigating to NodeRed editor redirected to OAUTH login as it should. After logging in, Dashboard worked again.
> 
> Dev console in Edge showed the following error when Dashboard blank screen appeared: `Access to fetch at 'https://example_authentik_oauth_server.com' (redirected from 'https://example_nodered_instance.com/dashboard/_setup') from origin 'https://example_nodered_instance.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled. `
>

cgjgh: This may be an isolated scenario with my deployment, we could remove `no-cors` and see if issue persists.

colinl: I have removed the no-cors option from that line on the version I am testing to see if it is still ok. I will have to wait for my google auth to expire before I will know whether it still redirects to google ok.

cstns: Thank you!

I have a feeling that the addition of the [Workbox](https://developer.chrome.com/docs/workbox/) lib and it's caching strategies is doing most of the heavy lifting when it comes to resolving auth cache problems but didn't get a chance to properly test it because the hole setup is quite extensive and in the small setup I was able to test it in seemed to work just fine.

Looking forward to your feedback!

cgjgh: It appears that the CORS error no longer occurs when I conduct the same test without using the `no-cors` option. Although I’m not entirely sure what has changed, the test now works correctly. 

If @colinl doesn’t encounter any issues with his test, I believe we can safely proceed to remove the `no-cors` option without any adverse consequences.

cstns: Glad to hear! 

> If @colinl doesn’t encounter any issues with his test, I believe we can safely proceed to remove the no-cors option without any adverse consequences.
Definately!

ui/src/main.mjs
```diff
@@ -175,4 +169,9 @@ fetch('_setup')
     })
     .catch((err) => {
         console.log('auth error:', err)
+        console.log('redirecting to:', window.location.origin + '/dashboard')
+
+        // window.location.replace(window.location.origin + '/dashboard') //Original, seems to have no issues with Edge and Chrome on Windows, doesn't work on Android (Not tested on Linux browser)
+        // window.location.href = window.location.origin + window.location.pathname + window.location.search + (window.location.search ? '&' : '?') + 'reloadTime=' + Date.now().toString() + window.location.hash; // Also works on Edge + Chrome on windows, doesn't work on android
+        window.location.replace(window.location.origin + '/dashboard' + '?' + 'reloadTime=' + Date.now().toString() + Math.random()) // Seems to work on Edge and Chrome on Windows Chromium and Firefox on Linux, and also on Chrome Android (an also as PWA App)
```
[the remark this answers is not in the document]
cgjgh: @cstns  The error that occurs for me with `no-cors` at the following line is this:
```javascript
const r = new URL(e.url).pathname.replace("/_setup", "")
``` 

```
TypeError: Failed to construct 'URL': Invalid URL at https://nodered.mydomain.com/dashboard/assets/index-Cl3n5Iiw.js:239:407585
message: "Failed to construct 'URL': Invalid URL"
``` 

### Without `no-cors` it occurs at:

```javascript
fetch('_setup')
``` 

```
TypeError: Failed to fetch at https://nodered2.cgiothub.ddnsfree.com:1020/dashboard/assets/index-DvjY5ADf.js:239:425401
message: "Failed to fetch"
```

Would you suggest something like this in error handler: 


```javascript
if (error instanceof TypeError) {
    if (error.message === "Failed to construct 'URL': Invalid URL") {
      // handle specific "Invalid URL" error
      console.error("Invalid URL error occurred:", error);
    } else if (error.message === "Failed to fetch") {
      // handle specific "Failed to fetch" error
      console.error("Failed to fetch error occurred:", error);
    }
  } else {
    // handle general errors here
    console.error("An error occurred:", error);
  }
```

ui/src/main.mjs
arturv2000: Also tried without the `no-cors` line, and using the same setup as last week I am able to login in Windows (Edge) and Android (Chrome + PWA)

ui/src/main.mjs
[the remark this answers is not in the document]
cstns: Before making any changes to the catch redirect, please try another thing. 

Based on the things you mentioned, I have a feeling the whole setup might be failing because we're receiving an invalid response.url from the _setup call which causes `ui/src/main.mjs:71` to fail.

Would you be able to log the response and check if there's anything out of the ordinary with it?

I assume that due to what might be a failing cors request we might be receiving something we're not expecting on the response and passing it to the URL constructor.

ui/src/main.mjs
cstns: > Also tried without the `no-cors` line, and using the same setup as last week I am able to login in Windows (Edge) and Android (Chrome + PWA)

That's very good news!

colinl: On one of my Android devices and on Edge on the laptop it redirected to the google auth page and logged in ok.  The second Android device had the dashboard in the background. When I pulled it forward it showed the Loading screen with the array of flashing dots and stuck there.  I haven't seen that with the cors option included, but maybe it would happen occasionally with it, I don't know.  Pulling down to force a reload then redirected ok.  So at the moment I am not sure what to make of that.

ui/src/main.mjs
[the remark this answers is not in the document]
cgjgh: @cstns 
### With `no-cors`:

`Response {type: 'opaque', url: '', redirected: false, status: 0, ok: false, …}`

So, url is an empty string causing the error it seems.

### Without `no-cors`:

I do get this error again:

`
Access to fetch at 'https://auth.mydomain.com' (redirected from 'https://nodered.mydomain.com/dashboard/_setup') from origin 'https://nodered.mydomain.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
`
but the error handler still catches the error this time resulting in the desired behavior of redirecting to dashboard.

Error being caught here is: 
`TypeError: Failed to fetch at https://nodered.mydomain.com/dashboard/assets/index-DvjY5ADf.js:239:425401
message: "Failed to fetch"`

Regardless of whether the `no-cors` setting is enabled or not, it seems to be functioning correctly. My thinking would be to implement conditional logic to accommodate both cases, ensuring redirection is handled appropriately in either scenario.

ui/src/main.mjs
cgjgh: At any rate, in the interest of caution, I'd suggest removing the `no-cors` option and expand our testing to a larger audience to determine if any problems continue to occur.

ui/src/components/PWABadge.vue
cgjgh: I do not quite understand this. I tried adding the attribute but got the following error on build:
`Error: [vite-plugin-pwa:build] Preprocessor dependency "sass" not found. Did you install it? Try 'npm install -D sass.'`

From what I can observe, it functions properly in its absence, and the background variable appears to be integrating as expected. Could you provide further details?

ui/src/components/PWABadge.vue
cgjgh: @cstns I have incorporated the converted component into a new file named `PWABadgeOpt.vue`. Could you please review it and verify its accuracy against the original component to confirm that the conversion was executed correctly?

I will replace with original afterwards.

ui/src/components/PWABadge.vue
joepavitt: I agree with @cgjgh - there is n need for `lang=scss` here, it's using standard CSS variables

ui/src/main.mjs
cstns: Agreed!

ui/src/main.mjs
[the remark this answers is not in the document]
cstns: I think it's functioning correctly because of a fluke!

So the flow as I understand it is as follows:
Vue app boots up
 -> makes fetch call to _setup
 -> auth session lost so redirected to auth provider on cross domain
 
 from here we diverge in two sub flows:
 
 1. with cors enabled
  -> the cors error is caught by catch
  -> error is logged, thus the cors error you provided
  -> page location is replaced with cache bust
  -> taken to auth login, all is well
  
  2. without cors i.e. `no-cors`
   -> the cors error returned as a valid response, but not the response we need to complete the setup steps
   -> at line 71 we're passing null to the URL constructor which fails
   -> the error is caught by catch
   -> error logged and page location is replaced with cache bust
   -> taken to auth login
   
The fetch spec seems to confirm this:
image
From what understand from the docs, an [opaque filtered response](https://fetch.spec.whatwg.org/#concept-filtered-response-opaque) for sub-flow 2 means a successful response 
 
I'd treat the cors [network error](https://fetch.spec.whatwg.org/#concept-network-error) out of the two and redirect with cache bust _only_ in that scenario. But I don't have enough experience with the dashboard just yet and the redirect would serve multiple ailments

ui/src/components/PWABadge.vue
cstns: Ill try an review it by EOD!

ui/src/components/PWABadgeOpt.vue
```diff
@@ -0,0 +1,126 @@
+<template>
```
cstns, line 1: Very well done! Can proceed with replacing the composition component!

ui/src/components/PWABadge.vue
cstns: My IDE was complaining about it, can be safely ignored

cstns (approved):

PR #869, opened 2024-05-16
UI Base - Update how we configure client data inclusion and injection of "auth" plugins

joepavitt:
## Description

- Migrate the client constraints/data configuration sidebar from the FF Auth plugin to core
- Introduces socket constraints/control to _all_ widget types, including third-party widgets
- Introduce a "client data providers" list in the sidebar to make it clear what data is appending to `msg._client`
- Update documentation for plugins
- Add documentation for how to use the client configuration (and update the link in the Editor)

## Related Issue(s)

Closes #853 
Related to https://github.com/FlowFuse/node-red-dashboard-ff-auth/pull/14

PR #883, opened 2024-05-22
Docs design patterns

joepavitt:
## Description

Attempt to document the "Design Patterns" for Dashboard 2., most notably:

- Single source of truth (as per Dashboard 1.0)
- Client-driven data (now possible since we introduced https://github.com/FlowFuse/node-red-dashboard/pull/869

I'm not 100% happy with the example here, and think there could be more clarity of the patterns themselves, but I've gone over this over and over for the past few hours, and i think I just need to cut _something_ and get it in.


docs/getting-started.md
```diff
@@ -86,6 +97,78 @@ Built into the framework of the UI is a side navigation bar, along with the top,
...
+<video controls>
+    <source src="./assets/videos/demo-design-patterns.mp4" type="video/mp4">
+    Your browser does not support the video tag.
+</video>
+
+In the video above we see that in some cases, data is sent to just the client that triggered it (e.g. button clicks), and in others, the data is shared across all client sessions (e.g. the visualisation of slider value on the chart).
+
+If you'd like to play with this example, the flow is as follows:
+
+<FlowViewer :flow="examples['design-patterns']" height="425px" style="margin-bottom: 24px;"/>
```
Steve-Mcl, line 153: In testing this, the Click Me button and "Your Latest Button Click:" ui-text was updating on BOTH.  I did not have `ui-text` or `ui-button` [x] checked in the side-bar client data (it worked once i set these).  
Should we be making a reference to that here?

joepavitt: Ah, I think this is probably lost when we do the duplicate `ui-base` check - may need to improve that merger to account for the fact this value will probably differ `ui-base` to `ui-base`, rather than just keeping the _older_ `ui-base` as we currently do.

joepavitt: Issue Opened: https://github.com/FlowFuse/node-red-dashboard/issues/884

PR #894, opened 2024-05-24
Ensure consistency on group ordering

### LINKED ISSUE #892: Fix group ordering
Both editor and client-sider ordering seems broken and inconsistent, need to add tests to check for regressions and ensure this is fixed moving forward.

joepavitt:
## Description

- Ensures groups are ordered client-side, and not dependent upon server-side having ordered in advance of sending via `ui-config`
- Updates the `.order` field on _all_ groups in a `groupsOL` if a new items has been added/removed, or if it's re-sorted.
- Adds E2E coverage to make sure we don't regress on this

## Related Issue(s)

Fixes #892


ui/src/layouts/Notebook.vue
```diff
@@ -64,6 +64,9 @@ export default {
                     }
                     return true
                 })
+                .sort((a, b) => {
```
cstns, line 67: These orderedGroups computed methods look pretty similar. We could extract the logic in a composable/mixin so we don't repeat the code everywhere

joepavitt: Will conduct as a follow on task - don't need it blocking this as is

Issue #769, opened 2024-04-14
Gauge node is not responsive


thebaldgeek:
### Current Behavior

When viewing gauges on mobile, they are too small to read.

Gauges on desktop / laptop:

![image](https://github.com/FlowFuse/node-red-dashboard/assets/7385116/8d15ac75-4d92-4951-a85d-0fb47d8b584c)

(Note, not sure why the top three are super jumbo sized, they are all built the same and no changes made from default - They all should be the same size as the KBOS on on the bottom).

When viewed on Android or iPhone the gauges look like this:

![image](https://github.com/FlowFuse/node-red-dashboard/assets/7385116/3d4429d3-a722-4627-b649-f11dad1f3e2a)

### Expected Behavior

I would expect the gauge to be rendered differently on a small screen, or drop to a text only display of the value.

### Steps To Reproduce

Place two or more gauges on a group.
View on mobile device or F12 -> mobile view on a browser.

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

thebaldgeek:
From a post in the Node-RED Discord.
http://temples.tplinkdns.com:2880/dashboard/current
(Note, the site seems to be down quite often)

Here is how the weather page looks on a Pixel 7 and 8.

![image](https://github.com/FlowFuse/node-red-dashboard/assets/7385116/62e46c52-91f3-463e-8929-b9fb6ce98161)

jtemple967:
@thebaldgeek I think you may have copied/pasted the wrong link. I actually had 2 dashboards available for public viewing for a while, here's the one for the weather:

http://temples.tplinkdns.com:1880/dashboard/current

joepavitt:
Closing this out as I think https://github.com/FlowFuse/node-red-dashboard/pull/979 covers a fix for this. It wasn't an issue with gauges per say, more that our breakpoints on this screen size weren't sensible, so we were trying to squeeze too much content in horizontally.

PR #910, opened 2024-05-29
UI Gauge - Improve text size across different gauge sizes

### LINKED ISSUE #888: UI Gauge - Improve icon sizing at varying sizes
### Current Behavior

https://discourse.nodered.org/t/gauge-icon-text-sizing/88176

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

joepavitt:
## Description

- By no means a "perfect" solution, but it's better than we have.
- This is a complicated problem given the variations available for gauge sizing and types of gauges available.

## Related Issue(s)

Closes #888 

Some impact on #769 too

## Examples

https://github.com/FlowFuse/node-red-dashboard/assets/99246719/0201b3f4-4aad-4347-b0ef-ce090f21f506

### DIFF
--- a/docs/nodes/widgets/ui-gauge.md
+++ b/docs/nodes/widgets/ui-gauge.md
@@ -85,3 +85,12 @@ Values for the gauges can be set by sending a numerical value in `msg.payload`. 
 ![Examples of some Tile Gauges](/images/node-examples/ui-gauge-tiles.png "Examples of some Tile Gauges"){data-zoomable}
 *Examples of some Tile Gauges*
 
+## Overriding CSS
+
+The gauge can be styled further by adding custom CSS to the `ui_template` node. Some useful classes that are available for styling include:
+
+- `.nrdb-ui-gauge-value span` - The value in the middle of the gauge
+- `.nrdb-ui-gauge-value label` - The unit label
+- `.nrdb-ui-gauge-value i` - The icon
+- `.nrdb-ui-gauge-icon-only i` - Available when a gauge has an icon but no "unit" label
+- `.nrdb-ui-gauge #limits` - The containing `<g>` element that wraps min/max `<text>` elements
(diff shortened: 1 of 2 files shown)


cstns (commented):
This appears to be a D3 implementation detail rather than a dashboard-specific one. 

While having an out-of-the-box solution is beneficial and simplifies usage, users should be aware of this and have the ability to override line heights and font sizes or disable the feature altogether. 

Otherwise, it might complicate matters for those that want to add custom styling.

joepavitt:
You're right to a certain extent, but the sizings are driven by `css` classes:

```
/* Size Overrides */

/* xs */
.nrdb-ui-gauge-size-xs .nrdb-ui-gauge-value span {
    font-size: 1.5rem;
    line-height: 1.75rem;
}

/* sm */
.nrdb-ui-gauge-size-sm .nrdb-ui-gauge-value span {
    font-size: 2rem;
    line-height: 2.25rem;
}

/* md */
.nrdb-ui-gauge-size-md .nrdb-ui-gauge-value span {
    font-size: 2.5rem;
    line-height: 3rem;
}
.nrdb-ui-gauge-size-md .nrdb-ui-gauge-value label {
    font-size: 1rem;
    line-height: 1.25rem;
}

/* lg */
.nrdb-ui-gauge-size-lg .nrdb-ui-gauge-value span {
    font-size: 3rem;
    line-height: 3.25rem;
}
```

So can be overriden within a user's own CSS too

cstns:
Yes, that should be mentioned in the docs somewhere so users won't have to dig through the code base to find that out.

Other than that I don't any see technical improvement that could be made. I found a scaling solution which uses the d3 lib to place a watcher on the window and updates graph .tick (labels) font-size on window resize:
```js
d3.select(window).on("resize", function() {
    const newWidth = d3.select("svg").style("width");
    const newFontSize = 10 * (600 / parseInt(newWidth));
    d3.selectAll(".tick").select("text")
        .style("font-size", newFontSize)
});
```
I don't think it brings much to the table except for not using breakpoints to scale font-size and has the drawback of not being overwritten.

I feel that the approach you took is better for this use case

joepavitt:
Thanks Serban - would appreciate a review along with the comments :)

joepavitt:
Docs for class overrides added @cstns

cstns (approved):
Looks good! Thank you!

PR #920, opened 2024-05-30
Ensure service worker is auto-updated on refresh

joepavitt:
## Description

When developing locally, the new service worker introduced kept prompting a "reload" button every time I did a new build, and the UI would not correctly update until this was acknowledged.

This skips the confirmation requirement, and replaces the old service worker automatically, speeding up local development times.


cstns (approved):

joepavitt:
Just flagging this to @cgjgh in case we've missed something obvious that this would be a requirement for? Possibly the mobile auth blocks, where the "reload" fires a fresh service worker, but my understanding is that this would fire anyway, with this skip in place?

PR #921, opened 2024-05-30
GH Action - Version ump actions/upload-page-artifact

joepavitt:
## Description

With the recent dependabot updates, we can no longer deploy our documentation. The [failing build](https://github.com/FlowFuse/node-red-dashboard/actions/runs/9304191790/job/25608198285) complains:

```
Ensure artifacts are uploaded with actions/upload-artifact@v4 or later.
```

We aren't using that action, but are using https://github.com/actions/upload-pages-artifact which was still on `v1`, whereas `v3` is available, hoping this update resolves the problem.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -27,7 +27,7 @@ jobs:
         env:
             VITE_POSTHOG_APIKEY: ${{ vars.POSTHOG_API_KEY }}
       - uses: actions/configure-pages@v5
-      - uses: actions/upload-pages-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           path: docs/.vitepress/dist
       - name: Deploy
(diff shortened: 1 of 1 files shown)


cstns:
Don't think the version bump will help, there's a deprecation notice for `actions/upload-pages-artifact@v1`, `@v2` & `@v3`
image


The solution would be to use `actions/upload-artifact@v4` instead

joepavitt:
Ah, missed that `v3` in that list - thanks - I'll update the action

cstns:
Is it possible to do a dry-run with the new action?

joepavitt:
> Is it possible to do a dry-run with the new action?

Nope, welcome to GH Actions 😁

joepavitt:
@cstns updated :)

cstns (approved):
:crossed_fingers:

Issue #928, opened 2024-05-31
Notification node does not send output messages anymore


bartbutenaers:
### Current Behavior

Hi guys,

I wanted to create a new pull request for the Notification node, but that node does not send output messages anymore (both for `widget-action` and `widget-change`) when the button is clicked or at timeouts.   Some time ago it worked well, and the [ui_notification.js](https://github.com/FlowFuse/node-red-dashboard/blob/main/nodes/widgets/ui_notification.js) code hasn't been changed meanwhile.

When I click the button, the msg is being swallowed [here](https://github.com/FlowFuse/node-red-dashboard/blob/main/nodes/config/ui_base.js#L516):

![image](https://github.com/FlowFuse/node-red-dashboard/assets/14224149/290bcea8-8d61-4330-8059-6af0bf76355a)

So I assume I can fix it by applying an onAction handler in the ui-notification.js file?
But I am a bit confused why it worked before, because it looks to me that the above check is already quite some time available in the ui_base.js file...

Does anybody has an idea what could have caused this?

Thanks!!
Bart

### Expected Behavior

The Notification node sends an output message when I click the button.

### Steps To Reproduce

Click the button on the Notification node.

### Environment

- Dashboard version: 1.10.0
- Node-RED version: v3.1.3
- Node.js version: v18.19.0
- npm version: 10.2.3
- Platform/OS: Windows
- Browser: Chrome

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

colinl:
With 1.11.1 I am finding that it does send a message using this flow, but it sends it immediately rather then when it times out.
```
[...]

Issue #733, opened 2024-04-03
FR Notification: ability to close a notification with an input message


SynoUser-NL:
### Description

We are using the Notification node to display a Please wait.. message for actions that take a little bit of time. It would be nice to be able to have the flow close the Notification once the action is completed.
Currently, the only possibility is to have the Notification time-out or have the user close the Notification.

### Have you provided an initial effort estimate for this issue?

I am no FlowFuse team member

SynoUser-NL:
Thanks @joelvandal ! 
Tried your solution and this works exactly how I think the Notification node should behave. So great work!

That being said, I do think that this functionality should be part of the DB2 Notification node. The bit you're doing in the <scripts> section and even the <v-card-actions> section isn't really low-code anymore, so adjusting this for more buttons would not be easy for most low-code users. Also, when trying with the Cancel button enabled, clicking that button does not generate a payload message to take action on (although I'm sure that could be programmed into the behaviour of the modal).

For us, should we want to make NR instances available for more people to be used as a low-code solution, we'd like something like your solution to just be possible with the standard Notification node that is also maintained as part of the Dashboard package.
Hoping that you and @joepavitt  agree..

SynoUser-NL:
Looks great!
So this will be an additional Dashboard 2 node, to be installed with the Palette Manager?

Steve-Mcl:
This is more possible via #932

Issue #606, opened 2024-02-23
UI Notification: Confirmation Buttons


joepavitt:
### Description

![image](https://github.com/FlowFuse/node-red-dashboard/assets/99246719/df608cfd-fb19-4873-88d7-f50e54a6c03a)

In Dashboard 1.0, it's possible to define the "Layout" as a confirmation dialog instead. Currently, we only support a timed notification.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

colinl:
Is this the same issue as https://github.com/FlowFuse/node-red-dashboard/issues/193 "Captures the "Dialog" functionality from ui-toast in Dashboard 1.0, that has not been ported to the ui-notification widget in D2.0"

joepavitt:
Ah, better project management than I can handle there @colinl - thanks for spotting! Will close this is a duplicate

Issue #193, opened 2023-09-12
UI Dialog


joepavitt:
### Description

Captures the "Dialog" functionality from `ui-toast` in Dashboard 1.0, that has not been ported to the `ui-notification` widget in D2.0

This should utilise 

### Properties

- Labelling:
    - Cancel
    - OK/Confirm
- Properties from Vuetify we could implement:
    - Full Screen (https://vuetifyjs.com/en/components/dialogs/#fullscreen)
    - Transitions (https://vuetifyjs.com/en/components/dialogs/#transitions)

### Events

- **on-action**: On submitting/confirming the dialog, a `msg` should be sent on with relevant content.

### Existing Examples

https://vuetifyjs.com/en/components/dialogs/

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

colinl:
When this is implemented I think it is important that the message sent on clicking a button includes all the extraneous properties of the incoming message.  That allows the nodes after the dialog to know what it is that is being OKed or Cancelled.

bartbutenaers:
@joepavitt,
I think this is the last feature I need to be able to migrate my dashboard.
Bit confused now: is this a new feature in the exististing ui-notification widget, or a new widget?
I can have a look at this one if you like.  But without the fancy CSS stuff...

joepavitt:
> Bit confused now: is this a new feature in the exististing ui-notification widget, or a new widget?

Hadn't been settled upon, so open to discussion.

I see merit on both sides. Functionality wise, I think it's sensible on its own, but at the same time, I don't like overwhelming the palette with excessive nodes.

bartbutenaers:
Yes indeed, you are right.  
I don't think it can harm if we allow it (optionally) in the ui-notification widget.
Afterwards some kind of full blown ui-dialog node perhaps can see the daylight one day...
I will try to have a look at it when I have time.

bartbutenaers:
@joepavitt,
I want to add an (optional) confirmation button to the notification ui node, but would like to have your opinion.  Currently that node works like this:

1. When you click the 'Close' button, a payload `clicked` is being send:
   ```
   <v-btn variant="text" @click="close('clicked')"
   ```
2. After a specified timeout, a payload `timeout` is being send:
   ```
   this.timeouts.close = setTimeout(() => {
      this.close('timeout')
   }, time)
   ```

[...]

PR #932, opened 2024-05-31
notification ui node enhancements

bartbutenaers:
## Description

This PR adds some enhancements to the Notification ui node.  Some of these enhancements were a bit too much related, so I decided to combine them in a single PR.

## 1. Help panel
There was no help panel yet for this node, so I added one:

image

## 2. Output msg fix
As described in issue [928](https://github.com/FlowFuse/node-red-dashboard/issues/928) the output messages were not being send anymore.  Don't know since which version, or which change caused it.  Just fixed it...

## 3. Close via msg
As described in issue [733](https://github.com/FlowFuse/node-red-dashboard/issues/733) it should be able to close the notification dialog using an input message.  

I used some kind of toggle mechanism: when you inject a msg and the dialog is closed, then it will be opened.  And the other way around:

toggle_notification

So I don't look at the content of the payload.  Don't know if that is ok?  I have **_not_** mentioned it yet in the documentation for this reason.

## 4. Confirmation button
Beside the already existing dismiss button, there is now also a confirmation button.  As described in issues [606](https://github.com/FlowFuse/node-red-dashboard/issues/606) and (partially) [193](https://github.com/FlowFuse/node-red-dashboard/issues/193).

+ By default the new confirmation button is inactive, to behave like older instances of this node:

   image

    In that case only the existing dismiss button is being displayed, similar to in the previous version:

   image

   When clicked, a message is send:

   image

   As discussed this is a ***breaking change*** because previously a payload *"clicked"* was being used, which would not be very useful in a context of two buttons...

+ When activated, it becomes possible to specify the label of the second button:

   image

   In this case the two buttons will be displayed:

   image

   When the new confirmation button is clicked, a message is send:

   image

+ When only the new confirmation button is acivated:

   image

   Then only the new button is being displayed:

   image

+ And when custom labels are specified for both buttons:

   image

   Then both labels will be displayed:

   image

## Related Issue(s)

Issue [928](https://github.com/FlowFuse/node-red-dashboard/issues/928)
Issue [733](https://github.com/FlowFuse/node-red-dashboard/issues/733)
Issue [606](https://github.com/FlowFuse/node-red-dashboard/issues/606) 
Issue [193](https://github.com/FlowFuse/node-red-dashboard/issues/193) (partially)

## Checklist

 - [ X ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


colinl:
I thought that the OK/Cancel feature was going to be handled by a new ui-dialog rather than being added to the notification node.
See #193.

colinl:
OK, I have just seen the latest posts on #193 where it was decided to do it the way D1 did it.

ui/src/widgets/ui-notification/UINotification.vue
```diff
@@ -98,7 +111,10 @@ export default {
         },
         close (payload) {
             this.show = false
-            this.$socket.emit('widget-action', this.id, payload)
+
+            const msg = this.messages[this.id] || {}
+            msg.payload = payload
```
joepavitt, line 116: This is causing a very brief flash changed of the notification as it closes. We'd be better off here making a new `msg` object to emit, rather than overriding the local scope.

ui/src/widgets/ui-notification/UINotification.vue
```diff
@@ -98,7 +111,10 @@ export default {
...
+            this.$socket.emit('widget-action', this.id, msg)
```
joepavitt, line 117:
```suggestion
            this.$socket.emit('widget-action', this.id, {
                ...msg,
                payload
            })
```

ui/src/widgets/ui-notification/UINotification.vue
joepavitt, line 116:
```suggestion
```

joepavitt (changes requested):

ui/src/widgets/ui-notification/UINotification.vue
bartbutenaers: Hey @joepavitt 
Thanks for reviewing this in depth!!
I hadn't noticed the brief flashing.  Will commit your changes...

joepavitt:
For the close via `msg` - currently, that would close for _any_ payload received? I think we need to be a little more strict in it's requirements. To be consistent with other approaches, we could use a `msg.ui_update` option, but that's generally reserved for dynamic/persistent changes. Here I think a `msg.show = false` would be appropriate?

bartbutenaers:
@joepavitt,
You can implement this in a million different ways.  For me `msg.show = false` is fine.  I don't really have an opinion about this one.  Just let me know.

joepavitt:
Actually, let's go `msg.clear_notification = true` - feels more instructive

colinl:
Should `msg.reset` be a candidate, as that is used elsewhere for clearing things?

joepavitt:
> as that is used elsewhere for clearing things?

is it?

colinl:
Delay and Trigger nodes for example.

bartbutenaers:
Yes but inside the dashboard it is more common usage to clear stuff e.g. using an empty string in the payload (see dropdown, radiobuttons, ...)

colinl:
True.

bartbutenaers:
@joepavitt,
I have updated the code to close the window if `msg.clear_notification = true` is injected.

However when I inject, `msg.clear_notification = false` I expect nothing to happen.  Instead the title of the notification disappears:

notification_header_disappears

Below the root cause is explained, but I don't know how to solve it.  A bit too much Vue specific for me...

image

1. When the message arrives in the frontend, it is stored
2. Since the `msg.clear_notification = false` I ignore the message.
3. The value changes because of the stored message.  But since my msg has no payload, the value will be undefined.
4. The changed value will let the title disappear.

Perhaps I should only store the message in the else section, i.e. only store the message if it is going to be used to show a new dialog?

bartbutenaers:
@colinl,
You are right.  Will change it tonight.
After a long heavy day at work, the dashboard code isn't always clear to me.
As I said before, storing (and more worse replaying) of messages has been a source of troubles in the ui nodes over many years.  
It makes no sense at all to do it that way, and it continues to trigger unexpected weird side behaviour...

joepavitt:
I'm happy to cover this @bartbutenaers 

Colin is right, we just want a conditional to check a `payload` is present, before storing the data and overriding `msg` in the client

bartbutenaers:
Thanks @joepavitt !!!

@colinl 
To explain my last feedback above a bit more.

My issue from last night is another nice example of [this](https://github.com/FlowFuse/node-red-dashboard/issues/61#issuecomment-1684530029) discussion.  I inject a message which does _**not**_ contain any information about the notification title, and hence the title is gone.  Because my message becomes the last stored message, and the widget expects to find ALL his information in that last message.   But the last message only contains only a little part of all the required state information of course.  And now you as a node developer need to start thing about which message should be stored and which one not.  Only to make sure that the required information is available in the last stored message.  Which imho is not a pretty solution anymore...

I would expect that:
1. The first message would update all kind of properties in the server side state (e.g. `content`, `title`, ...).
2. That state delta is forwarded to the required clients, which update their Vue template and show the notification.
3. My last message only updates the `show` property to `false` in the server side state.
4. That state delta is forwarded to the required clients, which update their Vue template and hide their notification.

Sorry guys to repeat this theory over and over again.  But I still wake up screaming in the middle of the night, being haunted by hundreds of replayed messages with faulty content ;-)

joepavitt:
As an example if you've defined `colour: blue` on the node's config.

You send a message which defined `msg.color = red` with a `msg.payload`, and then after that, sent just a msg.payload, what colour should it be?

I'd say blue?

In the pattern you've proposed though, I understand you'd ask for `red` to have been stored if we merge `msg` objects?

bartbutenaers:
> In the pattern you've proposed though, I understand you'd ask

@joepavitt,
I have added my feedback to my original discussion about server side state: see [here](https://github.com/FlowFuse/node-red-dashboard/issues/61#issuecomment-2166717542), to keep everything nicely together.  I have added an example that hopefully makes it a bit more clear.

Steve-Mcl:
should the command not be packed into `ui_control` as per other nodes control methods for symetry?  

Additionally, shouldnt we emulate the options for both close and confirm e.g. `msg.ui_control.close_notification` AND/OR `msg.ui_control.confirm_notification`  ?

joepavitt:
@Steve-Mcl I had initially [ruled that out](https://github.com/FlowFuse/node-red-dashboard/pull/932#issuecomment-2162387772) because it wasn't persisting any changes, and note we changed it to `msg.ui_update`. That terminology change is important here.

`ui_control` reads better in this case, and we could introduce that _alongside_ `ui_update`, to differentiate between one-off calls/commands vs. overriding of a property?

joepavitt:
> ui_control reads better in this case, and we could introduce that alongside ui_update, to differentiate between one-off calls/commands vs. overriding of a property?

What are your thoughts @bartbutenaers @colinl ?

colinl:
`ui_control` sounds good to me where it is an action to do something, and `ui_update` for where it is a state change.  When closing the notification I guess it is actually a state change of whether it is visible or not, but from the point of the user it is an action to Close it.

bartbutenaers:
@joepavitt,
Sorry but have never used ui_control, so unfortunately I can't do any suggestions about that.

joepavitt:
> Sorry but have never used ui_control, so unfortunately I can't do any suggestions about that.

It's a new proposal for Dashboard 2.0 - so just general thoughts around having the two `ui_update` and `ui_control` where appropriate, or do we just stick with the single `ui_update` approach?

bartbutenaers:
Hi @joepavitt,
Imho @colinl has summarized it 100% correctly.  
From my developer point of view I would call it a state change of property `visible`.  However our target audience in a low code environment are not developers, so I would go for ui_control.
Glad that @Steve-Mcl mentioned it.

joepavitt:
Thought about this more, I think we stick with `ui_update`, no need to ahve many options here and confuse users - _anything_ "dynamic" falls under `ui_update` - easier that way

bartbutenaers:
Hi @joepavitt,
yes indeed I agree with you.
After all reading I did last week, I also think ui_update makes the most sense.

I wanted to implement this quickly tonight, but then I stumbled upon the todo described [above](https://github.com/FlowFuse/node-red-dashboard/pull/932#issuecomment-2163693608).
Which means I have to avoid storing the message in case of a "hide" command, because otherwise the title is lost.
But as soon as you would start injecting separate messages for title, color, ... in the future, then this wouldn't work anymore.

To be honest, for me all of this feels like creating workarounds for the limitations caused by storing and replaying messages.  

I am not sure if you already had a look at the [discussion](https://discourse.nodered.org/t/how-to-sync-the-full-config-between-server-and-client-side-state/89252/29?u=bartbutenaers) I had with @colinl...
Imho we only need something like this e.g. for the Notification node:

image

Summarized: updating the state store via input messages, and automatically syncing the state store with the client(s).
You could even override some general property values (which apply to all clients) for a single client:

image

All clients would get color `b`, except client yyy would get color `d`.
As a result no messages to be stored or replayed...

If I overlooked something, don't hesitate to let me know why this simple approach would not work!

joepavitt:
> I am not sure if you already had a look at the [discussion](https://discourse.nodered.org/t/how-to-sync-the-full-config-between-server-and-client-side-state/89252/29?u=bartbutenaers) I had with @colinl...

I have replied to that :)

bartbutenaers:
@joepavitt,
I will continue this pull request after you have published the [new helper functions](https://github.com/FlowFuse/node-red-dashboard/pull/1123).  Otherwise I need to refactor this one again afterwards

joepavitt:
FYI @bartbutenaers https://github.com/FlowFuse/node-red-dashboard/pull/1123 is now merged

bartbutenaers:
@joepavitt
I made some progress for the dynamic properties:

notification_dynamic_properties
```
[{"id":"beaecd4993da0288","type":"ui-switch","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"","label":"switch","group":"182c087288129e9e","order":1,"width":0,"height":0,"passthru":false,"decouple":false,"topic":"topic","topicType":"msg","style":"","className":"","onvalue":"Machine Started","onvalueType":"str","onicon":"","oncolor":"","offvalue":"Machine Stopped","offvalueType":"str","officon":"","offcolor":"","x":130,"y":160,"wires":[["584b6aaa89870656"]]},{"id":"b669d2649ba2f496","type":"debug","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Notification output","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","statusVal":"","statusType":"auto","x":690,"y":160,"wires":[]},{"id":"584b6aaa89870656","type":"ui-notification","z":"4aad778b57d4f47b","g":"6394e962049dadeb","ui":"be29745a6e568f30","position":"center center","colorDefault":false,"color":"#f41515","displayTime":"0","showCountdown":true,"outputs":1,"allowDismiss":true,"dismissText":"Skip","allowConfirm":false,"confirmText":"","raw":false,"className":"","name":"","x":470,"y":160,"wires":[["b669d2649ba2f496"]]},{"id":"6c971642351df5e3","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Close notification","props":[{"p":"ui_update.close","v":"true","vt":"bool"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":160,"y":100,"wires":[["584b6aaa89870656"]]},{"id":"deb020c3693a2513","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Position top center","props":[{"p":"ui_update","v":"{\"position\":\"top center\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":210,"y":220,"wires":[["584b6aaa89870656"]]},{"id":"12dacd88f6be4e7d","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Color orange","props":[{"p":"ui_update","v":"{\"color\":\"orange\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":230,"y":320,"wires":[["584b6aaa89870656"]]},{"id":"5be412e2e8fc9d34","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Show confirm button","props":[{"p":"ui_update","v":"{\"allowConfirm\":true,\"confirmText\":\"Let's go\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":210,"y":420,"wires":[["584b6aaa89870656"]]},{"id":"cdfb2e3bd87b1815","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Hide dismiss button","props":[{"p":"ui_update","v":"{\"allowDismiss\":false}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":210,"y":560,"wires":[["584b6aaa89870656"]]},{"id":"1af3465cd2962de9","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Dynamic title 1","props":[{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"Dynamic title 1","payloadType":"str","x":220,"y":620,"wires":[["584b6aaa89870656"]]},{"id":"8f99e507001af622","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Hide countdown bar","props":[{"p":"ui_update","v":"{\"showCountdown\": false}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":210,"y":760,"wires":[["584b6aaa89870656"]]},{"id":"1f42a55f41791c8a","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Progress color light green","props":[{"p":"ui_update","v":"{\"progressColor\": \"lightGreen\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":550,"y":820,"wires":[["584b6aaa89870656"]]},{"id":"51f13a9f6ba4549d","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Accept raw html","props":[{"p":"ui_update","v":"{\"raw\": true}","vt":"json"},{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"<b style=\"color:red;\">Red text</b>","payloadType":"str","x":220,"y":820,"wires":[["584b6aaa89870656"]]},{"id":"85616f21ca71b9fa","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Show dismiss button","props":[{"p":"ui_update","v":"{\"allowDismiss\":true,\"dismissText\":\"Cancel\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":210,"y":520,"wires":[["584b6aaa89870656"]]},{"id":"c30c6cbbb799895b","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Show countdown bar","props":[{"p":"ui_update","v":"{\"showCountdown\": true}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":200,"y":720,"wires":[["584b6aaa89870656"]]},{"id":"593ad4d37c38b413","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Position center center","props":[{"p":"ui_update","v":"{\"position\":\"center center\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":200,"y":260,"wires":[["584b6aaa89870656"]]},{"id":"81502b8c45d8df7a","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Color light green","props":[{"p":"ui_update","v":"{\"color\":\"lightgreen\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":220,"y":360,"wires":[["584b6aaa89870656"]]},{"id":"3ac83bac84929c3a","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Hide confirm button","props":[{"p":"ui_update","v":"{\"allowConfirm\":false}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":210,"y":460,"wires":[["584b6aaa89870656"]]},{"id":"93ed9dedd53112e7","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Dynamic title 2","props":[{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"Dynamic title 2","payloadType":"str","x":220,"y":660,"wires":[["584b6aaa89870656"]]},{"id":"2e30133557f53167","type":"inject","z":"4aad778b57d4f47b","g":"6394e962049dadeb","name":"Don't accept raw html","props":[{"p":"ui_update","v":"{\"raw\": false}","vt":"json"},{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"<b style=\"color:red;\">Red text</b>","payloadType":"str","x":200,"y":860,"wires":[["584b6aaa89870656"]]},{"id":"182c087288129e9e","type":"ui-group","name":"Historiek","page":"b262ac19caee413d","width":"11","height":"1","order":1,"showTitle":true,"className":"","visible":"true","disabled":"false"},{"id":"be29745a6e568f30","type":"ui-base","name":"UI Name","path":"/dashboard","includeClientData":true,"acceptsClientConfig":["ui-notification","ui-control","ui-chart","ui-text-input","ui-dropdown"],"showPathInSidebar":false,"titleBarStyle":"default"},{"id":"b262ac19caee413d","type":"ui-page","name":"Sensors historiek","ui":"be29745a6e568f30","path":"/sensors","icon":"thermometer","layout":"grid","theme":"a965ccfef139317a","order":10,"className":"","visible":"true","disabled":"false"},{"id":"a965ccfef139317a","type":"ui-theme","name":"Default","colors":{"surface":"#404040","primary":"#109fbc","bgPage":"#e8e8e8","groupBg":"#d6d6d6","groupOutline":"#6fbc10"},"sizes":{"pagePadding":"12px","groupGap":"12px","groupBorderRadius":"4px","widgetGap":"12px"}}]
```
The new helper functions really simplify developing things like this!! Great job!!

However the part that goes via the data store is still not clear to me.  I currently only store the input message in the data store, if it contains a payload, because the payload contains the text that needs to be displayed.   I have used the payload to send the text content, because I assume that that is the *'value'* for this widget?  Because that is the value that will be different between most notifications.   Not sure if that is correct?

Because I also need to show/hide the notification.  Should that be a boolean dynamic property `msg.ui_update.show`, because I cannot send a boolean in my payload (because I already use the payload for my notification text)?

Hope that my question is a bit clear?  I don't want the message to be displayed every time I inject a message to change some or another dynamic property...

joepavitt:
> I currently only store the input message in the data store, if it contains a payload, because the payload contains the text that needs to be displayed.

I think this is sensible given our current API

> Should that be a boolean dynamic property msg.ui_update.show, because I cannot send a boolean in my payload (because I already use the payload for my notification text)?

I'd be sending that via `msg.show`. For me, `msg.ui_update.<property>` should be used to override Node-RED configuration, and are values that would persist. In this case, `msg.show` is a "control" type event, not an override of a configuration/property.

bartbutenaers:
To explain my doubt a bit.  I had understood the new way of working like this: all dynamic properties are stored in the state store.  And the "***value***" of the widget is stored in the payload, so it will end up in the data store.  But what is considerd as 'the value' of e.g. a notification widget?  It could be the text that you display.  But it could also be the fact whether you show or hide the notification.  Would be nice to hear your thoughts about this...

bartbutenaers:
Hi guys,
I "think" everything is ok now.  It looks to me that the failed E2E tests are not related to this PR.

All the properties can now be overridden dynamically.
And the `msg.show` non-persistent field is being used to hide/show the notification dialog.

Here is my latest example flow:
```
[{"id":"beaecd4993da0288","type":"ui-switch","z":"4aad778b57d4f47b","name":"","label":"switch","group":"182c087288129e9e","order":1,"width":0,"height":0,"passthru":false,"decouple":false,"topic":"topic","topicType":"msg","style":"","className":"","onvalue":"Machine Started","onvalueType":"str","onicon":"","oncolor":"","offvalue":"Machine Stopped","offvalueType":"str","officon":"","offcolor":"","x":150,"y":100,"wires":[["59cd96e254b1e44d"]]},{"id":"b669d2649ba2f496","type":"debug","z":"4aad778b57d4f47b","name":"Notification output","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","statusVal":"","statusType":"auto","x":830,"y":100,"wires":[]},{"id":"584b6aaa89870656","type":"ui-notification","z":"4aad778b57d4f47b","ui":"be29745a6e568f30","position":"center center","colorDefault":false,"color":"#f41515","displayTime":"0","showCountdown":true,"outputs":1,"allowDismiss":true,"dismissText":"Skip","raw":false,"className":"","name":"","x":610,"y":100,"wires":[["b669d2649ba2f496"]]},{"id":"6c971642351df5e3","type":"inject","z":"4aad778b57d4f47b","name":"Show notification (all clients)","props":[{"p":"show","v":"true","vt":"bool"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":340,"y":160,"wires":[["584b6aaa89870656"]]},{"id":"deb020c3693a2513","type":"inject","z":"4aad778b57d4f47b","name":"Position top center","props":[{"p":"ui_update","v":"{\"position\":\"top center\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":370,"y":260,"wires":[["584b6aaa89870656"]]},{"id":"12dacd88f6be4e7d","type":"inject","z":"4aad778b57d4f47b","name":"Color orange","props":[{"p":"ui_update","v":"{\"color\":\"orange\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":390,"y":360,"wires":[["584b6aaa89870656"]]},{"id":"5be412e2e8fc9d34","type":"inject","z":"4aad778b57d4f47b","name":"Show confirm button","props":[{"p":"ui_update","v":"{\"allowConfirm\":true,\"confirmText\":\"Let's go\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":370,"y":460,"wires":[["584b6aaa89870656"]]},{"id":"cdfb2e3bd87b1815","type":"inject","z":"4aad778b57d4f47b","name":"Hide dismiss button","props":[{"p":"ui_update","v":"{\"allowDismiss\":false}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":370,"y":600,"wires":[["584b6aaa89870656"]]},{"id":"1af3465cd2962de9","type":"inject","z":"4aad778b57d4f47b","name":"Dynamic title 1","props":[{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"Dynamic title 1","payloadType":"str","x":380,"y":660,"wires":[["584b6aaa89870656"]]},{"id":"8f99e507001af622","type":"inject","z":"4aad778b57d4f47b","name":"Hide countdown bar","props":[{"p":"ui_update","v":"{\"showCountdown\": false}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":370,"y":800,"wires":[["584b6aaa89870656"]]},{"id":"51f13a9f6ba4549d","type":"inject","z":"4aad778b57d4f47b","name":"Accept raw html","props":[{"p":"ui_update","v":"{\"raw\": true}","vt":"json"},{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"<b style=\"color:red;\">Red text</b>","payloadType":"str","x":380,"y":860,"wires":[["584b6aaa89870656"]]},{"id":"85616f21ca71b9fa","type":"inject","z":"4aad778b57d4f47b","name":"Show dismiss button","props":[{"p":"ui_update","v":"{\"allowDismiss\":true,\"dismissText\":\"Cancel\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":370,"y":560,"wires":[["584b6aaa89870656"]]},{"id":"c30c6cbbb799895b","type":"inject","z":"4aad778b57d4f47b","name":"Show countdown bar","props":[{"p":"ui_update","v":"{\"showCountdown\": true}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":360,"y":760,"wires":[["584b6aaa89870656"]]},{"id":"593ad4d37c38b413","type":"inject","z":"4aad778b57d4f47b","name":"Position center center","props":[{"p":"ui_update","v":"{\"position\":\"center center\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":360,"y":300,"wires":[["584b6aaa89870656"]]},{"id":"81502b8c45d8df7a","type":"inject","z":"4aad778b57d4f47b","name":"Color light green","props":[{"p":"ui_update","v":"{\"color\":\"lightgreen\"}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":380,"y":400,"wires":[["584b6aaa89870656"]]},{"id":"3ac83bac84929c3a","type":"inject","z":"4aad778b57d4f47b","name":"Hide confirm button","props":[{"p":"ui_update","v":"{\"allowConfirm\":false}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":370,"y":500,"wires":[["584b6aaa89870656"]]},{"id":"93ed9dedd53112e7","type":"inject","z":"4aad778b57d4f47b","name":"Dynamic title 2","props":[{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"Dynamic title 2","payloadType":"str","x":380,"y":700,"wires":[["584b6aaa89870656"]]},{"id":"2e30133557f53167","type":"inject","z":"4aad778b57d4f47b","name":"Don't accept raw html","props":[{"p":"ui_update","v":"{\"raw\": false}","vt":"json"},{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"<b style=\"color:red;\">Red text</b>","payloadType":"str","x":360,"y":900,"wires":[["584b6aaa89870656"]]},{"id":"59cd96e254b1e44d","type":"change","z":"4aad778b57d4f47b","name":"show notification (1 client)","rules":[{"t":"set","p":"show","pt":"msg","to":"true","tot":"bool"}],"action":"","property":"","from":"","to":"","reg":false,"x":350,"y":100,"wires":[["584b6aaa89870656"]]},{"id":"f0fdaa8e1cd83af7","type":"inject","z":"4aad778b57d4f47b","name":"Close notification (all clients)","props":[{"p":"show","v":"false","vt":"bool"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":340,"y":200,"wires":[["584b6aaa89870656"]]},{"id":"6330f32fe804b0c6","type":"inject","z":"4aad778b57d4f47b","name":"Display time endless","props":[{"p":"ui_update","v":"{\"displayTime\": 0}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":370,"y":960,"wires":[["584b6aaa89870656"]]},{"id":"71a8a17786c3e85a","type":"inject","z":"4aad778b57d4f47b","name":"Display time 5 secs","props":[{"p":"ui_update","v":"{\"displayTime\": 5}","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":370,"y":1000,"wires":[["584b6aaa89870656"]]},{"id":"182c087288129e9e","type":"ui-group","name":"Historiek","page":"b262ac19caee413d","width":"11","height":"1","order":1,"showTitle":true,"className":"","visible":"true","disabled":"false"},{"id":"be29745a6e568f30","type":"ui-base","name":"UI Name","path":"/dashboard","includeClientData":true,"acceptsClientConfig":["ui-notification","ui-control","ui-chart","ui-text-input","ui-dropdown"],"showPathInSidebar":false,"titleBarStyle":"default"},{"id":"b262ac19caee413d","type":"ui-page","name":"Sensors historiek","ui":"be29745a6e568f30","path":"/sensors","icon":"thermometer","layout":"grid","theme":"a965ccfef139317a","order":10,"className":"","visible":"true","disabled":"false"},{"id":"a965ccfef139317a","type":"ui-theme","name":"Default","colors":{"surface":"#404040","primary":"#109fbc","bgPage":"#e8e8e8","groupBg":"#d6d6d6","groupOutline":"#6fbc10"},"sizes":{"pagePadding":"12px","groupGap":"12px","groupBorderRadius":"4px","widgetGap":"12px"}}]
```

Although the _**info panel**_ might need to be updated  Currently all the properties are described under ***Properties*** but also under ***Dynamic properties*** because all properties can be overridden.   Not sure if that is correct?  On the other hand the Properties section contain the descriptions like in the config screen, while the Dynamic properties section contains the msg property names.  So that is useful information, but every property description text is twice available in the info panel...

bartbutenaers:
BTW using the `msg.show` to show/hide the notification dialog is a _**breaking change**_.
Although I am wondering if this widget is currently being used a lot, since it contains quite a number of issues.

joepavitt:
> BTW using the msg.show to show/hide the notification dialog is a breaking change.

Why so? Are you no longer showing the notification on a new `msg.payload` anymore?

I'd interpreted the `msg.show` feature as an override to force open/close, even if a new `msg.payload` wasn't sent.

If a msg with a `msg.payload` is injected, then I would still expect it to show

bartbutenaers:
Ok then indeed no breaking change.  
Had interpreted you incorrectly, that the `msg.show` was the (only) way to go.
But makes sense indeed...  
Will update the code tonight. 
Enjoy your (small) break!!

bartbutenaers:
The logic has been adapted, to show the notification when there is an `msg.payload` or when `msg.show` is true.  When that is not the case, the notification will be hidden when `msg.show` is false.

bartbutenaers:
I had forgotton to add documentation for the `msg.show` field.  
Don't think there are "known" issues, so I consider this pull request ready for review...

Steve-Mcl:
> I had forgotton to add documentation for the `msg.show` field. Don't think there are "known" issues, so I consider this pull request ready for review...

If this is ready for final review Bart, can you refresh the review request please? 

image
_(hit the spinner button)_

ui/src/widgets/ui-notification/UINotification.vue
```diff
@@ -5,21 +5,29 @@
...
         multi-line
         :timeout="-1"
-        :location="props.position"
+        :location="position"
         :style="{'--nrdb-ui-notification-color': color}"
     >
-        <div v-if="props.showCountdown" class="nrdb-ui-notification-countdown">
-            <v-progress-linear v-model="countdown" :color="messages[id]?.color || (props.colorDefault ? 'primary' : color)" style="display: block; width: 100%" />
+        <div v-if="showCountdown" class="nrdb-ui-notification-countdown">
+            <v-progress-linear v-model="countdown" :color="color" style="display: block; width: 100%" />
```
gayanSandamal, line 12: @bartbutenaers I noticed that the progress bar of the notification is not picking the fill colour even the `Show timeout countdown bar on notification
 ` is checked. And when I investigate I found that this change `:color="color"` has caused it. 
When the default colour option is checked in the NR it picks the group colour `rgb(var(--v-theme-group-background))` which is handled in [this line](https://github.com/FlowFuse/node-red-dashboard/pull/932/files#diff-ba8467bf1f890995ffb760b3d600a311b3d3ef2d09afc82407ee769f7c8fb01aR72:~:text=.colorDefault)%20%7B-,return,-%27rgb(var)

Screenshot 2024-08-16 at 15 53 23

As in the existing implementation group colour applies to the background of the notification widget as well.

gayanSandamal (changes requested):
@bartbutenaers Thanks for improving the ui notification widget. Improvements are really helpful.

I found that there's a minor issue with the fill colour of the notification progress and added my comment

ui/src/widgets/ui-notification/UINotification.vue
bartbutenaers: @gayanSandamal Thanks a lot for your review!

I really think that I had updated that code, but will have done something wrong in one or another commit.    Appreciate that you have shared your solution!!  Will change it tonight.

> As in the existing implementation group colour applies to the background of the notification widget as well.

Just to make sure: is this also solved using your fix?

gayanSandamal: @bartbutenaers no worries at all with reviewing 🙌

It worked when I reverted and tried with the old code at line [12](https://github.com/FlowFuse/node-red-dashboard/pull/932/files/b0c10830ab60128686e41d02966aed7f4bfb13f5#diff-ba8467bf1f890995ffb760b3d600a311b3d3ef2d09afc82407ee769f7c8fb01a:~:text=%3Cv%2Dprogress%2Dlinear%20v%2Dmodel%3D%22countdown%22%20%3Acolor%3D%22messages%5Bid%5D%3F.color%20%7C%7C%20(props.colorDefault%20%3F%20%27primary%27%20%3A%20color)%22%20style%3D%22display%3A%20block%3B%20width%3A%20100%25%22%20/%3E)

bartbutenaers: @gayanSandamal,
I had indeed not paid attention enough to the progress bar color.
But not sure how to handle it, so some help would be welcome...

The original color for the progress bar worked like this:
```
"messages[id]?.color || (props.colorDefault ? 'primary' : color)
```

So I need to introduce a new computed field `progressColor` with the same logic as the original one:

image

However recently @joepavitt has introduced (kudos!) a new way of working for dynamic properties.  So the color is not available anymore in the messages, which means I cannot use that statement anymore.  Not sure if I should get the dynamic color now from somewhere else.  Because in the 'else' part I will get the static or dynamic color via getProperty (via `this.color`).

I am not sure which behaviour you guys prefer, but perhaps you have something simple in your head?

Thanks!!!!

gayanSandamal: You can achieve this by using the following code. I've moved the same existing logic to the `progressColor` computed property:

```
progressColor () {
    return this.messages[this.id]?.color || (this.props.colorDefault ? 'primary' : this.color)
},
```

Screenshot 2024-08-18 at 20 37 51

@joepavitt What do you think regarding this?

bartbutenaers: @gayanSandamal,
Thanks for thinking out loud!!  

BTW I sometimes ask questions during the weekend (because I have a bit more free time then), but that doesn't mean I expect you guys to respond during the weekends ;-)

> I've moved the same existing logic to the progressColor computed property

That was indeed my first thought, but I didn't even tried it because the theory in my brain tells me it wouldn't work.  But perhaps my theoretical knowledge is incorrect...

+ Recently Joe had redesigned the dynamic property mechanism.  Since then I expect the color to be provided via `msg.ui_update.color` (instead of `msg.color`), which will be stored in the *state store*.  
+ The messages are - from this PR on - only stored in the *data store* when a payload is provided.  Because the text (to be displayed) does not to be persisted in the state store, so it is stored (as `msg.payload`) in the data store.  Because if you would inject a message without payload (e.g. to set only the color), then the message would disappear.
+ So I don't expect that the color can be found anymore in the stored messages (in the data store).  Which means I don't expect the `this.messages[this.id]?.color` to be available anymore.
+ And in the new dynamic property mechanism, the `this.color` will - via `this.getProperty('color')` - will return the dynamic color value or the static property value.  So in fact the old `this.messages[this.id]?.color` is now returned by `this.color` (only if a dynamic property value has been injected).

This is the reason why I am doubting how the statement should be migrated...

gayanSandamal: @bartbutenaers thanks for the clarification.

Regarding "this.messages[this.id]?.color" I will check this with @joepavitt offline and provide my feedback

joepavitt: We are in that awakwars position where we haven't yet migrated over `ui-notification` to the new API - so as Bart has put it is fine. Don't want to overload this PR with the migration to the new API _as well_

gayanSandamal: @joepavitt Thanks for the clarification.

@bartbutenaers Based on Joe's feedback, it makes sense to proceed with the current approach, sticking with msg.color for now.

gayanSandamal (approved):
This looks good to me

ui/src/widgets/ui-notification/UINotification.vue
bartbutenaers: > Don't want to overload this PR with the migration to the new API as well

@joepavitt,
I had last week also updated this node (in this PR) to migrate it to the new API.  Thought that would be helpful for you guys, to get a bit of pressure from your shoulders.  Hopefully that is not a problem, because I don't really have much time to refactor it again to use the old API...

But that means that `msg.color` is not used anymore for this node, which is why I don't know how to migrate `this.messages[this.id]?.color`

joepavitt: > this.messages[this.id]?.color

Would instead then just be covered by the new API's `getProperty`

Issue #810, opened 2024-04-27
Dashboard does not reconnect if network is down for a significant time.


colinl:
### Current Behavior

I find that if, on a PC, I open the dashboard and disconnect the network and leave it for 30 mins (not sure what the min time to show the problem is) then it never reconnects when I connect the network again, it just sits there with Connection Lost popup. I see the same on Android sometimes when I re-open the dashboard window after it has been in the background for some time. I suspect it is the same issue.

### Expected Behavior

Not sure.  In the same circumstance the editor shows a Reconnecting in nn seconds and a Try Now link. I think it tries every 30 secs or maybe 60.

I think that in the same circumstance, D1 reconnects automatically, and quickly, as I don't see an issue there in the same circumstances (on Android at least).  I am going to try some tests on the PC with D1 to check how it performs.

In the same circumstance the editor shows a 'Reconnecting in nn seconds' popup with a Try Now link. I think it tries every 30 secs or maybe 60, but it would be a pain if every time I open the dashboard on the phone I have to click something to get it to reconnect.

### Steps To Reproduce

Open the dashboard and disconnect the network. Leave for 30 mins and then reconnect.

### Environment

- Dashboard version: 1.8.1
- Node-RED version: 4.0.0 Beta 2
- Node.js version:
- npm version:
- Platform/OS: Browser on Ubuntu laptop, server on Pi
- Browser: Edge

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

colinl:
In the case where I suspend the laptop and then start it again some time later then again I get the connection lost message and it does not reconnect.  Perhaps what is happening here is that node red gets going again before the network is up so starts the reconnection sequence, but since it is a long time since the last good connection it immediately hits the timeout and so does not try again.  If that fits the way the code works then perhaps the algorithm should work based on the number of retry attempts rather than a timeout, so in the case of suspend (or Android pausing it when it is running in the background) it would keep trying to connect for some time after the device is woken up.

colinl:
Fixed by PR #950

Issue #946, opened 2024-06-06
Improvements needed to handling of socket connection lost


colinl:
### Description

Currently, in the client, if the socket disconnects then it pops up Connection Lost alert and enters a sequence of repeated attempts to reconnect.  Initially it tries every 5 seconds, for a period of 60 seconds, then it drops back to trying every 30 seconds, After a total of 5 minutes it gives up entirely, leaving the Connection Lost, Trying to reconnect message on screen, which is not correct as it is no longer retrying.  Some other action needs to be taken at this time.  In addition there are some reasons for connection lost that will not recover (such as an authorisation token expiring), the retry sequence is not appropriate in these cases.

I am currently attempting to identify possible causes of disconnection and how to identify them. I will update this as I build the list, and will propose appropriate recovery action.

The first class of failure is for situations where attempting to reconnect for  period is entirely appropriate.  This includes network failures that prevent access to the server, and problems with node red itself that prevent the connection from succeeding, for example the server is stopped or is in the process of being restarted.
The retry algorithm described above has two problems:
1. If a laptop is suspended then the network may be shut down and the connection may fail before node red is suspended.  If it is left suspended for more than 5 minutes then the retry algorithm notices that more than 5 minutes has elapsed since the connection failed, so does not try again.  I believe a similar thing may happen on Android if the the OS decides to put the app to sleep when the screen is off or the app is put in the background.  To avoid this I propose to change the retry algorithm from time based to retry count based, so it will try 12 times at 5 second intervals, then another 8 times at 30 second intervals.
2. If the connection is not restored within the 5 minutes then the user is left with a notification saying the app is trying to reconnect, but in fact it is not trying.  I propose that after the 5 minutes, rather than just hanging, then we should force a page reload.  I will do some tests to see what the effect of this is when the connection still cannot be restored.

The second class of failures is for situations where it is permanently not possible to reconnect.  For these there is no point retrying, the solution is to immediately force a page reload. An examples of this is where an authorisation token has expired and it is necessary to reload in order to load the logon page. The only situation I have been able to test for this is a system that uses Cloudflare Zero Trust to connect to the server, and in that case, when the token expires, after a little while a `socket.on('connect_error', (err) => {` is triggered, where the error contains `{"code": "parser error"}`.  I propose to trap that, and immediately force a page reload for that situation.  There may well be other such permanent failure modes, but unless anyone can identify them now I don't see what we can do other than wait for them to be identified.

[...]

PR #950, opened 2024-06-08
Use retry count based algorithm for socket reconnections

colinl:
## Description

Rather than using a time based algorithm for determining how often to retry a socket connection failure, use a count of how many retries have been performed.  This overcomes the fact that during laptop suspend, browser page in the background, Android putting a background to sleep, and other similar situations, the connection may fail as the page is suspended, and will not start retrying until the page is re-activated.  The result is that the existing code thinks that a long time has elapsed since the connection fails, and may just give up immediately.

## Related Issue(s)

Fixes issue #810, addresses the first class of failures in #946. Also see #784.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/main.mjs
+++ b/ui/src/main.mjs
@@ -108,10 +108,10 @@ fetch('_setup')
         store.commit('setup/set', setup)
 
         let disconnected = false
-        let retryCount = 0      // number of reconnection attempts made
+        let retryCount = 0 // number of reconnection attempts made
 
         let reconnectTO = null
-        const MAX_RETRIES = 22      // 4 at 2.5 seconds, 10 at 5 secs then 8 at 30 seconds
+        const MAX_RETRIES = 22 // 4 at 2.5 seconds, 10 at 5 secs then 8 at 30 seconds
 
         const socket = io({
             ...setup.socketio,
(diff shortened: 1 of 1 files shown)


colinl:
Having used this for a couple of days I have adjusted the sequence slightly. When using the dashboard with Android, if the app is left in the background for a while then when it is reopened it usually has to reconnect, and 5 seconds is quite a long time to wait, so I have adjusted it to have four goes at 2.5 sec interval, then 5 secs for the rest of the first minute, then 30 seconds up to 5 mins as before.

joepavitt (approved):
Great contribution @colinl - really appreciate it

joepavitt:
@colinl looks like you have some linting errors to resolve before we can merge: https://github.com/FlowFuse/node-red-dashboard/actions/runs/9443384413/job/26074664351?pr=950

colinl:
Thanks Joe, lint errors fixed. Only one space allowed before \\ seems a bit odd to me, but if that is the rule then so be it.

colinl:
Is the E2E test failure something I need to do anything about?

joepavitt:
Nope, that's not on you! I'll merge tomorrow

colinl:
OK, thanks.

Issue #912, opened 2024-05-29
Format of timescale axis in chart node


bartbutenaers:
### Description

Hi guys,

I added a chart to my dashboard to show some timeseries data, so I selected the timescale x-axis type:

![image](https://github.com/FlowFuse/node-red-dashboard/assets/14224149/f4c52488-cb5b-4545-bd7a-bbec48dd7f11)

Which looks like this in the dashboard:

![image](https://github.com/FlowFuse/node-red-dashboard/assets/14224149/b02f59ba-b73c-49e8-bc3a-ddfedcabf2d9)

Me and my familly are Dutch speaking, so the PM and AM is quite confusing for us.  
We are used that times are represented in a 24 hour format, for example `07:24` or `23:59`.  

Not sure if there is already a possibility to configure it somehow like that?  I assume not...

Thanks!
Bart

### Have you provided an initial effort estimate for this issue?

I am no FlowFuse team member

joepavitt:
@bartbutenaers what's the equivalent in Dashboard 1.0 for this? Or is it 24 hr clock by default?

bartbutenaers:
@joepavitt
To be honest I have never made a decent dashboard before...

But by looking at the config screen of the Chart ui node in dashboard D1, it looks like there were already some formatting options available:

![image](https://github.com/FlowFuse/node-red-dashboard/assets/14224149/fe7e19e5-a01a-4e82-903a-ffb09691ada8)

When selecting "custom" it was also possible to enter any kind of format:

![image](https://github.com/FlowFuse/node-red-dashboard/assets/14224149/67f6b006-4adf-4d84-b008-66408f9e6b31)

bartbutenaers:
Joe, if I you want I can try to create a PR for this one...

joepavitt:
PRs always welcome 😊

PR #951, opened 2024-06-09
Chart x axis format

bartbutenaers:
## Description

This PR adds a setting to the Chart ui node, to change the format of the displayed time values on the X axis.  This setting becomes visible when the X-axis type is `timescale`:

date_format

I added the same time formats in the dropdown as in the original Chart node in the old dashboard.  Although I had to change the tokens, because the old dashboard uses the [moment](https://momentjs.com/docs/#/displaying/format/) tokens while the new dashboard uses the [luxon](https://moment.github.io/luxon/#/formatting?id=table-of-tokens) tokens:

| Old dashboard  | New dashboard |
| ------------- | ------------- |
| HH:mm:ss  | HH:mm:ss  |
| HH:mm  | HH:mm  |
| Y-M-D  | y-L-d  |
| D/M  | d/L  |
| dd HH:mm  | ccc HH:mm  |

Both the info panel and the readme page contain a link to the overview of all available tokens.

As soon as a new format is deployed, it will be showed without need for refreshing.

I have ***not*** implemented a way to allow this property to be changed dynamically via an input message, because the whole dynamic properties mechanism is bit too recent.

## Related Issue(s)

Closes [912](https://github.com/FlowFuse/node-red-dashboard/issues/912)

## Checklist

 - [ X ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ X ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/nodes/widgets/ui-chart.md
+++ b/docs/nodes/widgets/ui-chart.md
@@ -13,6 +13,9 @@ props:
     Point Shape: Define the shape of the point shown in Scatter & Line charts.
     Point Radius: Define the radius (in pixels) of each point rendered onto a Scatter or Line chart.
     X-Axis Type: <code>Timescale</code> | <code>Linear</code> | <code>Categorical</code>
+    X-Axis Format: <code>HH:mm:ss</code> | <code>HH:mm</code> | <code>YYYY-MM-DD</code> | <code>DD/MM</code> | <code>dd HH:mm</code> | <code>Custom</code> | <code>Auto</code>
+        Defines how the values are displayed on the axis, when X-Axis type is <code>'timescale'</code>.
+        See <a target="_blank" href="https://moment.github.io/luxon/#/formatting?id=table-of-tokens">here</a> for an overview of all available Luxon tokens.
     X-Axis Limit: Any data that is before the specific time limit (for time charts) or where there are more data points than the limit specified will be removed from the chart.
     Properties:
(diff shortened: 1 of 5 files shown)


joepavitt (approved):
Other than having a quick back/forth on the above comment - this is excellent - thanks @bartbutenaers

bartbutenaers:
@joepavitt,

I completely agree that the `L` token is very confusing for indicating a month.

After I had made my first implementation - partly based on the code from the old dashboard - the displayed formats were completely not what I had expected.  Until I found out that the old dashboard uses a different token library.

1. Using the `M` only in the dropdown labels, seems very confusing to me (like you already say).  People might create a custom format, and most probably start from an existing format string from the dropdown.  And that won't work...

2. Replacing the `M` by `L` from a custom format might work.  But then you can't refer to the luxon token table in the documentation.  By then you need to add your own (luxon based) token table in the the documentation.  Not sure if it is worth all the efforts to be honest...

3. Would it help if we add an example after each dropdown option?

joepavitt:
I agree us forcing in an `M` isn't going to help given the inconsistency that enforces with the documentation. You link well to the docs - and so let's keep it as you have it for now, and move forward based on any feedback/ideas received in the future.

Issue #882, opened 2024-05-22
Investigate failing Cypress test on GH


joepavitt:
### Current Behavior

We have some Cypress tests that are fairly consistently failing in the GH automation, but pass consistently when run locally. From what I can gather, it's linked to Websocket traffic not reaching, or the UI reacting/refreshing unprompted, meaning that any effects caused by the socket traffic is undone.

### Expected Behavior

Tests should behave the same locally and in GH autoamtions

### Steps To Reproduce

- Locally, you can just run the Cypress tests, they'll pass.
- You can then inspect the runs for _any_ PR done on Dashboard over the past couple of months to find the `controls.spec.js` tes(s) that fail regularly.

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #964, opened 2024-06-12
UI Slider: provide configuration options for slider colours

github-cyprien:
## Description

Add colors to sliders, it's possible to adjust all colors, thumb colors and track colors (see https://vuetifyjs.com/en/components/sliders/#colors)
There is a problem with thumb color that is keeping blue, do you have any ideas ?


github-cyprien:
I will add documentation tomorrow...

github-cyprien:
Don't understand why e2e-tests is failing with the documentation commit...

joepavitt:
> I don't understand why e2e-tests is failing with the documentation commit... Any ideas ?

Our E2E's are sensitive (see #882) - it's nothing personal to this PR

github-cyprien:
But there is still the problem of the thumb color that is not changing...

joepavitt:
> Have you any ideas ? Perhaps it comes from the css...

Indeed it does:

https://github.com/FlowFuse/node-red-dashboard/blob/2cb8ca9f7f0baece242ab029ab45031b7be21a62/ui/src/stylesheets/common.css#L345-L350

The default colouring theme in Vuetify uses the "accent" as the colour, which means it needs to be the same as the navigation bar, it just doesn't suit Dashboard, so I overrode it.

If you're able to have the fallback to `rgb(var(--v-theme-primary));` when a user hasn't set anything, then we can remove the `v-slider-thumb__surface` from this CSS definition

github-cyprien:
Should be ok now...

nodes/widgets/ui_slider.html
```diff
@@ -145,6 +148,23 @@
                 </div>
             </div>
         </div>
+        <div class="form-row">
+            <label for="node-input-color"><i class="fa fa-adjust"></i> Color</label>
+            <div style="display: grid; grid-template-columns: repeat(3, 1fr); width: 70%">
+                <div>
+                    <span for="node-input-color">main</span>
```
joepavitt, line 155:
```suggestion
                    <span for="node-input-color">slider</span>
```

Think this reads a little better than "main"

ui/src/stylesheets/common.css
```diff
@@ -334,14 +334,14 @@ main {
...
     color: rgb(var(--v-theme-on-background));
 }
-
+/*
 .v-slider-track__background,
 .v-slider-track__fill,
 .v-slider-track__tick,
 .v-slider-thumb__surface {
     background-color: rgb(var(--v-theme-primary));
 }
```
joepavitt, line 343: Just look at this in more detail, the reason the thumb wasn't overriding is because we were setting `background-color` on the thumb, but Vuetify actually sets the `color` field with it's `color` setting. As such, the fix is:

```suggestion
.v-slider-track__background,
.v-slider-track__fill,
.v-slider-track__tick {
    background-color: rgb(var(--v-theme-primary));
}

.v-slider-thumb__surface {
    background-color: rgb(var(--v-theme-primary));
}
```

this way, we're overriding correctly, adn when the `thumb-color` property is then set, it overrides our CSS

ui/src/widgets/ui-slider/UISlider.vue
```diff
@@ -68,6 +81,17 @@ export default {
...
+        const themePrimaryColor = 'rgb(' + getComputedStyle(document.documentElement).getPropertyValue('--v-theme-primary') + ')'
+        if (this.color.trim() === '') {
+            this.dynamic.color = themePrimaryColor
+        }
+        if (this.colorTrack.trim() === '') {
+            this.dynamic.colorTrack = themePrimaryColor
+        }
+        if (this.colorThumb.trim() === '') {
+            this.dynamic.colorThumb = themePrimaryColor
+        }
```
joepavitt, line 94: Because of [how we override the Vuetify theming](https://github.com/FlowFuse/node-red-dashboard/blob/0741c0d76f9df60240dffa1e25d3218a29921187/ui/src/layouts/Baseline.vue#L205) with the NR-defined colour set, this isn't a reliable way of getting the actual value set here...

We also actually don't need to have the dynamic setting here either, as it falls back to `null`, which then uses our CSS fix (left in another review comment) do it's magic.

```suggestion
```

joepavitt (changes requested):

github-cyprien:
Thank you for all these corrections Joe.
I think that there is also a problem : textbox are not big enough. I don't know if we could put these options in a fieldset that is folded by default ?

github-cyprien:
When you type a color like #a5a5a5, it doesn't fit in textbox. But anyway, if it's ok for you, it's ok for me :)

ui/src/stylesheets/common.css
```diff
@@ -334,14 +334,17 @@ main {
...
 .v-slider-track__background,
 .v-slider-track__fill,
-.v-slider-track__tick,
-.v-slider-thumb__surface {
+.v-slider-track__tick {
     background-color: rgb(var(--v-theme-primary));
 }
 
+.v-slider-thumb__surface {
+    background-color: rgb(var(--v-theme-primary));
```
joepavitt, line 345:
```suggestion
   color: rgb(var(--v-theme-primary));
```

joepavitt:
Just merged locally with `main` to check if the `src` error is still present, and it isn't, so this'll be good to go once we uncomment the CSS and change `background-color` to `color` as previously requested

github-cyprien:
Thank you for the merge. Didn't understood everything with your CSS and with your strange error, but it's fine :)

joepavitt (approved):
All working perfectly, thanks for the PR @github-cyprien

PR #981, opened 2024-06-15
Widget: File Upload

### LINKED ISSUE #44: Widget: File Upload
### Description

Could not see this in the existing Dashboard 1.0. I suspect this is because there is no guarantee that Node-RED has access to a local file system.

### Properties

- Labelling
- Style
- Behaviour

### Events

- **on-action**: this widget would need to send a message back to Node-RED when it's action is triggered, e.g. button clicked
- **on-change**: this widget would need to send a message back to Node-RED when it's value is changed/updated

### Controls

- **enabled**: this widget can be enabled/disabled
- **visible**: this widget can be hidden

### Existing Examples

Vuetify Component: https://vuetifyjs.com/en/components/file-inputs/

joepavitt:
## Description

- Adds new "File Input" widget. Seen lots of people recently trying to implement their own, so we should be providing this out of the box really.
- Provides good collection of properties to get started with:
    - icon
    - label
    - allow multiple
    - accept file types)

Todo:
- Need documentation
- Need to chat with @Steve-Mcl about alternative comms methods to Node-RED, and consequences of of events architecture. Websockets for file uploading isn't ideal as no tracking of upload progress, and we don't get feedback on success. HTTP is obvious choice, but then that doesn't fit with any other node type.

## Related Issue(s)

Closes #44

### DIFF
--- a/ui/src/widgets/ui-file-input/UIFileInput.vue
+++ b/ui/src/widgets/ui-file-input/UIFileInput.vue
@@ -1,0 +1,142 @@
+<template>
+    <div class="nrdb-ui-file-input">
+        <v-file-input
+            v-if="!uploading && !uploaded"
+            v-model="files"
+            density="compact"
+            :disabled="!state.enabled"
+            :label="label" :prepend-icon="icon"
+            :accept="accept"
+            :show-size="showFileSize"
+            variant="outlined"
+            hide-details="auto"
+        />
+        <div v-else-if="!uploaded" class="nrdb-ui-file-input--progress">
+            <v-progress-linear :model-value="progress" :height="24">
+                <template #default="{ value }">
+                    <strong>{{ Math.ceil(value) }}%</strong>
+                </template>
+            </v-progress-linear>
+        </div>
+        <div v-else class="nrdb-ui-file-input--uploaded">
+            <v-icon icon="mdi-check-circle" color="success" />
+            <label class="v-label">File Uploaded!</label>
+            <label class="v-label" @click="reset()">
+                <a class="nrdb-anchor">Upload Another File</a>
+            </label>
+        </div>
+        <v-btn variant="flat" :disabled="!files || uploading || uploaded" @click="upload(files)">
+            {{ uploading ? 'Uploading...' : 'Upload' }}
+        </v-btn>
+    </div>
+</template>
+
+<script>
+
+import { useDataTracker } from '../data-tracker.mjs' // eslint-disable-line import/order
+import { mapState } from 'vuex' // eslint-disable-line import/order
+
+export default {
+    name: 'DBUIFileInput',
+    inject: ['$socket'],
+    props: {
+        id: { type: String, required: true },
+        props: { type: Object, default: () => ({}) },
+        state: { type: Object, default: () => ({}) }
+    },
+    setup (props) {
+        useDataTracker(props.id)
+    },
+    data () {
+        return {
+            files: null,
+            uploading: false,
+            uploaded: false,
+            progress: 0
+        }
+    },
+    computed: {
+        ...mapState('data', ['messages']),
+        icon: function () {
+            const icon = this.props.icon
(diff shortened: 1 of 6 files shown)


joepavitt:
Agreed with Steve that steps will be as follows:

- Create a `maxFileSize` property on the node that is driven by the `RED.settings` max file size for socketIO, with the default size if no override is presented.
- This will be enforced client-side to ensure SocketIO connection isn't broken
- As a follow-on task, we will then look into chunking Websocket traffic for the file transfer, which would enable better progress reporting client-side, and enable larger files to be transferred

ui/src/widgets/ui-file-input/UIFileInput.vue
```diff
@@ -0,0 +1,161 @@
...
+            return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i]
+        },
+        reset: function () {
+            this.files = null
+            this.uploading = false
+            this.uploaded = false
+            this.progress = 0
+        },
+        upload: function (file) {
+            if (file && !this.multiple) {
```
Steve-Mcl, line 94: I cannot see how multiple files are uploaded Joe?

ui/src/widgets/ui-file-input/UIFileInput.vue
```diff
@@ -0,0 +1,161 @@
...
+        },
+        send (msg) {
+            this.$socket.emit('widget-send', this.id, msg)
+        },
+        maxFileSize (files) {
+            if (files) {
+                const size = files[0]?.size
+                const maxSize = this.props.maxFileSize
+                return !files || !files.length || size < maxSize || `File size should be less than ${this.formatFileSize(maxSize)}!`
+            }
```
Steve-Mcl, line 134: doesnt compute all files in multiple selection

ui/src/widgets/ui-file-input/UIFileInput.vue
```diff
@@ -0,0 +1,161 @@
...
+            </v-progress-linear>
+        </div>
+        <div v-else class="nrdb-ui-file-input--uploaded">
+            <v-icon icon="mdi-check-circle" color="success" />
+            <label class="v-label">File Uploaded!</label>
+            <label class="v-label" @click="reset()">
+                <a class="nrdb-anchor">Upload Another File</a>
+            </label>
+        </div>
+        <v-btn variant="flat" :disabled="!files || uploading || uploaded || maxFileSize(files)" @click="upload(files)">
```
Steve-Mcl, line 29: `files` is an array

```suggestion
        <v-btn variant="flat" :disabled="!files?.length || uploading || uploaded || maxFileSize(files)" @click="upload(files)">
```

Steve-Mcl (commented):
Joe, it kinda looks like multiple upload was disabled (through the code) but the option appears to still be visible and is mentioned in the help.

can you clarify please?

ui/src/widgets/ui-file-input/UIFileInput.vue
joepavitt: yep - good point, I got caught in two minds about `multiple`, and appears I forgot i was halfway with it - will address

joepavitt: Addressed in https://github.com/FlowFuse/node-red-dashboard/pull/981/commits/1fe18a05881f60fd743016f375281a02e8f3cb81

ui/src/widgets/ui-file-input/UIFileInput.vue
joepavitt: Addressed in https://github.com/FlowFuse/node-red-dashboard/pull/981/commits/1fe18a05881f60fd743016f375281a02e8f3cb81

joepavitt:
@Steve-Mcl have fixed/implemented the multiple file upload, and updated docs to reflect the detail that it'll send multiple `msg` events

Steve-Mcl (approved):

PR #1008, opened 2024-06-21
UI Page - Better handling of dropdown boolean types for visibility

### LINKED ISSUE #1004: Dashboard Page set to `hide` after updating parameters
### Current Behavior

After editing a parameter for a page, after clicking the `update` button, the page automatically sets to `hide`.

Recording2024-06-20203018-ezgif com-video-to-gif-converter

Check forum post:
https://discourse.nodered.org/t/page-set-to-hide-after-updating-parameters/88824

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

joepavitt:
## Description

Because NR/html jquery dropdown doesn't support boolean types, we fell into some type inconsistency traps. This handles those cases where older pages may have different visibility/disabled values.

## Related Issue(s)

Closes #1004

### DIFF
--- a/nodes/config/ui_page.html
+++ b/nodes/config/ui_page.html
@@ -69,12 +69,12 @@
             }
 
             // backwards compatibility
-            if (this.disabled === undefined || this.visible === 'false' || this.disabled === false) {
-                this.disabled = false
-                $('#node-config-input-disabled').val('false')
-            } else if (this.disabled === 'true' || this.disabled === true) {
+            if (this.disabled === 'true' || this.disabled === true) {
                 this.disabled = true
                 $('#node-config-input-disabled').val('true')
+            } else {
+                this.disabled = false
+                $('#node-config-input-disabled').val('false')
             }
         },
         oneditsave: function () {
--- a/nodes/config/ui_page.html
+++ b/nodes/config/ui_page.html
@@ -59,15 +59,22 @@
                 typeField: '#node-input-layoutType'
             })
 
-            // backwards compatibility
-            if (this.visible === undefined || this.visible === 'true') {
+            // backwards compatibility - because NR can't do boolean values on dropdowns
+            if (this.visible === undefined || this.visible === true || this.visible === 'true') {
                 this.visible = true
                 $('#node-config-input-visible').val('true')
+            } else if (this.visible === false || this.visible === 'false') {
+                this.visible = false
+                $('#node-config-input-visible').val('false')
             }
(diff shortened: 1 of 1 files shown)


nodes/config/ui_page.html
```diff
@@ -59,15 +59,22 @@
...
-            // backwards compatibility
-            if (this.visible === undefined || this.visible === 'true') {
+            // backwards compatibility - because NR can't do boolean values on dropdowns
+            if (this.visible === undefined || this.visible === true || this.visible === 'true') {
                 this.visible = true
                 $('#node-config-input-visible').val('true')
+            } else if (this.visible === false || this.visible === 'false') {
+                this.visible = false
+                $('#node-config-input-visible').val('false')
             }
```
Steve-Mcl, line 69:
```suggestion
            if (this.visible === false || this.visible === 'false') {
                this.visible = false
                $('#node-config-input-visible').val('false')
            } else {
                this.visible = true
                $('#node-config-input-visible').val('true')
            }
```

while I realise by normal means `.visible` will be `undefined`, `"false"` or `false`, corruption and LLMs can (and do) put random guff like `null` or `"True"` as the value.

If we re-arrange the logic to have a "fallback `} else { ... }` then it will handle this case and others (and is somewhat more readable)

nodes/config/ui_page.html
```diff
@@ -59,15 +59,22 @@
...
+
             // backwards compatibility
-            if (this.disabled === undefined || this.visible === 'false') {
+            if (this.disabled === undefined || this.visible === 'false' || this.disabled === false) {
                 this.disabled = false
                 $('#node-config-input-disabled').val('false')
+            } else if (this.disabled === 'true' || this.disabled === true) {
+                this.disabled = true
+                $('#node-config-input-disabled').val('true')
             }
```
Steve-Mcl, line 78:
```suggestion
            if (this.disabled === 'true' || this.disabled === true) {
                this.disabled = true
                $('#node-config-input-disabled').val('true')
            } else {
                this.disabled = false
                $('#node-config-input-disabled').val('false')
            }
```

Same as before.

Steve-Mcl (commented):
I think a swapparoo on the logic makes this more readable and fail safe Joe.

Steve-Mcl (approved):

PR #1014, opened 2024-06-23
UI Slider - Add Thumb "always", ticks & vertical sliders

joepavitt:
### Always Show Thumb

Screenshot 2024-06-23 at 08 46 55

### Tick Demo

Screenshot 2024-06-23 at 08 46 17

### Vertical Sliders:

Screenshot 2024-06-23 at 08 39 46

## Related Issue(s)

Closes #372 
Closes #624

PR #1015, opened 2024-06-23
UI Slider - Add icons before and after

github-cyprien:
## Description

Add icons to Sliders. Click option make icons clickable to change the value by step.

## Related Issue(s)

No PR.

Two small problems 

- when i read this.props.step in UISlider.vue, it returns me a text, so i need to use **parseInt(this.props.step)** 
- If you don't check click option, icons are clickable but doing nothing. It could be better to make a conditional @click:prepend="clickPrepend" but i don't know how to do that... [Something like](https://stackoverflow.com/questions/48042274/conditional-event-binding-vuejs) @[isClickable&&`click`]="clickHandler" could be ok ?

Hope you will not have any merging problems with PR #1014


joepavitt:
I'm not going to merge the `click` behaviour, as it feels a little out of place vs how we handle icons on other widgets. Also, in testing it locally, it didn't seem to work too well, but I really like the `icon` work, and don't want to hold that up.

RE: Icon work, I'm going to copy in the pattern we use in other widgets too, that means it's optional to include the `mdi-` as part of the icon name

joepavitt:
Although having just seen your documentation, I know understand the idea of the icons (adding a +/- button) - open to the idea, but it would need to be in a follow up PR so we can assess it properly as a standalone feature

joepavitt (approved):

PR #1016, opened 2024-06-23
UI Form - Dynamic options & label

### LINKED ISSUE #1000: Dynamic Properties: UI Form
Capturing this as a single issue to represent a request that's come up multiple times:

- https://github.com/FlowFuse/node-red-dashboard/issues/984
- https://github.com/FlowFuse/node-red-dashboard/issues/287
- https://github.com/FlowFuse/node-red-dashboard/issues/253
- https://github.com/FlowFuse/node-red-dashboard/issues/967
- https://github.com/FlowFuse/node-red-dashboard/issues/363

Will close those issues out and use this issue as primary source going forward

joepavitt:
## Description

- Adds dynamic `options` and `label` fields for UI Form
- Allows setting of values via `msg.payload.<property>` injection
- Update documentation
- Adds E2E Tests

## Related Issue(s)

Closes #1000

### DIFF
--- a/ui/src/widgets/ui-form/UIForm.vue
+++ b/ui/src/widgets/ui-form/UIForm.vue
@@ -92,7 +92,6 @@ export default {
         },
         onInput (msg) {
             if (msg.payload) {
-                console.log('form input', msg.payload, this.input)
                 const payload = msg.payload
                 for (const key in payload) {
                     this.input[key] = payload[key]
--- a/ui/src/widgets/ui-form/UIForm.vue
+++ b/ui/src/widgets/ui-form/UIForm.vue
@@ -51,10 +51,10 @@ export default {
     computed: {
         ...mapState('data', ['messages']),
         label: function () {
-            return this.dynamic.label || this.props.label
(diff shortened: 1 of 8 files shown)


ui/src/widgets/ui-form/UIForm.vue
```diff
@@ -38,17 +38,28 @@ export default {
...
+                label: null,
+                options: null
+            }
         }
     },
     computed: {
-        ...mapState('data', ['messages'])
+        ...mapState('data', ['messages']),
+        label: function () {
+            return this.dynamic.label || this.props.label
```
Steve-Mcl, line 54: This will not permit user to have an empty label. 

As per other nodes (ui dropdown, ui slider), we should allow empty string) e.g.

`return this.dynamic.label !== null ? this.dynamic.label : this.props.label`

That is assuming the default value is `null`

This really does back up the need for common/shared helper methods as was mentioned in another recent PR in similar situation.

Steve-Mcl (approved):
looks fine joe (but not had time to pull and try out like i normally do)

Issue #372, opened 2023-11-24
Slider "Show thumb label" addition


i8beef:
### Description

It appears that the underlying vue.js control for the slider allows for a boolean OR `always` for `thumb-label`. I would love to be able to set it to this third value so the slider is ALWAYS displaying the current value. I would assume this would require changing the "Show thumb labels" setting to a dropdown or radio for the third value?

### Have you provided an initial effort estimate for this issue?

I can not provide an initial effort estimate

joepavitt:
Just to summarise the request here:

**before:** true/false
**after:** true/false/always

i8beef:
Yessir, if I'm reading the docs for vuetify right... basically I'd like to be able to see the current value for any slider, and that seemed to fit with the old dashboard's setting to always show the value "thumb".

i8beef:
Just adding a link to the docs on this: https://vuetifyjs.com/en/components/sliders/#thumb

This will allow you to see what the actual numeric value currently is / will be when released, etc. and would make the sliders easier to use overall.

i8beef:
Thank you that works for the thumb!

FYI though I think we might have a regression in "step", it doesn't seem to work anymore.

joepavitt:
Thanks for flagging, I can see the problem, will fix it in the next couple of hours

i8beef:
Looks good again, thank you!

PR #1023, opened 2024-06-25
UI Slider: Ensure we have defined step as a dynamic variable

joepavitt:
## Description

Fix missing declaration for `step`

## Related Issue(s)

Problem was introduced via a web-based commit suggestion in https://github.com/FlowFuse/node-red-dashboard/pull/1014 and mentioned to us in https://github.com/FlowFuse/node-red-dashboard/issues/372#issuecomment-2187694616


Steve-Mcl (approved):
aaaarrrrgggghhhh - didnt notice this wasnt defined already!

Issue #1036, opened 2024-06-27
UI-Slider: Visual/Display issue when `Thumb` is set to `Always`


arturv2000:
### Current Behavior

If having the option `Thumb` set to `Always` there are some issues regarding the widget size

![image](https://github.com/FlowFuse/node-red-dashboard/assets/17281826/b47cd204-4153-4a80-b0e3-c25fc0312bbf)

It will overlap with the previous component or group border.

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

joepavitt:
So, in the overlap of the group, the obvious solution is to show it above. For the other use cases though, what would you hope to see instead?

arturv2000:
When it is configured to `always` maybe the widget `height` should be increased to be able to accommodate this element.

I know this will give different heights to different rows, but that already happens depending on widget type.

hotNipi:
I'v taken the anatomy of the slider into bits and pieces. 
Why it renders out of boundaries is caused by the row height which for dashboard is 8px smaller than vuetify standard. 
Known that vuetify at least tries to serve all equally but for Node-RED dashboard the industrial kind of usage has to be valued more. That often requires much more condense layouts than a blog or news site.

There is a bit of waist for horizontal slider as for default slider there is no use of area for messages.
![image](https://github.com/FlowFuse/node-red-dashboard/assets/35291460/8b559eda-2764-4801-b93b-58ff41389298)
That can be forced down to 0 so the slider by nature will be rendered to the center, not that 6px too high.

But I don't really think that the thumb label overflow is that as big issue.   And I'm against different row heights. I think most of the users don't like it. 

Bigger issue is introduced with the ticks currently, I have tried to figure out some kind of solution but it's pretty hard. 

Current solution renders tick per step thus if you have min=0, max=100 step =1, there will be 100 ticks overlap each over if the slider is not very long. And it is not responsive also. Overlap is unavoidable.
And there's also a bit of CSS which breaks first and last tick placement.

The tick count can't be told to the slider other that with setting tick labels. Label can be of course empty string so nothing but ticks are then visible. But to provide some kind of smartness behind it (based on min max and step) will be trap. Too many edge cases so I'm afraid it should be given to user to configure

```
 tickLabels: {
                    100: '100',
                    200: '',
                    300: '',
                    400: '',
[...]

PR #1047, opened 2024-07-01
fix slider ticks positioning and allow easy user level customization

hotNipi:
## Description

With a little bit different CSS treatment the slider ticks are now positioned correctly. 
But pay attention, the strategy stands on the `tick-size` property being equal to `track-size` which then both turned to CSS variables by vuetify. 

The user can override CSS (variables) to change the appearance of the ticks. Example added.

This change doesn't solve the problem with count of the ticks.

## Related Issue(s)

There's no direct issue rised but it is closely related to https://github.com/FlowFuse/node-red-dashboard/issues/1036

## Checklist

 - [ X] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [X ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-slider/UISlider.vue
+++ b/ui/src/widgets/ui-slider/UISlider.vue
@@ -3,7 +3,7 @@
 <template>
     <v-slider
         v-model="value" :disabled="!state.enabled" :label="label" hide-details="auto"
-        :class="className" :thumb-label="thumbLabel"
+        :class="className" :style="`--nrdb-slider-track-color:${colorTrack};`" :thumb-label="thumbLabel"
         :min="min" :direction="direction"
         :tick-size="4" :track-size="4"
         :color="color" :track-color="colorTrack" :thumb-color="colorThumb"
--- a/ui/src/widgets/ui-slider/UISlider.vue
+++ b/ui/src/widgets/ui-slider/UISlider.vue
@@ -3,7 +3,7 @@
 <template>
     <v-slider
         v-model="value" :disabled="!state.enabled" :label="label" hide-details="auto"
-        :class="className" :style="`--nrdb-slider-track-color:${colorTrack};`" :thumb-label="thumbLabel"
+        :class="className" :style="`--nrdb-slider-track-color:${colorTrack};--nrdb-slider-tick-scaleY:${tickScaleY};--nrdb-slider-tick-scaleX:${tickScaleX};`" :thumb-label="thumbLabel"
         :min="min" :direction="direction"
         :tick-size="4" :track-size="4"
         :color="color" :track-color="colorTrack" :thumb-color="colorThumb"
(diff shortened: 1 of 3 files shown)


joepavitt:
Could you share a before/after please Endel? I made improvement to tick sizings here last week, which may have been missed: https://github.com/FlowFuse/node-red-dashboard/pull/1014

hotNipi:
Before:
image

After:
image

Issues with before:
- Last tick is rendered where first tick should go
- Tick shabe is stretched from circle thus not sharp
- Colors for active part make tick disappear outside of track
- Customizing via CSS is nightmare cos the positional tretment is not created wisely

hotNipi:
I'll try to explain my choices a bit more detail.
 
 The vuetify original provides ticks on top of the track and with color chage for filled part of the track. Filled color if match with the grop background dosen't let ticks to be just extended. So the most easy will be to change filled color. (Which you did) But with that change the original is not possible anymore and it is way too complicated to get that original look and feel back with CSS override. 
 Also as the user can configure color for track - the ticks still use base color. Some may call it a bug. I would.
 There's not any nice way to provide configuration for ticks cos it makes config way too overcrowded and complex. 
 
 I do understand your intention to make ticks more prominent and render out from track by default. But as it steals the vuetify original behavior - for users - do they get more ore less now? CSS override is common practice. It doesn't take too much of time when users will ask for possible solutions to make something look a bit different. (For every bit of DB, not just the slider) And then it will be better if it can be done easily.

joepavitt:
> I do understand your intention to make ticks more prominent and render out from track by default. But as it steals the vuetify original behavior - for users - do they get more ore less now? CSS override is common practice. It doesn't take too much of time when users will ask for possible solutions to make something look a bit different. (For every bit of DB, not just the slider) And then it will be better if it can be done easily.

Very fair point, it would be nice for us to show taller ticks by default as per:

Screenshot 2024-07-04 at 08 33 18

But still allow for the overrides, I'm not a fan of the circular ticks. Also need to check how the new `color` properties on the Slider are impacted here.

Also worth noting in your docs about `v-slider-track__tick--filled` which covers any ticks beneath/left of the thumb and I think makes it more visually appealing.

joepavitt:
> Default tick size: width - half of track size

Half of the track would be _large_ ?

hotNipi:
image

Ignore the colors and numbers but for concept of tick sizing defaults 
 
Track width - 4px 
Filled track width - 6px
Ticks - 12px / 2px

joepavitt:
looks good to me @hotNipi - thanks for the work :+1:

hotNipi:
Some CSS rules turned out quite complicated so if needed, can explain in detail any time.

ui/src/widgets/ui-slider/UISlider.vue
```diff
@@ -47,6 +48,12 @@ export default {
         direction: function () {
             return this.props.height > this.props.width ? 'vertical' : 'horizontal'
         },
+        tickScaleX: function () {
+            return this.props.height > this.props.width ? 3 : 0.5
```
joepavitt, line 52: 3 is quit high a value, especially if lots of ticks? 2 maybe better?

hotNipi: That goes into the territory of art. I'm not good at art at all. So I happily leave this kind of twaks for experts to find out.

joepavitt (approved):

PR #1055, opened 2024-07-02
Docs: Add example on how to map user data in a global/flow context store

joepavitt:
## Description

Realised I had no explicit example of this, and it's quite a major one given the multi-tenancy stuff.

With the new support for injecting `msg.payload` values into a form, this is a neat demo


Steve-Mcl (approved):
nice n clear

Issue #909, opened 2024-05-29
 'widget-load' event sent multiple times to widget


omrid01:
### Current Behavior

In custom dashboard 2.0 nodes, in some scenarios, the widget (client node) receives multiple identical 'widget-load' notifications

### Expected Behavior

Dashboard 2.0 widgets (client nodes) should receive a single 'widget-load' notification from the server (when there is data in the Node-red data store for this node).

### Steps To Reproduce

To avoid external influences, the reproduction can be done on a clean `ui-example` node.

1. change the server node (`ui-example.js`) to set some data to the datastore upon node startup (to ensure that it will send 'widget-load' events to the clients: 
```
if (!base.stores.data.get(node.id))
   base.stores.data.save(base, node, "XYZ");
```
2. Set a counter & printout in the widget code (`UIExample.vue`)
```
mounted () {
   this.$socket.on('widget-load:' + this.id, (msg) => {
      this.evCount++;
      console.log(`widget-load: node=${this.id},msg=${msg},count=${this.evCount},timestamp=${Date.now()}`);
      return;
...   
```
3. When opening or refreshing a new page, the notification arrives once, as expected:
![image](https://github.com/FlowFuse/node-red-dashboard/assets/100078999/6f3ab808-3fcf-4837-84a5-7fbd61866324)

4. However, if the client stays open, and we restart the Node-red server process, or deploy/restart flows in the editor, we get multiple notifications (the number of notifications increases on every deploy, and their timestamps go backwards):
![image](https://github.com/FlowFuse/node-red-dashboard/assets/100078999/8ed154a5-feac-4eaa-b626-9e99ac894216)

5. 

### Environment

- Dashboard version: 2.0, 1.10.0
- Node-RED version: 3.19
- Node.js version: 20.10.0
- npm version: 10.2.3
- Platform/OS: Windows
- Browser: Chrome, Edge

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

omrid01:
While checking this, it would be nice to modify the 'widget-load' notification behavior, so that it will always be sent (even if the datastore is empty, in which case it will have a null msg).
This shall allow the widget to rely on this notification, and do initialization tasks at one point in time, based on the data (or lack of it) in the datastore.

colinl:
[...]

Issue #871, opened 2024-05-17
Property `class` not being updated when changing tabs


arturv2000:
### Current Behavior

Hello

This can be a unique situation due to the way i have things implemented in Node-Red.

I have a dashboard with several tabs (up to 6 tabs, 42 groups, and a total of 332 widgets). Since i have devices communicating at a rate of 100ms (up to 3 devices at the same time) i have filters before any widget to only send stuff to the widgets if the value `payload` for that specific widget has changed. This was implemented because i started to see performance degradation of too much traffic in the websocket.

What is happening, that i find strange is that for example:

- I am in tab '1', and a message is received from the device that requires updating a text label in tab `2`
- Node-Red sends the message for the text label correct (I can see the message was sent in the _debug page)
![image](https://github.com/FlowFuse/node-red-dashboard/assets/17281826/ea470a06-4b8a-4e66-af34-61484cb925fe)
- The tab properties still indicate the old `class` value
![image](https://github.com/FlowFuse/node-red-dashboard/assets/17281826/c04bad14-3fc5-4424-b77f-018e7857385a)
- If I change to tab `2` the `text` content is correctly updated, but the class is still the old one (in this case the color of the text is still `orange`), the `tab properties` maintains the same indication as before.

If I force a `refresh` of the page, the correct text and color is presented.

Is this behaviour expected?

At the moment the only workaround seems to have a ui-config (that seems to have also some issues) to detect when the tab is loaded, and issue a `msg.reset = true` to all filter nodes, to force the update of values.

### Expected Behavior

In the same way that the `text` content is updated so the other dynamic properties should be.

### Environment

- Dashboard version: v1.7.1
- Node-RED version: v3.1.7
- Node.js version:
- npm version:
- Platform/OS:
- Browser: Edge

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

colinl:
Can you build a simple example flow showing the problem that we can run to see if we see the same issue?

arturv2000:
This should do it.

Using the inject nodes to change the color of the of the text labels, when viewing the other "page"

```
[...]

PR #1059, opened 2024-07-02
UI-button: Make `ui_update` restore even after tab change / Page refresh

arturv2000:
## Description

Needed a way to preserve the changes we send via `ui_update` even if the widget is hidden in another tab or after a page refresh.

As it was now, depending on the widget, it would only recuperate the last `ui_update` property sent. 
For the moment only did this for the `UI-Button` but if it is correct will make changes to at least `UI-Dropdown`, `UI-Slider` and `UI-Switch`, they are the widgets that at the moment I require this behaviour.

Changes:

- Unsubscribe from event 'widget-load:' to avoid multiple widget-Load messages, seems to solve #909
- When emitting a `widget-load` event, add to the message the `ui_update` parameters in store
- Add `ui_update.enable` as a valid parameter for `UI-Button`
- Add `onLoad` event to `UI-Button`
- Store the last message sent to  `UI-Button` to force a `widget-load` event upon refresh of page/widget.
    - This one, not quite sure about it. I needed a way for the `UI-Button` to trigger a `widget-load` event and this was the only way I could find to do that

This if applied to other widgets should also solve this issue #871

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -678,6 +678,9 @@ module.exports = function (RED) {
                     return
                 }
 
+                // Add ui_update parameter to message to be sent on widget load
+                msg.ui_update = statestore.getAll(id)
+
                 conn.emit('widget-load:' + id, msg)
             }
             // wrap execution in a try/catch to ensure we don't crash Node-RED
--- a/nodes/widgets/ui_button.js
+++ b/nodes/widgets/ui_button.js
@@ -1,3 +1,4 @@
+const datastore = require('../store/data.js')
 const statestore = require('../store/state.js')
 const { appendTopic } = require('../utils/index.js')
 
@@ -16,6 +17,8 @@ module.exports = function (RED) {
             // retrieve the payload we're sending from this button
             let payloadType = config.payloadType
             let payload = config.payload
+
+            console.log('Button Before Send Enter')
 
             if (payloadType === 'flow' || payloadType === 'global') {
                 try {
@@ -62,9 +65,22 @@ module.exports = function (RED) {
                     // dynamically set "label" property
                     statestore.set(group.getBase(), node, msg, 'iconPosition', updates.iconPosition)
                 }
+                if (typeof updates.enabled !== 'undefined') {
+                    // dynamically set "label" property
+                    statestore.set(group.getBase(), node, msg, 'enabled', updates.enabled)
+                }
+                if (typeof updates.visible !== 'undefined') {
+                    // dynamically set "label" property
+                    statestore.set(group.getBase(), node, msg, 'visible', updates.visible)
+                }
             }
+            console.log(statestore.getAll(node.id))
+            msg.ui_update = statestore.getAll(node.id)
+
+            console.log('Button Before Send Exit -> ' + error)
 
             if (!error) {
+                datastore.save(group.getBase(), node, msg)
                 return msg
             } else {
                 node.error(error)
--- a/nodes/widgets/ui_button.js
+++ b/nodes/widgets/ui_button.js
@@ -18,8 +18,6 @@ module.exports = function (RED) {
             let payloadType = config.payloadType
             let payload = config.payload
 
-            console.log('Button Before Send Enter')
-
             if (payloadType === 'flow' || payloadType === 'global') {
                 try {
                     const parts = RED.util.normalisePropertyExpression(payload)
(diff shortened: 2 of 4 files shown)


joepavitt:
This _shouldn't_ be required... When sending the state on `widget-load`, we combine the "Dynamic Properties" (saved in the store) with the Node-RED config for the node, which should consequently load the relevant state as a combined entity

joepavitt:
I'm assuming this is because, currently `ui-button` is an action-driven widget, it doesn't maintain state. However, if we want to maintain a `enabled` state, we need to have this stored server-side to persist it?

arturv2000:
> This _shouldn't_ be required... When sending the state on `widget-load`, we combine the "Dynamic Properties" (saved in the store) with the Node-RED config for the node, which should consequently load the relevant state as a combined entity

It isn't working either for `ui-switch` and `ui-dropdown` they were the three widgets I was testing to check the behaviour, the same was true for other properties, only the last `property` sent to the widget was being passed, since only the last message was being replayed.


> I'm assuming this is because, currently `ui-button` is an action-driven widget, it doesn't maintain state. However, if we want to maintain a `enabled` state, we need to have this stored server-side to persist it?

In this case (and most others i believe) the `enable` is not directly part of the `dynamic properties` it was only in the `props` field. I added to the `dynamic properties` and added also the store server-side in order to persist.

colinl:
> I'm assuming this is because, currently `ui-button` is an action-driven widget, it doesn't maintain state. However, if we want to maintain a `enabled` state, we need to have this stored server-side to persist it?

I believe that msg.enabled does work correctly with ui-button, and survives a refresh.  I am using it without problems.

arturv2000:
> > I'm assuming this is because, currently `ui-button` is an action-driven widget, it doesn't maintain state. However, if we want to maintain a `enabled` state, we need to have this stored server-side to persist it?
> 
> I believe that msg.enabled does work correctly with ui-button, and survives a refresh. I am using it without problems.

The msg.enable survives a page refresh, but doesn't survive a page change (okay maybe not the best expression), it is in the case you are viewing Tab2 and send a msg.enable to a button (it also happens on other widgets and with other properties) on Tab1, when you return to tab1 the msg.enable is not taken in account, in other cases if you send multiple messages when the tab is not visible only the last message is replayed and only the properties that where updated in that last message are applied. 

It is the same situation as #871 but in this case for the msg.enable property.

There is an obvious solution to this, the user must keep track of all the properties / msg.ui_update that are sent to a widget, and check the event of tab change to `resend` it to the widget. But since we have a serve state store, we should make full use of it to simplify the process.

colinl:
> The msg.enable survives a page refresh, but doesn't survive a page change

It seems you are correct, though I cannot now understand why I don't see that in my 'production' system where it appears to work as it should, so I will have to investigate that.
In my nodes I maintain the complete state in the server state store, so that the client side gets the full state on refresh. When a message comes in I don't save the complete message in the state store, I just update the store with any data from the message. Then the message is passed to the client, and it uses that to update its copy of the server side state.
I have not yet worked out how to update the server state from the client, but so far that has not been necessary.  When I send a message from the client I always use the mode that does not update the store. I think I will have to implement a custom event on the socket to send an update to the server.

arturv2000:
> In my nodes I maintain the complete state in the server state store, so that the client side gets the full state on refresh. When a message comes in I don't save the complete message in the state store, I just update the store with any data from the message. Then the message is passed to the client, and it uses that to update its copy of the server side state.


That is what I tried to do, that when a `widget-load` event is generated in `nodes/config/ui_base.js` to go the store and get the `ui_update` that are stored for that node

```js
msg.ui_update = statestore.getAll(id)
```
And send that message to the widget
And it was necessary to also include a `onLoad` method on the widget (Vue code) in order to process the message.


> 
> Edit: Just a note to say that I think the docs can be confusing in this area. When the browser is refreshed it does not resend the last message to the client, it sends the current contents of the server data store, which may not be the same thing.

Well, I think that is the latest message (sent or received, not quite sure, is one of that things that my brain finds confusing... too much habit of only writing/seeing C code), at least is what i could understand by what i saw on the websocket connection and some console.logs that I included in some points. 
From what i can understand access to all the content in the server store, is only made in `node.register` function, don't know exactly when it is called, and is passing the store contents to a 'state' variable.

colinl:
> Well, I think that is the latest message (sent or received, not quite sure

I am taking about third party nodes.  I think that in the core nodes the data store is always updated with the latest message in or out, and it seems to me that this is the cause of the problems.  In third party nodes it is up the author to decide how to handle the state store.  The technique I have used is to make sure that the data store always has the latest complete state information in it.

arturv2000:
Close, since #1122 should handle this in a better / more organized way.

Issue #393, opened 2023-11-24
UI Text Input - Tooltip


joepavitt:
### Description

Dashboard 1.0 offers an option to define a "tooltip" for a `ui-text-input`.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

kd9lsv:
It looks like this issue can be closed.

Resolved #1064

PR #1064, opened 2024-07-05
UI Test Input - Add "tooltip" option

chrblabla:
## Description

As requested in #393, a text input field now has a property called `tooltip` which can be configured. If configured, it will show a tooltip on hover in the dashboard.
For the implementation, the [tooltip component from Vuetify](https://vuetifyjs.com/en/components/tooltips) has been used.

## Related Issue(s)

#393

## Checklist

    - [ ] Upgrade instructions
    - [x] Configuration details
    - [ ] Concepts

## Labels


chrblabla:
Example:

Bildschirmfoto 2024-07-04 um 18 32 01

joepavitt:
Thanks for the PR @chrblabla - this works great with a "tooltip" defined, but breaks if one is not... I'll try and put in a quick fix so we can get it merged

chrblabla:
Hi @joepavitt, thank you for your comment. Do you mind elaborating what exactly breaks?

joepavitt:
For migrating from older versions, there will be existing `ui-text-input` nodes that have `props.tooltip` as `undefined`, so the `tooltip.length` on your `:disabled` logic falls over.

I'm putting in a fix as we speak

ui/src/widgets/ui-text-input/UITextInput.vue
```diff
@@ -1,14 +1,20 @@
...
-        :disabled="!state.enabled" class="nrdb-ui-text-field"
-        :label="label" :type="type" :rules="validation" variant="outlined" hide-details="auto" @update:model-value="onChange" @keyup.enter="onEnter" @blur="onBlur"
-    />
-    <v-textarea
-        v-else
-        v-model="value" :disabled="!state.enabled" class="nrdb-ui-text-field"
-        :label="label" variant="outlined" hide-details="auto" @update:model-value="onChange" @blur="send"
-    />
+    <v-tooltip :text="tooltip" :disabled="tooltip.length === 0" location="bottom">
+        <template v-slot:activator="{ props }">
```
joepavitt, line 3:
```suggestion
    <v-tooltip :text="tooltip" :disabled="!tooltip?.length" location="bottom">
        <!-- eslint-disable-next-line vue/no-template-shadow -->
        <template v-slot:activator="{ props }">
```

Safer logic check for older `ui-text-input` nodes that would have no `tooltip` defined. Also added a `eslint-disable` as it's complaining about the use of `props` here (which is the correct usage)

joepavitt: @chrblabla if you can merge/accept this suggestion, then it's good to go. I don't have permissions to push to your branch directly, so couldn't resolve it myself

chrblabla:
@joepavitt thank you for the explanation and good catch 🙂 I accepted the suggestion and updated the branch to also reflect the recently merged icon feature.

joepavitt (approved):

PR #1077, opened 2024-07-10
UI-Dropdown: Use `v-select` option instead of `v-combox`

arturv2000:
## Description

This allows the user to use the vue component `v-select` instead of `v-combox`.

The main difference is that `v-select` is not "writable". Specifies that the list is displayed by clicking the down arrow and that the text portion is not editable. This means that the user cannot enter a new value. Only values already in the list can be selected

This prevents that the user can type an invalid value (that do not exist on the list and sending payload as null as result). It allows typing for filtering values.

The text description in the config menu could be better, but having no ideia how. Other option is to use the v-select instead of v-combox, but this could be tricky for legacy reasons.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/nodes/widgets/ui-dropdown.md
+++ b/docs/nodes/widgets/ui-dropdown.md
@@ -18,8 +18,8 @@ props:
     Clearable:
         description: Clear selection with button.
         dynamic: false
-    Use v-select:
-        description: Use <code>v-select</code> component. Specifies that the list is displayed by clicking the down arrow and that the text portion is not editable. This means that the user cannot enter a new value. Only values already in the list can be selected. Allows typing to filter possible results.         
+    Allow Search:
+        description: Allows user to type a new value, filtering the list of possible values to choose.         
 dynamic:
     Label:
         payload: msg.ui_update.label
--- a/docs/nodes/widgets/ui-dropdown.md
+++ b/docs/nodes/widgets/ui-dropdown.md
@@ -7,7 +7,7 @@ props:
         description: The text shown to the left of the dropdown.
         dynamic: true
     Options:
-        description: A list of the options available in the dropdown. Each row defines a `label` (shown in the dropdown) and `value` (emitted on selection) property.
+        description: A list of the options available in the dropdown. Each row defines a 'label' (shown in the dropdown) and `value` (emitted on selection) property.
         dynamic: true
     Allow Multiple:
         description: Whether or not a user can select multiple options, if so, checkboxes are shown, and value is emitted in an array.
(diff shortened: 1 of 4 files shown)


joepavitt:
I'd actually go further here - and replace `combobox` with `select` entirely as an iteration/ If I see requests come in for being able to define custom values in this, then we can re-introduce the `combobox` as an option, but surface it as "allow custom input", rather than exposing any concept of Vuetify components to end-users at the Node-RED level

arturv2000:
> I'd actually go further here - and replace `combobox` with `select` entirely as an iteration.

Ok, I can make that change, and completely replace with `v-select`.


> If we see requests come in for being able to define custom values in this, then we can re-introduce the `combobox` as an option, but surface it as "allow custom input", rather than exposing any concept of Vuetify components to end-users at the Node-RED level

Actually, for that to work more modifications are necessary, as it is now, the user can "write" any value, and it will be present on the UI but it is not pass to node-red.

joepavitt:
> Actually, for that to work more modifications are necessary, as it is now, the user can "write" any value, and it will be present on the UI but it is not pass to node-red.

Even more reason for us to switch over to `v-select` then 😄

joepavitt:
Wait, actually, I just tried the `v-select`, the `v-combobox` I do prefer as it allows you to search your options (if you have hundreds this is _very_ useful)

It doesn't allow definition of a new option, and won't submit with custom values.

In which case, if you do have value is disabling the "search" - let's brand it that way. Have an option surfaced in Node-RED config which is "Allow Search" - enabled by default. If toggled off, we use the standard `v-select`

arturv2000:
`v-select` allows search, but is necessary to write the the value correct from "start", if we have items that start with ´L´ and type `L` it will go to the first `L` start item in the list, but it does not `filter` the probable values like `combobox`.

Maybe the option should be called `Allow Search with Filter`?

Ii is a pity that `combobox` doesn't have a option like `Disable input values not in List` or a error indication if the value is not in the list.

joepavitt:
Ah yes, I do see that now, I think I prefer `combobox` for the moment as the default. I think an "Allow Search" is a good starting point (true by default), and we can iterate from there.

arturv2000:
Updated...

Also added the prop [auto-select-first](https://vuetifyjs.com/en/components/combobox/#props-auto-select-first) when using combo-box, to update the selection (send message to node-red) based on item highlighted / selected via search.

joepavitt (approved):

arturv2000:
Would say yes.

If deselecting the `Allow Search` it will use the `v-select` component that do not allow direct text input

colinl:
OK, I will test it when it has been released.

PR #1080, opened 2024-07-10
Docs: Add UI Template examples for building custom charts with ChartJS

joepavitt:
## Description

- Requested by a client
- Event though we may end up switching over to Apache eCharts, these would still be valid examples for importing ChartJS, and also provide a foundation for working with incoming data for any charting library


docs/nodes/widgets/ui-chart.md
```diff
@@ -435,3 +446,255 @@ const msg = {
 ```
 
 This would render two bars, one labelled `value` with a value of `34`, and one labelled `nested.value` with a value of `12`.
+
+## Building Custom Charts
+
+ChartJS has a rich set of configuration options, of which we only expose a small subsection via the Node-RED configuration. If you want to customise the appearance of your chart further, or even render charts we don't yet support, you can do so with a UI Template node.
+
+Currently, although not ideal, we do need to load the ChartJS library from a CDN, and then watch for the file to have been loaded before we can use it, as per the [Loading External Dependencies](http://localhost:5173/nodes/widgets/ui-template.html#loading-external-dependencies) details in the UI Template documentation.
```
knolleary, line 454:
```suggestion
Currently, although not ideal, we do need to load the ChartJS library from a CDN, and then watch for the file to have been loaded before we can use it, as per the [Loading External Dependencies](/nodes/widgets/ui-template.html#loading-external-dependencies) details in the UI Template documentation.
```

knolleary (approved):
One url to fix, but otherwise passes my quick scan

Issue #1073, opened 2024-07-09
`ui-template` How to get a flow variable?


kevinshane:
Hi noob here, in node red we can get a flow or global variable by using `flow.get("blah")`, however it seems inside `ui-template` I can't read or get a flow value, currently I have to use function node to read a flow variable and then pass that variable as a `msg.blah` to template node. So my question is, how exactly I can read flow variable directly inside a `ui-template` ? Thanks guys!

kevinshane:
> Here, we assign `msg.payload` (could be anything) to a value from the global/flow stores
Thanks for the explaination. If I understanding correctly, when a value is updated and passed to `ui-template` node, it will cause the entire dashboard to re-render (assuming db is build only with `ui-template` node), is it correct?

joepavitt:
>  it will cause the entire dashboard to re-render (assuming db is build only with ui-template node), is it correct?

Not necessarily, it'll just update the components using the `msg.` properties in `ui-template`.

kevinshane:
OK thank you for your explaination.

PR #1083, opened 2024-07-11
Docs: Add example for accessing flow/global data in a ui-template

joepavitt:
## Description

- Adds an explicit mention of the best practice to access `flow`/`global` data inside a `ui-template`
- Also adds some eaders to make finding the `this.$socket.on` and `this.send` functions easier to find

## Related Issue(s)

Adds documentation to cover #1073


docs/nodes/widgets/ui-template.md
```diff
@@ -85,11 +85,23 @@ When accessing the `msg` variable inside a `<script />` tag, you need to prefix
...
 ### Built-in Functions
 
 We also offer some helper functions for the Node-RED integration too:
 
+#### Sending Data
+
 - `this.send` - Send a message to the Node-RED flow. If a non-Object value is sent, then Dashboard will automatically wrap that into a `msg.payload` object.
+
+#### Receiving Data
+
```
Steve-Mcl, line 104: Should we not first mention `watch`ing the `msg` as best practice (over `this.$socket.on`)?

docs/nodes/widgets/ui-template.md
```diff
@@ -85,11 +85,23 @@ When accessing the `msg` variable inside a `<script />` tag, you need to prefix
 
 ***Important Note:*** On first load, `msg.payload` may be `null` or `undefined`, and trying to access a nested property will throw an error. Using the **optional chaining** (?.) operator, e.g. `msg.payload?.nested?.property` will prevent these errors occuring.
 
+#### Accessing Node-RED Global/Flow Context
+
+We are unable to send the full `flow`/`global` context stores from Node-RED to the Dashboard UI, as such, the best practice here is to use a "Change" node prior to the `ui-template` node to assign a `msg.<property>` to the relevant value from the `flow.` or `global.` store:
```
Steve-Mcl, line 90:
```suggestion
The `flow`/`global` context stores are not available in the Dashboard UI, as such, the best practice here is to use a "Change" node prior to the `ui-template` node to assign a `msg.<property>` to the relevant value from the `flow.` or `global.` store:
```

I am a little ineasy with "unable to". Might be better to just say is it not available?

Steve-Mcl (approved):
Approved with a couple of comments. Will leave for you to decide.

docs/nodes/widgets/ui-template.md
joepavitt: It's a good point - there are differences though. The `watch` will fire on load, _and_ on new messages. The detailed approach here will fire only on new inputs. I will document both approaches.

PR #1087, opened 2024-07-12
UI-Button: Add Background, text and icon color option

### LINKED ISSUE #375: UI Button: Colors
### Description

Dashboard 1.0 offers customisation on the text and background colour of the button. Dashboard 2.0 should do this too.

arturv2000:
## Description

Add properties to allow the user to set the color to be used for:
- Button Background Color
- Text Color
- Icon Color

If empty string in any of these properties, the default theme color will be used.

The text color allows the use of [Helper Class](https://vuetifyjs.com/en/styles/text-and-typography/#text-and-typography), for example having the field defined as `red text-decoration-underline text-h3 font-weight-black` results in:

Button Example

Alternatively, for the text part we could allow html injection like it was done to `ui-text`. 

Most if not all of these are possible to be done via `class` using CSS, this is a more "direct"/"user friendly" way of doing it.

## Related Issue(s)

Close #375 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/nodes/widgets/ui-button.md
+++ b/docs/nodes/widgets/ui-button.md
@@ -12,6 +12,15 @@ props:
     Label:
         description: The text shown within the button. If not provided, then the button will only render the icon.
         dynamic: true
+    Button Color:
+        description: Button Color. If not provided, default theme color will be used.
+        dynamic: true
+    Text Color:
+        description: Text (Label) color. If not provided calculated automatically based on Button color to be Black or White.
+        dynamic: true
+    Icon Color:
+        description: Icon color. If not provided, will have the same color as text / label.
+        dynamic: true
     Emulate Button Click: If enabled, any received message will trigger a button click, emitting the relevant payload and topic.
 controls:
     enabled:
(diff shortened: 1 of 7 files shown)


joepavitt:
> Alternatively, for the text part we could allow html injection like it was done to ui-text.

Definitely something I'd like to do

arturv2000:
Just noticed an error in `textColor` since I then added the possibility to use the `Helper Class` it no longer supports using color codes `#555555` or `rgb(127,127,127)`.

> Definitely something I'd like to do

Ok, I can remove the property `textColor` and add `<span v-html="label" />` for the text.

joepavitt (approved):
Changes I've made as part of the review;

- Ensure the `<template>` elements only render if an icon is defined, and if append/prepend icon is present, this was causing icon-only buttons to render off-center.
- Rearrange the Button config so that the styling content renders at the bottom and under a "Custom Styling" sub-section
- Made the placeholders a little lighter to read, the detail is in the docs.
- Change the `icon ??` check to `iconOnly ??`

PR #1090, opened 2024-07-13
UI-Switch: Prevent change state when no `msg.payload` is present, fix `topic` not present at restart

### LINKED ISSUE #366: Switch widget should accept msg.enabled true/false to enable it, as D1 does.
### Description

The Switch widget should accept msg.enabled true or false to enable/disable its operation.

I am no FlowFuse team member

### LINKED ISSUE #456: A Switch widget configured to send a fixed string as the topic does not send the topic, unless a message is sent to its input.
If a Switch is configured with a fixed string as the topic then it does not send the topic unless a message is sent to its input.
To replicate, import the flow below. Do **not** click the inject nodes. Exercise the switch in the UI and it will be seen that the topic is not sent.  Now click one of the inject nodes and thereafter the configured topic will be sent when the switch is exercised.  Note that the input messages have a different topic.  ~~Restarting node red did not seem to be sufficient to show the problem again, I had to delete the node, Deploy, and re-import it.~~

Edit: Using Dashboard version 0.11.1, NR 3.1.3, nodejs 20.5.1

Edit: On retesting with dashboard 1.1.0, NR 3.1.5, I find that the symptom is repeatable on node red restart.

```
[{"id":"43d268a322a00bb8","type":"debug","z":"997da33a0beedade","name":"debug 2473","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","statusVal":"","statusType":"auto","x":790,"y":1020,"wires":[]},{"id":"26801ad4f7fb6aa3","type":"ui-switch","z":"997da33a0beedade","name":"","label":"string topic test switch","group":"7545fdb22e4a388c","order":0,"width":0,"height":0,"passthru":false,"topic":"some/topic","topicType":"str","style":"","className":"","onvalue":"true","onvalueType":"bool","onicon":"","oncolor":"","offvalue":"false","offvalueType":"bool","officon":"","offcolor":"","x":570,"y":1020,"wires":[["43d268a322a00bb8"]]},{"id":"dc38f39b5ded4394","type":"inject","z":"997da33a0beedade","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"another/topic","payload":"true","payloadType":"bool","x":280,"y":1000,"wires":[["26801ad4f7fb6aa3"]]},{"id":"de7f3f046aa9dbf5","type":"inject","z":"997da33a0beedade","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"another/topic","payload":"false","payloadType":"bool","x":290,"y":1060,"wires":[["26801ad4f7fb6aa3"]]},{"id":"7545fdb22e4a388c","type":"ui-group","name":"D2 tests","page":"d7fcdd33578e5d34","width":"6","height":"1","order":-1,"showTitle":true,"className":""},{"id":"d7fcdd33578e5d34","type":"ui-page","name":"d2","ui":"ID-BASE-1","path":"/d2","layout":"flex","theme":"4bff2b59c4c518e1","order":-1,"className":""},{"id":"ID-BASE-1","type":"ui-base","name":"Dashboard","path":"/dashboard"},{"id":"4bff2b59c4c518e1","type":"ui-theme","name":"Another Theme","colors":{"surface":"#ffffff","primary":"#0094ce","bgPage":"#eeeeee","groupBg":"#ffffff","groupOutline":"#cccccc"}}]
```

### LINKED ISSUE #622: Sending enable/disable message to ui-switch destroys the saved msg.topic from previous messages, also the switch is switched off
A ui-switch node can be configured to use msg.topic as the topic to send on a state change.
If an enable/disable message is sent (with no topic) using msg.enabled true or false then the topic saved from a previous message setting the switch state is lost.
I don't think that an enable/disable message should have such side effects.

Edit: Also if the switch is On and is then disabled and enabled then the switch is set to Off.  That should not happen.

arturv2000:
## Description

- Prevents that a message without `msg.payload` can change the switch state, specifically if the message only has `msg.enabled` for example.
- Make sure that `topic` is added to message in case a message is sent by switch toggle before a message was sent to the switch by Node-Red (#456)

## Related Issue(s)

Partial Fix #622 , regarding the topic issue (when is configured to be `msg.topic`), it happens with other widgets, the best option I can think at the moment for this widget is that in the `onInput` handler if the current message does not has a ` topic` to get the `topic` from the last message and add it to this new message, or in this case the `topic` should be handled as a `ui_update` property and saved/restored from `statestore`.

Closes #366 (It was already supported, just to close the issue)

Fix #456 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/nodes/widgets/ui-switch.md
+++ b/docs/nodes/widgets/ui-switch.md
@@ -10,6 +10,10 @@ props:
     Off Icon: If provided, this <a href="https://pictogrammers.com/library/mdi/" target="_blank">Material Design icon</a> will replace the default switch when in "off" state. No need to include the <code>mdi</code> prefix.
     On Color: If provided with a icons, this colour is used for the icon when in "on" state
     Off Color: If provided with a icons, this colour is used for the icon when in "off" state
+controls:
+    enabled:
+        example: true | false
+        description: Allow control over whether or not the switch can be toggled via the UI.
 dynamic:
     Class:
         payload: msg.class
@@ -30,6 +34,10 @@ Adds a toggle switch to the user interface.
 ## Dynamic Properties
 
 <DynamicPropsTable/>
+
+## Controls
+
+<ControlsTable/>
 
 ## Example
 
--- a/nodes/widgets/ui_switch.js
+++ b/nodes/widgets/ui_switch.js
@@ -32,6 +32,10 @@ module.exports = function (RED) {
                 const off = RED.util.evaluateNodeProperty(config.offvalue, config.offvalueType, wNode)
                 msg.payload = value ? on : off
 
+                if (config.topic || config.topicType) {
+                    msg = await appendTopic(RED, config, node, msg)
+                }
+
                 datastore.save(group.getBase(), node, msg)
 
                 // simulate Node-RED node receiving an input
--- a/ui/src/widgets/ui-switch/UISwitch.vue
+++ b/ui/src/widgets/ui-switch/UISwitch.vue
@@ -18,8 +18,10 @@ export default {
         props: { type: Object, default: () => ({}) },
         state: { type: Object, default: () => ({}) }
     },
-    setup (props) {
-        useDataTracker(props.id)
+    data () {
+        return {
+            selection: null
+        }
     },
     computed: {
         ...mapState('data', ['messages']),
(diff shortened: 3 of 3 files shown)


joepavitt (approved):
Just added some docs to cover the use of `msg.enabled` - this is now good to go

Thanks as always @arturv2000

PR #1091, opened 2024-07-13
UI-Text-Input: Prevent text delete when no msg.payload is present

arturv2000:
## Description

- Prevents that a message without `msg.payload` can delete the text, specifically if the message only has `msg.enabled` for example.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-text-input/UITextInput.vue
+++ b/ui/src/widgets/ui-text-input/UITextInput.vue
@@ -112,9 +112,6 @@ export default {
     created () {
         // can't do this in setup as we are using custom onInput function that needs access to 'this'
         useDataTracker(this.id, this.onInput, this.onLoad, null)
-
-        // let Node-RED know that this widget has loaded
-        this.$socket.emit('widget-load', this.id)
     },
     methods: {
         onInput (msg) {
--- a/ui/src/widgets/ui-text-input/UITextInput.vue
+++ b/ui/src/widgets/ui-text-input/UITextInput.vue
@@ -35,12 +35,10 @@ export default {
         props: { type: Object, default: () => ({}) },
         state: { type: Object, default: () => ({}) }
     },
-    setup (props) {
-        useDataTracker(props.id)
-    },
     data () {
         return {
-            delayTimer: null
+            delayTimer: null,
+            textValue: null
         }
     },
     computed: {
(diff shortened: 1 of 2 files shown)


ui/src/widgets/ui-text-input/UITextInput.vue
```diff
@@ -110,7 +109,36 @@ export default {
             }
         }
     },
+    created () {
+        // can't do this in setup as we are using custom onInput function that needs access to 'this'
+        useDataTracker(this.id, this.onInput, this.onLoad, null)
+
+        // let Node-RED know that this widget has loaded
+        this.$socket.emit('widget-load', this.id)
```
joepavitt, line 117: this is handled as part of the data tracker, so shouldn't be required?

ui/src/widgets/ui-text-input/UITextInput.vue
```diff
@@ -110,7 +109,36 @@ export default {
...
+    },
     methods: {
+        onInput (msg) {
+            // update our vuex store with the value retrieved from Node-RED
+            this.$store.commit('data/bind', {
+                widgetId: this.id,
+                msg
+            })
```
joepavitt, line 125: I'd argue here that actually, we only want to store this _if_ a `payload` is present? This encourages the argument for us using `datastore` as a bigger picture state, because here, we'd only get the latest `msg.enabled` and not the underlying `payload`

joepavitt (changes requested):
few thoughts added, but conceptually, this is good.

ui/src/widgets/ui-text-input/UITextInput.vue
arturv2000: Yes, you are probably right, just copied the function from another widget (don't recall which one) that had that. But indeed the `data tracker` on the `onMounted` always call this.

Will fix it.

ui/src/widgets/ui-text-input/UITextInput.vue
arturv2000: It was another copy (honestly don't know what is being stored in Vuex and for what is used), but not all the widgets have the `onInput`.

Will change to only store, if `msg.payload` is present in the message.

arturv2000: Just did some tests, and honestly don't quite understand to point of this.

Tried to pass code the "this.$store.commit" for only if `payload` exists, but the behaviour is the same, if a message arrives to the widget on NodeRed that message is "saved" and replayed later on "widget-load", and the `payload` may not appear if it wasn't in that message (in case of page refresh for example). The message seems to be saved here:
https://github.com/FlowFuse/node-red-dashboard/blob/f0a9cad51cf16655419b4d32ee4bf3f8b56d0bfa/nodes/widgets/ui_text_input.js#L20
And it is replayed, the saving (commit) to the `vuex` store seems to have no effect

joepavitt (approved):

Issue #343, opened 2023-11-15
Form widget not forgetting renamed fields


colinl:
To reproduce the problem
Create a Form with one or more fields and exercise it.
Edit the form and change the label/name of one of the fields and exercise it.
The node sends both the old fields and the new ones.
Restarting node red removes the old fields from the set message.

joepavitt:
Noting that @arturv2000 has already started this, so will mark as such

PR #1092, opened 2024-07-13
UI-Form: Allow `msg.enable`

### LINKED ISSUE #342: Form widget should send Number type fields as Numbers not strings
On D1 the Form node sends fields that are of type Number as Numbers.  With D2 they are sent as strings.  Ideally they should be numbers I think.

### LINKED ISSUE #935: ui-form submit button has to be clicked twice
### Current Behavior

when filling out a form, I cannot click the button first time

chrome_VmEInVT5FG

### Expected Behavior

Button should be clickable 

### Steps To Reproduce

add a form with 1 text entry (required)
deploy
enter text
click button
click button

### Environment

- Dashboard version: 1.11.0
- Platform/OS: Windows
- Browser: chrome

### LINKED ISSUE #1069: ui-form passing prepopulated and submit not clickable
### Current Behavior

When prepopulating form the data is passed through.
Also all fields have to be select after prepopulation before submit button becomes active

### Expected Behavior

Not to pass prepopulated data with out a submit
submit button to be active if all required fields are populated and correct in format.

### Steps To Reproduce

```
[{"id":"0872583c36c11d85","type":"inject","z":"d1395164b4eec73e","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"{\"name\":\"aaa\",\"phone\":\"1234\"}","payloadType":"json","x":110,"y":200,"wires":[["4e2122c1d3d77fd9"]]},{"id":"4e2122c1d3d77fd9","type":"ui-form","z":"d1395164b4eec73e","name":"","group":"9d83cd8b3d5ee33c","label":"","order":1,"width":0,"height":0,"options":[{"label":"Name","key":"name","type":"text","required":true,"rows":null},{"label":"Phone","key":"phone","type":"number","required":true,"rows":null}],"formValue":{"name":"","phone":""},"payload":"","submit":"submit","cancel":"clear","resetOnSubmit":true,"topic":"topic","topicType":"msg","splitLayout":"","className":"","x":310,"y":160,"wires":[["c6708522f7441ae1"]]},{"id":"c6708522f7441ae1","type":"debug","z":"d1395164b4eec73e","name":"debug 2561","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":470,"y":160,"wires":[]},{"id":"9d83cd8b3d5ee33c","type":"ui-group","name":"Group1","page":"c694d0ebe0d2b702","width":"6","height":"1","order":1,"showTitle":true,"className":"","visible":"true","disabled":"false"},{"id":"c694d0ebe0d2b702","type":"ui-page","name":"Page1","ui":"1805777f90e92057","path":"/page1","icon":"home","layout":"grid","theme":"a965ccfef139317a","order":2,"className":"","visible":"true","disabled":"false"},{"id":"1805777f90e92057","type":"ui-base","name":"dashboard","path":"/dashboard","includeClientData":true,"acceptsClientConfig":["ui-notification","ui-control"],"showPathInSidebar":false},{"id":"a965ccfef139317a","type":"ui-theme","name":"HN Theme","colors":{"surface":"#5c5c5c","primary":"#00fdff","bgPage":"#383838","groupBg":"#4f4f4f","groupOutline":"#858585"},"sizes":{"pagePadding":"12px","groupGap":"12px","groupBorderRadius":"4px","widgetGap":"12px"}}]
```
This flow shows issue.

### Environment

- Dashboard version: 1.12.4
- Node-RED version: 3..1.3
- Node.js version: 20
-  npm version:  9.6.6
- Platform/OS: Android
- Browser: chrome

arturv2000:
## Description

Several changes to `UI-Form`:

- Allow `ui-form` to accept the control message `msg.enabled`
- Prevent widget to forward a input message to output
- Validate fields (to enable `submit` button) when pre-loading the form values via `msg.payload`, #1069  
- Change trigger for validation from `blur` to `input` to prevent double click for validation to occur, #935 
- Send all form properties in `msg.payload` message, if after a `clear` the user did not interact with some field it would not be sent
- Make sure that fields `switch` and `checkbox` are sent as values `true` or `false`
- If field is type `number` send value as Number or null if no number present (same behaviour as DB1), #342 

## Related Issue(s)

Fix #1069 
Fix #935 
Fix #342 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/widgets/ui_form.html
+++ b/nodes/widgets/ui_form.html
@@ -51,7 +51,8 @@
                 topic: { value: 'topic', validate: (hasProperty(RED.validators, 'typedInput') ? RED.validators.typedInput('topicType') : function (v) { return true }) },
                 topicType: { value: 'msg' },
                 splitLayout: { value: '' },
-                className: { value: '' }
+                className: { value: '' },
+                passthru: { value: false }
             },
             inputs: 1,
             outputs: 1,
@@ -60,6 +61,9 @@
             label: function () { return this.name || this.label || 'form' },
             labelStyle: function () { return this.name ? 'node_label_italic' : '' },
             oneditprepare: function () {
+                // Set passthru to `false` to prevent message pass through in Form widget
+                this.passthru = false
+
                 if ($('#node-input-submit').val() === null) { $('#node-input-submit').val('submit') }
                 if ($('#node-input-cancel').val() === null) { $('#node-input-cancel').val('cancel') }
                 // if this groups parent is a subflow template, set the node-config-input-width and node-config-input-height up
--- a/nodes/widgets/ui_form.html
+++ b/nodes/widgets/ui_form.html
@@ -61,9 +61,6 @@
             label: function () { return this.name || this.label || 'form' },
             labelStyle: function () { return this.name ? 'node_label_italic' : '' },
             oneditprepare: function () {
-                // Set passthru to `false` to prevent message pass through in Form widget
-                this.passthru = false
-
                 if ($('#node-input-submit').val() === null) { $('#node-input-submit').val('submit') }
                 if ($('#node-input-cancel').val() === null) { $('#node-input-cancel').val('cancel') }
                 // if this groups parent is a subflow template, set the node-config-input-width and node-config-input-height up
--- a/nodes/widgets/ui_switch.js
+++ b/nodes/widgets/ui_switch.js
@@ -4,6 +4,7 @@ const { appendTopic } = require('../utils/index.js')
 module.exports = function (RED) {
     function SwitchNode (config) {
         // create node in Node-RED
+        config.passthru = false
         RED.nodes.createNode(this, config)
 
         const node = this
--- a/nodes/widgets/ui_switch.js
+++ b/nodes/widgets/ui_switch.js
@@ -4,7 +4,6 @@ const { appendTopic } = require('../utils/index.js')
 module.exports = function (RED) {
     function SwitchNode (config) {
         // create node in Node-RED
-        config.passthru = false
         RED.nodes.createNode(this, config)
 
         const node = this
--- a/ui/src/widgets/ui-form/UIForm.vue
+++ b/ui/src/widgets/ui-form/UIForm.vue
@@ -74,11 +74,11 @@ export default {
             const option = this.options
             option.forEach(opt => {
                 if (opt.type === 'checkbox' || opt.type === 'switch') {
-                    if (this.input[opt.key] === undefined || this.input[opt.key] === null) {
+                    if (typeof (this.input[opt.key]) === 'undefined' || this.input[opt.key] === null) {
                         this.input[opt.key] = false
                     }
                 } else if (opt.type === 'number') {
-                    if (this.input[opt.key] === undefined || this.input[opt.key] === null) {
+                    if (typeof (this.input[opt.key]) === 'undefined' || this.input[opt.key] === null) {
                         this.input[opt.key] = null
                     } else {
                         if (isNaN(this.input[opt.key])) {
@@ -88,7 +88,7 @@ export default {
                         }
                     }
                 } else {
-                    if (this.input[opt.key] === undefined || this.input[opt.key] === null) {
+                    if (typeof (this.input[opt.key]) === 'undefined' || this.input[opt.key] === null) {
                         this.input[opt.key] = ''
                     }
                 }
--- a/ui/src/widgets/ui-form/UIForm.vue
+++ b/ui/src/widgets/ui-form/UIForm.vue
@@ -127,8 +127,7 @@ export default {
                 for (const key in payload) {
                     this.input[key] = payload[key]
                 }
-                // May not be the best way, but if we call this directly will give error on the first `input`, it seems that vue has not yet updated
-                setTimeout(() => { this.validate() }, 0)
+                this.$nextTick(() => { this.validate() })
             }
         },
         onDynamicProperties (msg) {
--- a/ui/src/widgets/ui-form/UIForm.vue
+++ b/ui/src/widgets/ui-form/UIForm.vue
@@ -83,6 +83,9 @@ export default {
         reset () {
             this.$refs.form.reset()
         },
+        validate () {
+            this.$refs.form.validate()
+        },
         rules (row) {
             if (row.required) {
                 // is required
@@ -100,6 +103,8 @@ export default {
                 for (const key in payload) {
                     this.input[key] = payload[key]
                 }
+                // May not be the best way, but if we call this directly will give error on the first `input`, it seems that vue has not yet updated
+                setTimeout(() => { this.validate() }, 0)
             }
         },
         onDynamicProperties (msg) {
--- a/ui/src/widgets/ui-form/UIForm.vue
+++ b/ui/src/widgets/ui-form/UIForm.vue
@@ -1,6 +1,6 @@
 <template>
     <label v-if="label" class="nrdb-ui-form-label">{{ label }}</label>
-    <v-form ref="form" v-model="isValid" validate-on="blur" @submit.prevent="onSubmit">
+    <v-form ref="form" v-model="isValid" :disabled="!state.enabled" validate-on="blur" @submit.prevent="onSubmit">
         <div class="nrdb-ui-form-rows" :class="{'nrdb-ui-form-rows--split': props.splitLayout}">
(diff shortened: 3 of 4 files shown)


ui/src/widgets/ui-form/UIForm.vue
```diff
@@ -96,6 +127,8 @@ export default {
                 for (const key in payload) {
                     this.input[key] = payload[key]
                 }
+                // May not be the best way, but if we call this directly will give error on the first `input`, it seems that vue has not yet updated
```
joepavitt, line 130: Vue has a method for this `this.$nextTick(() => { ... })` which should work here

nodes/widgets/ui_form.html
```diff
@@ -60,6 +61,9 @@
             label: function () { return this.name || this.label || 'form' },
             labelStyle: function () { return this.name ? 'node_label_italic' : '' },
             oneditprepare: function () {
+                // Set passthru to `false` to prevent message pass through in Form widget
+                this.passthru = false
```
joepavitt, line 65: Not sure this is required here if we're not presenting the user with an object, it would also require users to have opened the palette for a form before this works. 

Instead, this could be placed into `ui_form.js` and be hard-coded for all forms by doing `config.passthru = false`

ui/src/widgets/ui-form/UIForm.vue
```diff
@@ -66,6 +70,30 @@ export default {
     },
     methods: {
         onSubmit: function () {
+            // Prevent sending null for switch and combobox, if type number send as Number or null if nothing present on text field and if other fields not present, send empty string
+            const option = this.options
+            option.forEach(opt => {
+                if (opt.type === 'checkbox' || opt.type === 'switch') {
+                    if (this.input[opt.key] === undefined || this.input[opt.key] === null) {
```
joepavitt, line 77:
```suggestion
                    if (typeof(this.input[opt.key]) === 'undefined' || this.input[opt.key] === null) {
```

I get told off by @Steve-Mcl every time I do this, so just to make sure we're consistent across the app, checking `type` is a stronger check than checking the value.

ui/src/widgets/ui-form/UIForm.vue
```diff
@@ -66,6 +70,30 @@ export default {
...
+                        this.input[opt.key] = false
+                    }
+                } else if (opt.type === 'number') {
+                    if (this.input[opt.key] === undefined || this.input[opt.key] === null) {
```
joepavitt, line 81:
```suggestion
                    if (typeof(this.input[opt.key]) === 'undefined' || this.input[opt.key] === null) {
```

ui/src/widgets/ui-form/UIForm.vue
```diff
@@ -66,6 +70,30 @@ export default {
...
+                        this.input[opt.key] = null
+                    } else {
+                        if (isNaN(this.input[opt.key])) {
+                            this.input[opt.key] = null
+                        } else {
+                            this.input[opt.key] = Number(this.input[opt.key])
+                        }
+                    }
+                } else {
+                    if (this.input[opt.key] === undefined || this.input[opt.key] === null) {
```
joepavitt, line 91:
```suggestion
                    if (typeof(this.input[opt.key]) === 'undefined' || this.input[opt.key] === null) {
```

joepavitt (changes requested):
This a great PR - thanks @arturv2000 

Few suggestions and bits of feedback to address

nodes/widgets/ui_form.html
arturv2000: Ok, indeed if the user never opens the config panel, the variable will always be `true` based on previous configuration.

Will change to `ui_forms.js`

ui/src/widgets/ui-form/UIForm.vue
arturv2000: Tanks, i knew it should exist a "better" way, but google was not my friend that day...

ui/src/widgets/ui-form/UIForm.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <label v-if="label" class="nrdb-ui-form-label">{{ label }}</label>
-    <v-form ref="form" v-model="isValid" validate-on="blur" @submit.prevent="onSubmit">
+    <v-form ref="form" v-model="isValid" :disabled="!state.enabled" validate-on="input" @submit.prevent="onSubmit">
```
joepavitt, line 3: validation on input can be quite annoying, as in the case of an `e-mail` it validates as "error" the first character you put in, until you type an `@`

Any particular reason you changed this?

arturv2000: Was a way to fix #935, but didn't try with email type, my test form only had `number` and `text`. Will check if there is a better option

joepavitt: a `validate()` in a `nextTick()` when a new message is received should do the trick there

arturv2000: In this case there is no message, the user is filling the `form` and if it goes directly from a text input to the validate button, and it is not `enabled`, two clicks are required, the first click trigger the `onBlur` event and enables the submit button for the second click.

arturv2000: @joepavitt just tried this, and even with the UI-Form set to `email` the validation seems OK, in reality it is not even validating that is a valid email (with `@`) as long as write any character the validation is "ok", it is just checking that the field cannot be empty.

joepavitt:
Just my own checklist for testing functionality:

- [x] Allow ui-form to accept the control message msg.enabled
- [ ] Prevent widget to forward a input message to output
- [ ] Validate fields (to enable submit button) when pre-loading the form values via msg.payload, https://github.com/FlowFuse/node-red-dashboard/issues/1069
- [ ] Change trigger for validation from blur to input to prevent double click for validation to occur, https://github.com/FlowFuse/node-red-dashboard/issues/935
- [ ] Send all form properties in msg.payload message, if after a clear the user did not interact with some field it would not be sent
- [ ] Make sure that fields switch and checkbox are sent as values true or false
- [ ] If field is type number send value as Number or null if no number present (same behaviour as DB1), https://github.com/FlowFuse/node-red-dashboard/issues/342

nodes/widgets/ui_switch.js
```diff
@@ -4,6 +4,7 @@ const { appendTopic } = require('../utils/index.js')
 module.exports = function (RED) {
     function SwitchNode (config) {
         // create node in Node-RED
+        config.passthru = false
```
joepavitt, line 7: This is in the `ui_switch.js` - think this should be in `ui_form.js`, no?

arturv2000: Yes, it is in the wrong place, sorry...

colinl:
@arturv2000 did you see #343?

arturv2000:
@colinl Sorry, i missed it.

When you say

> Edit the form and change the label/name of one of the fields and exercise it.

Is this, editing the `ui-form` in the config windows on node-red editor, changing fields name/label and deploying again?

I can give it a look, worst case can be a PR after this one.

colinl:
> Is this, editing the `ui-form` in the config windows on node-red editor, changing fields name/label and deploying again?

Yes.  I don't think it is a particularly critical bug, I just noticed it as I was configuring things.

arturv2000:
Just tried that, and don't see anything strange even in `main` branch, if possible can you confirm that it still happens? It may have been solved by some other PR not directly related...

arturv2000:
Ok, can see the issue if instead of doing "Full" deploy, only made a "modified flows" or "modified nodes".

May attempt tomorrow to take a look, but it seems that is something more internal to node-red / dashboard workings that it becomes a little outside my capabilities

joepavitt:
There is plenty of good stuff in this PR already, if you do attempt #343, please do it as a new PR, so that it doesn't block this PR being merged

arturv2000:
I have found what is happening, and it is fixable on `vue` side, not necessarily related to node-red.

The same thing happens if we load a new form elements via `msg.ui_update.options` the old key values will be present until a page refresh.

I can create a new PR after this one is merged.

joepavitt (approved):

PR #1114, opened 2024-07-20
UI-Slider: Add missing properties to msg.`ui_update`

arturv2000:
## Description

- Property `step` was not being used from dynamic properties in the vue template definition (was `props.step`)
- Properties `color`, `colorTrack` and `colorThumb` where not being update in `onDynamicProperties` function
- Properties `colorTrack` and `colorThumb` where not being correctly stored on `statestore`
- Fix / Add Properties to docs and help page

## Checklist

 - [X ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-slider/UISlider.vue
+++ b/ui/src/widgets/ui-slider/UISlider.vue
@@ -9,7 +9,7 @@
         :min="min" :direction="direction"
         :tick-size="4" :track-size="4"
         :color="color" :track-color="colorTrack" :thumb-color="colorThumb"
-        :max="max" :step="props.step || 1" :show-ticks="showTicks"
+        :max="max" :step="step || 1" :show-ticks="showTicks"
         @update:model-value="onChange" @end="onBlur"
     />
 </template>
--- a/ui/src/widgets/ui-slider/UISlider.vue
+++ b/ui/src/widgets/ui-slider/UISlider.vue
@@ -168,6 +168,15 @@ export default {
             if (typeof updates.iconPrepend !== 'undefined') {
                 this.dynamic.iconPrepend = updates.iconPrepend
             }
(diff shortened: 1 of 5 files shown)


ui/src/widgets/ui-slider/UISlider.vue
```diff
@@ -168,6 +168,15 @@ export default {
             if (typeof updates.iconPrepend !== 'undefined') {
                 this.dynamic.iconPrepend = updates.iconPrepend
             }
+            if (typeof updates.color !== 'undefined') {
```
gayanSandamal, line 171: @arturv2000 if we're considering truthy values including "null" as well as "undefined" if we can simply use as below

`if (updates.color) {`

instead of

`if (typeof updates.color !== 'undefined') {`

Anyway it's based on the approach

gayanSandamal (approved):
LGTM

ui/src/widgets/ui-slider/UISlider.vue
arturv2000: Thanks for the feedback,

Considering the use context of `ui_update` values i would say a better approach would be:

```js
if('color' in updates){
```

We may want to (let's say disable the property to attempt to return to default value). At least `null` is a valid value, is the way to return the variable/property to the default widget configuration value.

Personally, i don't like to use
```js
if (updates.color)
```

Since sometimes, that habit makes we use it in situations where a `false`, `null` or empty string may be valid values.

gayanSandamal: Thanks for the clarification

Issue #1117, opened 2024-07-21
Dropdown widget loses it’s entries when switching tabs


natcl:
### Current Behavior

I’m setting the options of a dropdown widget at startup using msg.options which works fine. If I switch to another tab and switch back the entries are gone, they don’t seem to persist.

### Expected Behavior

I would expect the options to persist between tab changes.

### Steps To Reproduce

Inject options via msg.options at startup on a widget on tab A. Switch to tab B and come back to tab A, options are now gone.

### Environment

- Dashboard version: 1.13.0
- Node-RED version: 4.0.2
- Node.js version: 20.15.1
- npm version:
- Platform/OS: Alpine 3.20
- Browser: Safari

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

natcl:
Slack conversation here:
https://node-red.slack.com/archives/C1DH77X1V/p1721421443800439

joepavitt:
Can reproduce. Oddly, the dynamic data is correctly stored server-side, and persists - you can check this by looking at our behind-the-scenes debug view at `your-instance:1880/dashboard/_debug`

So this suggests that the problem is that the `widget-load` event isn't correctly firing on a page-switch such that we load the latest state with updated options. Even as a backup though, we should have this data stored in the VueX store on the client-side, so it _should_ have two backups... but neither of which are functioning correctly.

@gayanSandamal can you investigate the details when you have a chance please?

gayanSandamal:
@joepavitt I'll investigate this further

arturv2000:
Just my 2cents, this is an issue similar to what happens in other widgets, the `onLoad` event may not handle / update the `msg.ui_update` properties, and the event `onDynamicProperties` not being "triggered" independently when the widget is loaded.

a possible fix is:
```js
onLoad (msg) {
    // update vuex store to reflect server-state
    this.$store.commit('data/bind', {
        widgetId: this.id,
        msg
    })
    this.onDynamicProperties(msg)  // Get the `ui_update` properties that may exist for this widget at load time, until is properly handled in `data_tracker.mjs`, needs to be called before `this.select`
    this.select(this.messages[this.id]?.payload)
},
```

[...]

Issue #1022, opened 2024-06-24
ui-dropdown - Not parsing `options` when widget is reload based on message `widget-load`


arturv2000:
### Current Behavior

The Widget `ui-dropdown` no longer accepts/parse the `options` part of the message that receives when it is loaded `widget-load`

Was working at least in V1.8.1.

It seems that the function:
https://github.com/FlowFuse/node-red-dashboard/blob/61e3d24cbed9ab588709dcfb873f0a2f69d5876d/ui/src/widgets/ui-dropdown/UIDropdown.vue#L92

Should not only call the `select` , but it seems to be more practical to call `onDynamicProperties` to reload the proprieties/options/payload for this widget.

### Expected Behavior

Options should load on widget load based on the last message.

### Steps To Reproduce

Small example:
```
[...]

Issue #793, opened 2024-04-21
'Widget load' event triggered only when there is data for the node in the Node-red datastore


omrid01:
### Current Behavior

Today, when a dashboard 2.0 widget stores its state in the Node-red datastore, when the widget loads (e.g. upon client open or refresh) it will receive a 'widget-load' notification, with a `msg` object holding the stored data. However, this message is triggered only if the datastore has any data of this node.

### Expected Behavior

We need the event to be triggered unconditionally, with an empty `msg `in case the datastore had no data for the node. The rationale is that if there is no stored state for this widget, we may want to initialize it with some defaults, but while we get an event when there is data, we can never tell that there is **no data** available. 

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Logic currently prevents the `widget-load` event firing as it's possible for plugins to _clear_ that object and consequently prevent a message being sent out at all (e.g. if an auth plugin says it's not permitted).

The difficulty here is how we differentiate the plugin removing a `msg`, and there being no `msg` in the first place. My thinking is, I need to change the API/pattern for the hooks, and get them to throw errors, and we catch them, to differentiate from a genuine empty `msg` being sent.

Issue #833, opened 2024-05-07
Modify/Update UI-widget parameters from external


xX-Nexus-Xx:
### Description

Hi,

in DB1 it was possible to modify/update UI-widget parameters via input (e.g msg.Range.min,msg.Range.max for UI-Slider range).

Would it be possible to have the same flexibility with the new UI-widgets in DB2?

Below UI-Slider is just an example (and the one I need most) but the feature request wuld apply to all the UI-widgets with its parameters

![image](https://github.com/FlowFuse/node-red-dashboard/assets/2195190/d229a5c9-8b3d-434d-9136-46a3ef679f14)

thanks for considering

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

bartbutenaers:
Hi @joepavitt,
If you want I can have a look at this one.  
Unless you need it fast, because that won't fit into my limited free time...
1. Are there any best practices I need to take into account?
2. I assume the dropdown node is again the best reference example for this?
3. And I assume this is not limited to the slider node?  If so, it might be better to create a PR per ui node?

joepavitt:
> And I assume this is not limited to the slider node? If so, it might be better to create a PR per ui node?

We have a considerable backlog of dynamic properties that are needed - I may try and dedicate a day this week you _just_ this.

I should be able to cover it, thanks for the offer @bartbutenaers

joepavitt:
Had already started work on dynamic props, with several PRs open following _Pattern 1_ from the below list. However, @Steve-Mcl has raised very valid concerns on this pattern, and he I are are currently talking through the options we have on architecture here:

### Pattern 1 - `<msg.property>`

- Any dynamic properties can be overridden with `msg.<property>`
- This occurs even if a value was defined in the node configuration
    - Worth noting this is against the NR standard, which will _only_ override, if a field is empty
 
#### Pros:

- Easy to follow and understand
- Easy for the user to define
- Easy to implement from a development standpoint

#### Cons:

- `msg` objects passed through nodes could then have a knock on effect, e.g. `msg.label` could end up running through many `ui-` nodes and override their label property
- `msg.<property>` is ambiguous in that it could easily overlap _accidentally_ with other, valid, uses of that same property
- Invisible "magic" property, not in the configuration explicitly

### Pattern 2 -  `<msg.ui.property>`

[...]

PR #1123, opened 2024-07-24
Dynamic Options - Implement new helper functions & update widget-load behaviour

### LINKED ISSUE #1122: Re-architect widget-load and data transfer workflow
### Description

Capturing this in a new issue after lengthy discussions and issues captured here:

- https://discourse.nodered.org/t/how-to-sync-the-full-config-between-server-and-client-side-state/89252/80
- https://github.com/FlowFuse/node-red-dashboard/issues/1117

#### Widget Load Loses State

The problem can be seen in the Events Architecture [here](https://dashboard.flowfuse.com/contributing/guides/events.html#loading-event-flow):

Screenshot 2024-07-24 at 14 22 57

Note that in the `widget-load` event, there is no reference to the `state` store, so when a user navigates away from a page, updates to that store can occur, but the `widget-load` events is not aware of them.

Loosely thinking, we need to have something like:

Screenshot 2024-07-24 at 14 23 39

#### Data Store Changes

There is also merit here in swapping how the `datastore` is used - rather than being a full `msg` store - we could treat it like `state` in that it's a singular, cumulative reference to the `data` for a single widget. For example, if `msg.payload` and `msg.otherProperty` are sent in _different_ messages, they would still be stored in the same place, and would be replayed accordingly.

joepavitt:
## Description

Okay, this is a fairly big shift, so will do my best to articulate the changes, and the _why_.

### `widget-load` changes

- We now include a `state` value in the `widget-load` event. This is because any changes, and dynamic properties, set on a node when not in view on the Dashboard would be updated server-side, but not client-side. So, when a user navigates (via the side navigation) to the relevant page, the state loaded is still the original configuration provided by Node-RED's latest deploy.
- In the client-side handler for this event, we now automatically put everything received in `state` straight into our VueX store under the `widget.state` field. This means that we don't need to define this on _every_ widget, which we would need to. Note though, that we still need to have the server-side definitions over what can go into `state`, and the client-side widgets need to reference this still via `computed` variables (explained next)
- This event will now fire for _every_ widget, as long as it has `msg` _or_ `state` saved in the server-side stores. Previously, it would only fire if something was present in `datastore` (i.e. last received `msg`)

### Move to a centralised store model

- I've removed the use of the `dynamic` design pattern that we had implemented in order to track the dynamic overrides. 
- Instead, we're referencing the `widget` state stored in vuex directly using `computed` variables. This ensures we have a single source of truth, and not many copies of the truth distributed out to each widget/node.

### New global helper functions

- In order to make the interface to the vuex store easier, I've created two new global helper functions for dealing with dynamic (and static) properties:
    - `updateDynamicProperty(property, value)`: Updates a property with the provided value, and checks against undefined types
    - `setDynamicProperties(config)`: Will set the provided properties in the vuex store for a given widget this is run on. This will automatically update the widget's state, and any references using this property.
    - `getProperty(property)`: Automatically gets the correct value for the requested property. Will first look in the dynamic properties, and if not found, will default to the static configuration (i.e. what's defined in Node-RED) that arrived via the `ui-config` event.

### `widgetState` store mutation changes

- The arguments for the `widgetState` mutation on the `ui` vuex store has changed to permit separation of the proeprties being set and the `id` of the widget.

## Progress

We need to sweep over all of the existing widgets that were already using the design pattern to ensure consistency:

- [x] UI Form
- [x] UI Button
- [x] UI Button Group
- [x] UI Dropdown
- [x] UI Slider

We can then continue https://github.com/FlowFuse/node-red-dashboard/issues/833 with the new pattern introduced here

## Related Issue(s)

Closes https://github.com/FlowFuse/node-red-dashboard/issues/793
Closes https://github.com/FlowFuse/node-red-dashboard/issues/1022
Closes https://github.com/FlowFuse/node-red-dashboard/issues/1117
Closes #1122

Need to double check once closed on whether it helps https://github.com/FlowFuse/node-red-dashboard/issues/871


joepavitt:
Only ui-dropdown to go, which is a little trickier as has some backward compatibility stuff left over with `msg.options` that we still need to support - will attack that tomorrow

ui/src/widgets/ui-button/UIButton.vue
```diff
@@ -93,26 +81,23 @@ export default {
...
+                this.setDynamicProperties({ buttonColor: updates.buttonColor })
             }
             if (typeof updates.textColor !== 'undefined') {
-                this.dynamic.textColor = updates.textColor
+                this.setDynamicProperties({ textColor: updates.textColor })
             }
             if (typeof updates.iconColor !== 'undefined') {
-                this.dynamic.iconColor = updates.iconColor
+                this.setDynamicProperties({ iconColor: updates.iconColor })
             }
```
Steve-Mcl, line 100: Joe, I think we can both cement and simplify this behaviour with an additional helper. 

### Additional helper, with guards and the `undefined` pattern built in for 100% consistency

```
                updateDynamicProperty(prop, config) {
                    if (!prop || !config || typeof config !== 'object') {
                       return // maybe better to throw error here as this is likely program error
                    }
                    if (typeof config[prop] === 'undefined') {
                       return // no change
                    }
                    this.$store.commit('ui/widgetState', {
                        widgetId: this.id,
                        config
                    })
                },
```

### Now, all calls, by all nodes (contrib too) become consistent
```
          this.updateDynamicProperty('label', updates)
          this.updateDynamicProperty('icon', updates)
          this.updateDynamicProperty('iconPosition', updates)
          this.updateDynamicProperty('buttonColor', updates)
          this.updateDynamicProperty('textColor', updates)
          this.updateDynamicProperty('iconColor', updates)
```


---

If necessary (i.e there is a reason to permit `undefined` as a value) we can add an options object to the `updateDynamicProperty` e.g.  `this.updateDynamicProperty('label', updates, { allowUndefined: true })`

I realise this is a little extra work but as this is a large change already, we have the opportunity to cement good practice at the core.

Happy to discuss over a call.

colinl:
How are validations being done on ui_update operations, to make sure invalid data are not allowed?

joepavitt:
> How are validations being done on ui_update operations, to make sure invalid data are not allowed?

Needs to be done on a node-by-node basis. Right now, not much validation.

Also PR opened for UI Example updates: https://github.com/FlowFuse/node-red-dashboard-2-ui-example/pull/17 will get it reviewed once these core PRs are merged

joepavitt:
> One question regarding the use of the dynamic properties, is it sensible that the user may want to return a specific property to its default value, (for example button color, that was defined in the config window as yellow, or the label text) by sending a null or an undefined?

If the value is set to `null`, then the property is reset. Our `getProperty` logic checks:

```js
return this.state && property in this.state && state !== null ? state : config
```

Note, if `state === null`, then the original `config` will be returned

arturv2000:
I was testing that scenario, and it doesn't work.

For example:
- I have a button that I defined in the config as the color is `yellow` and perform a full-deploy.
- I send a message setting the color to `red` using `ui_update.buttonColor`
- In this case, if I send `ui_update.buttonColor = null` the button color will return to `yellow`
- If I refresh the page, the button color return to the theme default, if send again the `ui_update.buttonColor = null` it has no effect. If before I refresh the page i set the color to `green` after the refresh sending a `null` will return the color to `green`.

Not a big issue from my part, but may be something to consider if it is necessary to have a way to revert to initial configuration.

joepavitt:
Will follow up with outstanding nodes in https://github.com/FlowFuse/node-red-dashboard/issues/833 once this is merged

Steve-Mcl:
### ui-dropdown

SQLiteStudio_mYtMWIPAd0


1. Setting label dynamically works but reverting by sending null doesnt appear to work as exepcted
2. Dynamically setting multiple `true`/`false` causes odd entries in the current selection



chrome_Yirme9hgqh

3. After applying dynamic label a refresh forgets selection

joepavitt:
UI Dropdown:

1. We don't support that currently, will be a follow up exercise.

2. If you're referring to the vertical alignment problem, that's fixed in a PR that Gayan opened yesterday, it was an underlying Vuetify bug that's fixed in newest version.

3. I'd expect selection to be forgotten anyway on refresh? Unless the last message received contains the `payload` it's a shortfall of the "replay last message" approach we've copied from Dashboard 1.0

Steve-Mcl:
> 2\. If you're referring to the vertical alignment problem, that's fixed in a PR that Gayan opened yesterday, it was an underlying Vuetify bug that's fixed in newest version

No, i am referring to what happens in the GIF ("two" is selected, I inject `msg.ui_update.multiple = true`, "two two" is now shown!)
image
I posted the demo flow too.


> 1. We don't support that currently, will be a follow up exercise.

Don't support dynamic label for ui-dropdown? 
As in not documented right? 
_(because is does work and there is code for it in this PR - so I tested it)_




> 3\. I'd expect selection to be forgotten anyway on refresh? Unless the last message received contains the `payload` it's a shortfall of the "replay last message" approach we've copied from Dashboard 1.0

Is this something we can fix/improve?

joepavitt:
> Don't support dynamic label for ui-dropdown?
As in not documented right?

The use of `null` to clear it, which is what I thought is what you're saying was not working?

joepavitt:
Now reliving what you mean re: multiple. That's likely an issue in current `main`, I'm AFK but if you're able to check that'd be great.

Don't think that's a problem been introduced here

joepavitt:
> Is the problem here that a message containing only ui_update settings is stored as the last message? I think that such messages should not be stored, as the ui_update data are stored in the state store.

This could work, but not in scope of this issue

joepavitt:
> I thought the point of making helper functions was to cement this common behaviour at the core

It's not listed as a supported or new feature, so no, this has lot been added, but can be in a follow up PR/issue

Steve-Mcl:
> > I thought the point of making helper functions was to cement this common behaviour at the core
> 
> It's not listed as a supported or new feature, so no, this has lot been added, but can be in a follow up PR/issue

Are any of these changes breaking or will users experience differing behaviour when this is merged Joe?  If yes, then I would personally recommend we get the behaviour nailed down in this PR before we release it into the wild.

Steve-Mcl:
#### ui-number-group

chrome_qTk75XCTdo


1. Sending payload with matching value does not slect the matching button. After a refresh it does become selected.
2. Sending an updated label correctly sets the label.   Sending `null` correctly reverts the label to static value. However, if the DB is refreshed, sending `null` no longer reverts label to static value.
3. Sending a new set of options with same/matching `value`s does not keep the current selection.

demo flow

```
[{"id":"8c7ef62c4af9d424","type":"inject","z":"f80624abfd69b02e","name":"ui_update.label Numbers","props":[{"p":"ui_update.label","v":"Numbers","vt":"str"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":1250,"y":80,"wires":[["75551956e485bac6"]]},{"id":"285f16e64f307827","type":"inject","z":"f80624abfd69b02e","name":"ui_update.label null","props":[{"p":"ui_update.label","v":"null","vt":"json"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":1270,"y":120,"wires":[["75551956e485bac6"]]},{"id":"48af77c3c271917d","type":"inject","z":"f80624abfd69b02e","name":"ui_update.options numbers","props":[{"p":"ui_update.options","v":"[{\"value\":1,\"label\":\"One\",\"icon\":\"numeric-positive-1\"},{\"value\":true,\"label\":\"Two\",\"icon\":\"numeric-2-box\"},{\"value\":\"tree\",\"label\":\"T_ree\",\"icon\":\"tree\"}]","vt":"json"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":1250,"y":200,"wires":[["75551956e485bac6"]]},{"id":"7a50a8231976b836","type":"inject","z":"f80624abfd69b02e","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"1","payloadType":"num","x":1310,"y":240,"wires":[["75551956e485bac6"]]},{"id":"17dc21e121c2184c","type":"inject","z":"f80624abfd69b02e","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"tree","payloadType":"str","x":1310,"y":320,"wires":[["75551956e485bac6"]]},{"id":"cb13084236dfcfc8","type":"inject","z":"f80624abfd69b02e","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"true","payloadType":"bool","x":1310,"y":280,"wires":[["75551956e485bac6"]]},{"id":"80798abd6bac3521","type":"debug","z":"f80624abfd69b02e","name":"debug 187","active":true,"tosidebar":true,"console":false,"tostatus":true,"complete":"payload","targetType":"msg","statusVal":"payload","statusType":"auto","x":1540,"y":200,"wires":[]},{"id":"8021da3ecf8e21f9","type":"inject","z":"f80624abfd69b02e","name":"ui_update.options null","props":[{"p":"ui_update.options","v":"null","vt":"json"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":1260,"y":360,"wires":[["75551956e485bac6"]]},{"id":"2c1654b9f7489958","type":"ui-button-group","z":"f80624abfd69b02e","name":"","group":"5c9407416fba1d1b","order":1,"width":6,"height":1,"label":"stars","rounded":false,"useThemeColors":true,"passthru":false,"options":[{"label":"Cap'n A","icon":"star-circle-outline","value":"1","valueType":"num","color":"#009933"},{"label":"Crescent","icon":"star-crescent","value":"true","valueType":"bool","color":"#999999"},{"label":"David","icon":"star-david","value":"tree","valueType":"str","color":"#ff6666"}],"topic":"topic","topicType":"msg","x":1530,"y":140,"wires":[["80798abd6bac3521"]]},{"id":"75551956e485bac6","type":"junction","z":"f80624abfd69b02e","x":1440,"y":140,"wires":[["2c1654b9f7489958"]]},{"id":"5c9407416fba1d1b","type":"ui-group","name":"PR-1123","page":"868aaffa6da2e1ce","width":"6","height":"1","order":1,"showTitle":true,"className":"","visible":"true","disabled":"false"},{"id":"868aaffa6da2e1ce","type":"ui-page","name":"PR-1123","ui":"22ea43815413e748","path":"/page13","icon":"home","layout":"grid","theme":"c2ff5ba1f92a0f0e","order":1,"className":"","visible":"true","disabled":"false"},{"id":"22ea43815413e748","type":"ui-base","name":"base","path":"/dashboard","includeClientData":true,"acceptsClientConfig":["ui-notification","ui-control"],"showPathInSidebar":false,"navigationStyle":"default","titleBarStyle":"default"},{"id":"c2ff5ba1f92a0f0e","type":"ui-theme","name":"Default","colors":{"surface":"#ffffff","primary":"#0094ce","bgPage":"#eeeeee","groupBg":"#ffffff","groupOutline":"#cccccc"},"sizes":{"pagePadding":"12px","groupGap":"12px","groupBorderRadius":"4px","widgetGap":"12px"}}]
```

joepavitt:
> Are any of these changes breaking or will users experience differing behaviour when this is merged Joe?

There should be 0 difference to UX as a result of this PR

Steve-Mcl:
> > Are any of these changes breaking or will users experience differing behaviour when this is merged Joe?
> 
> There should be 0 difference to UX as a result of this PR

Ok, well in which case my semi - thorough tests should probably be halter until I can establish a baseline to see if this Pr is causing the issues I see or if they are existing!


Before I halt, I will post findings for UI-Button (rather than let my findings go to waste)

#### ui-button

chrome_7lXqVChiYj

1. Setting dynamic label works. sending `null` for label reverts it. However, after a refresh, it can not be reverted anymore.
2. Setting dynamic icon + iconPosition works. sending `null` for icon + iconPosition reverts them. However, after a refresh, they can no longer be reverted.

chrome_o65l3ZoRzM

3.  Setting dynamic label works. sending `null` for label reverts it. However if we refresh with the label reverted via a `null`, the button has no label after a refresh
4. Setting dynamic icon + iconPosition works. sending `null` for icon + iconPosition reverts them. However, if we refresh with the icon + iconPosition revered via a `null`, the button is lost. An error in console reveals the issue is in `makeMdiIcon`
```
TypeError: Cannot read properties of null (reading 'replace')
    at Proxy.makeMdiIcon (index-BPCbQ1sF.js:47:36548)
    at Proxy.prependIcon (index-BPCbQ1sF.js:47:35741)
    at ReactiveEffect.fn (index-BPCbQ1sF.js:10:13076)
    at ReactiveEffect.run (index-BPCbQ1sF.js:10:1918)
    at get value (index-BPCbQ1sF.js:10:13345)
    at Object.get [as prependIcon] (index-BPCbQ1sF.js:17:34086)
    at Proxy._sfc_render$h (index-BPCbQ1sF.js:47:37151)
    at renderComponentRoot (index-BPCbQ1sF.js:17:7654)
    at ReactiveEffect.Ce [as fn] (index-BPCbQ1sF.js:17:54912)
    at ReactiveEffect.run (index-BPCbQ1sF.js:10:1918)
```
image


demo flow
```
[{"id":"a19fe7752981edcb","type":"inject","z":"f80624abfd69b02e","name":"icon & position left","props":[{"p":"ui_update.icon","v":"menu-left","vt":"str"},{"p":"ui_update.iconPosition","v":"left","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":1170,"y":80,"wires":[["225e4e1efcbc0ddc"]]},{"id":"019c68b113848730","type":"inject","z":"f80624abfd69b02e","name":"icon & position null","props":[{"p":"ui_update.icon","v":"null","vt":"json"},{"p":"ui_update.iconPosition","v":"null","vt":"json"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":1170,"y":120,"wires":[["225e4e1efcbc0ddc"]]},{"id":"303170327edd25aa","type":"inject","z":"f80624abfd69b02e","name":"ui_update.label BTN TWO","props":[{"p":"ui_update.label","v":"BTN TWO","vt":"str"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":1150,"y":240,"wires":[["1e59de55be3906d2"]]},{"id":"835d0f44edce12c9","type":"inject","z":"f80624abfd69b02e","name":"ui_update.label null","props":[{"p":"ui_update.label","v":"null","vt":"json"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":1170,"y":280,"wires":[["1e59de55be3906d2"]]},{"id":"97fc0235988ce987","type":"debug","z":"f80624abfd69b02e","name":"debug 188","active":true,"tosidebar":true,"console":false,"tostatus":true,"complete":"payload","targetType":"msg","statusVal":"payload","statusType":"auto","x":1440,"y":180,"wires":[]},{"id":"225e4e1efcbc0ddc","type":"ui-button","z":"f80624abfd69b02e","group":"5c9407416fba1d1b","name":"Button 1","label":"Button 1","order":1,"width":"3","height":"1","emulateClick":false,"tooltip":"","color":"","bgcolor":"","className":"","icon":"menu-right","iconPosition":"right","payload":"1","payloadType":"num","topic":"topic","topicType":"msg","buttonColor":"","textColor":"","iconColor":"","x":1380,"y":100,"wires":[["97fc0235988ce987"]]},{"id":"1e59de55be3906d2","type":"ui-button","z":"f80624abfd69b02e","group":"5c9407416fba1d1b","name":"Button 2","label":"Button 2","order":2,"width":"3","height":"1","emulateClick":false,"tooltip":"","color":"","bgcolor":"","className":"","icon":"","iconPosition":"right","payload":"two","payloadType":"str","topic":"topic","topicType":"msg","buttonColor":"","textColor":"","iconColor":"","x":1380,"y":260,"wires":[["97fc0235988ce987"]]},{"id":"5c9407416fba1d1b","type":"ui-group","name":"PR-1123","page":"868aaffa6da2e1ce","width":"6","height":"1","order":1,"showTitle":true,"className":"","visible":"true","disabled":"false"},{"id":"868aaffa6da2e1ce","type":"ui-page","name":"PR-1123","ui":"22ea43815413e748","path":"/page13","icon":"home","layout":"grid","theme":"c2ff5ba1f92a0f0e","order":1,"className":"","visible":"true","disabled":"false"},{"id":"22ea43815413e748","type":"ui-base","name":"base","path":"/dashboard","includeClientData":true,"acceptsClientConfig":["ui-notification","ui-control"],"showPathInSidebar":false,"navigationStyle":"default","titleBarStyle":"default"},{"id":"c2ff5ba1f92a0f0e","type":"ui-theme","name":"Default","colors":{"surface":"#ffffff","primary":"#0094ce","bgPage":"#eeeeee","groupBg":"#ffffff","groupOutline":"#cccccc"},"sizes":{"pagePadding":"12px","groupGap":"12px","groupBorderRadius":"4px","widgetGap":"12px"}}]
```


<br>

---

What I will state is, there is a common theme across many of the issues related to override/null/refresh issues

arturv2000:
Most of these issues exists in Main, regarding the `null`/reverting to original configuration issue was already talked about in this PR, and was decided to address that latter on.
https://github.com/FlowFuse/node-red-dashboard/pull/1123#issuecomment-2258110125

arturv2000:
Regarding the `ui-number-group` it is a `bug` from this PR

before was this:
```js
const updates = msg.ui_update
if (typeof updates?.label !== 'undefined') {
}
if (typeof updates?.options !== 'undefined') {
   this.dynamic.options = updates.options
}
```

And now is:
```js
const updates = msg.ui_update
this.updateDynamicProperty('label', updates.label)
this.updateDynamicProperty('options', updates.options)
```

But when we inject a message via Node-Red, the `onDynamicProperties` is called, but the object `ui_update` may not exits, or the property `updates.label` may not exit...


Replacing with:
```js
        onDynamicProperty (msg) {
            const updates = msg.ui_update
            if (!updates) {
                return
            }
            this.updateDynamicProperty('label', updates?.label)
            this.updateDynamicProperty('options', updates?.options)
        },
```
Solves the main issue (Not the null not returning to original configuration)

The same behaviour may be happening in other widgets

joepavitt:
So, I'm fairly sure this is all good to go now. @Steve-Mcl the `null` concerns will be treated as a follow up feature request, and are not a blocker for this PR

Steve-Mcl (approved):
So the newly introduced issues are indeed resolved. 

Code review all good.

NOTE: This is approved in the knowledge that there are a few issues related to mostly to setting then reseting dynamic props but they are existing and are to be addressed separately.  

In case they are not all spelled out in the issue, the ones I seen are:

* dropdown
    * Applying dynamic label then resetting it with a `null` followed by a refresh causes the label to disappear
    * Applying dynamic options then resetting them with a `null` followed by a refresh causes the options to disappear

* form
    * Applying dynamic label then resetting it with a `null` followed by a refresh causes the label to disappear
    * Applying dynamic options then resetting them with a `null` followed by a refresh causes the options to disappear
    * After Applying dynamic options, the first 2 fields appear to be bound (typing in one field will update the other)

* button
    * Applying dynamic icon then resetting it with a `null` followed by a refresh causes the button to disappear
    * Applying dynamic label then resetting it with a `null` followed by a refresh causes the buttons label to disappear

* button group
    * Applying dynamic label then resetting it with a `null` followed by a refresh causes the label to disappear
    * Applying dynamic options then resetting them with a `null` followed by a refresh causes the options to disappear


 NOTE: This ↑ was not exhaustive - only a selection of the dynamic ops and widgets were tested for the purposes of this review.

joepavitt:
Thanks Steve

omrid01:
So just to make sure, there is no backwards compatibility impact:
- server node will continue to send 'on-widget-load' to new clients, with their respective data from the datastore?
- Will on-widget-load now be sent also when the datastore has no data for this node?

joepavitt:
> server node will continue to send 'on-widget-load' to new clients, with their respective data from the datastore?

correct

> Will on-widget-load now be sent also when the datastore has no data for this node?

Yes, correct

omrid01:
Excellent, thank you.
Is there a way to know the dash-2.0 version (programmatically from within the server node?)

joepavitt:
Not that I can think of - any reason you'd need to know for this though, in your `load` handler, just account for `if (!msg)` being a ting, and it should be backward compatabile.

omrid01:
I'm trying to think ahead. Dash-2.0 is continuously evolving with new functionality, and it would be good to enable a node to know the dash-2.0 version it is running on (and determine which framework functionality is available). Equivalent to checking the NodeJS version to determine if a new JavaScript feature is available.

Specifically in my case, it's chicken & egg. I can easily check if a message is null or undefined once I receive the load notification, but will never get this notification in the first place if datastore was empty... :-)  so knowing the version will enable me to retain the current workaround (forcing a dummy datastore entry) for users who upgrade the node but are still on an older dashboard version.

PR #1089, opened 2024-07-13
UI-ButtonGroup: Add `msg.enabled` support, validate `msg.payload` before apply to selection

arturv2000:
## Description

- Add support for the `msg.enabled` property to allow enable/disable widget
- Validate that `msg.payload` exists and it is valid to apply to selection
- Add option to clear selection by sending an empty array

## Related Issue(s)

Fixes #721

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #1146, opened 2024-07-30
Fix: Ensure we check for existence of msg before using msg.payload

joepavitt:
## Description

- Removes `null` pointer in UI Event
- Ensure we check for existence of `msg` onLoad for Button Group
    - Introduced in: https://github.com/FlowFuse/node-red-dashboard/pull/1089
- Ensure we check for existence of `msg` onLoad for UI Text
    - Introduced in: https://github.com/FlowFuse/node-red-dashboard/pull/1091


joepavitt:
Fix for UI Switch, introduced in #1090 will be fixed as part of further work on going in UI Switch

gayanSandamal (approved):
LGTM

PR #1147, opened 2024-07-31
UI Switch - Add "Indicator" mode

### LINKED ISSUE #210: Add option to differentiate input/output state on ui-switch
### Description

https://github.com/flowforge/flowforge-nr-dashboard/issues/141#issuecomment-1725782775

joepavitt:
## Description

- Adds in the "show input value" option that existed in Dashboard 1.0
- If active, then the switch will not auto-toggle when clicked, instead it will output a message, and only respond when a new message is received into the switch.
- Puts switch into "loading" state if decouple and user interacts
    - User can still click over and over, even whilst in "loading" and an event will still be broadcast

## Related Issue(s)

Closes #210


Steve-Mcl (approved):
Looks good to me.

PR #1151, opened 2024-08-04
Show node icons in sidebar

bartbutenaers:
## Description

Hi guys,

I wanted to continue with the migration of my old dashboard, but I now noticed that all widgets in the sidebar have the same icon:

image

I find this rather unclear.  Because it is hard to find fast some specific node and they all look like svg nodes (because I already use the fa-image icon for the svg node).

Via this pull request the icon of the (core or third-party) node is being used:

image

When - for some reason - the icon has no node (or we cannot find it), then the [question](https://fontawesome.com/v4/icon/question) fontawesome node will be displayed.

## Related Issue(s)

I don't see at first sight that an issue has been reported yet.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.html
+++ b/nodes/config/ui_base.html
@@ -1038,7 +1038,7 @@
             addItem: function (container, i, /** @type {DashboardItem} */ widget) {
                 const titleRow = $('<div>', { class: 'nrdb2-sb-list-header nrdb2-sb-widgets-list-header' }).appendTo(container)
                 $('<i class="nrdb2-sb-list-handle nrdb2-sb-widget-list-handle fa fa-bars"></i>').appendTo(titleRow)
-                let widgetIcon = 'fa fa-image'
+                let widgetIcon = 'fa ' + (widget.node._def.icon || 'fa-question').replace('font-awesome/', '')
                 if (widget.isSubflowInstance) {
                     // In this MVP, subflow instances are constrained to stay within own group.
                     container.parent().addClass('red-ui-editableList-item-constrained')


gayanSandamal (approved):
@bartbutenaers Thanks for the PR

@joepavitt this is a valid issue and the fix looks good to me

joepavitt:
We can accept this PR as its still better than what we have, but should open a follow up issue to investigate a better solution to cover _all_ icon types

bartbutenaers:
@joepavitt,

Indeed that is correct if I look at the markdown node:

image

But I was wondering how I could get the path to the image file resource.
So I looked at the ui-markdown node, to see where the flow editor gets the resource:

image

However the png file is nowhere in my flow editor SVG available.  
Instead an SVG icon is being used, which contains a filled path.
Don't understand how it works...

joepavitt:
I'll investigate tomorrow, we may not be loading the resources correctly.

bartbutenaers:
Forget my previous comment.  I think this is the 'default icon', because it cannot find the png resource.  Going to digg further...

bartbutenaers:
Yes indeed, for my old onvif nodes the png is loaded correctly:

image

Which corresponds exactly to what I found now in the Node-RED [documentation](https://nodered.org/docs/creating-nodes/resources).

I will also have a look at the icon of the markdown node.  Sorry for the noise...

joepavitt:
Probably just a missing file or something in our part.

For this PR, we can follow it up with another to include an `<image/>` element instead of the `fa` if the `icon` field _isn't_ an `fa` icon

bartbutenaers:
> Probably just a missing file or something in our part.

I think there has something been mixed-up.
It looks to me that the ui-markdown.png is in fact an image only being used in the documentation:

image

Which shows how a rendered markdown looks like, as you can see in the [ui-markdown.png](https://dashboard.flowfuse.com/images/node-examples/ui-markdown.png) file.

Not sure what you want to do with that (e.g. using a fontawesome icon for the markdown node), so I am indeed not going to touch that now in the follow-up PR...

> For this PR, we can follow it up with another to include an <image/> element instead of the fa if the icon field isn't an fa icon

Ok deal!

bartbutenaers:
I had updated the code to this:
```
                let widgetIcon = widget.node._def.icon || 'fa fa-question'
                if (widget.node._def.icon.startsWith('font-awesome/')) {
                    widgetIcon = 'fa ' + widget.node._def.icon.replace('font-awesome/', '')
                }

                if (widget.isSubflowInstance) {
                    // In this MVP, subflow instances are constrained to stay within own group.
                    container.parent().addClass('red-ui-editableList-item-constrained')
                    // change the icon to use the built-in subflow icon
                    widgetIcon = 'nrdb2-sb-subflow-icon'
                    // apply a tooltip to further clarify this is a subflow
                    titleRow.attr('title', widget.subflowName || 'Subflow instance')
                }

                if (widgetIcon == 'nrdb2-sb-subflow-icon' || widgetIcon.startsWith('fa fa-')) {
                    $('<i>', { class: 'nrdb2-sb-icon nrdb2-sb-widget-icon ' + widgetIcon }).appendTo(titleRow)
                }
                else {
                    const href = 'resources/' + widget.type + '/' + widgetIcon
                    $('<image>', { class: 'nrdb2-sb-icon nrdb2-sb-widget-icon', href: href }).appendTo(titleRow)
                }
                $('<span>', { class: 'nrdb2-sb-title' }).text(widget.label?.trim() || widget.id).appendTo(titleRow)
```
But the 'resources/...' path is not always correct.
For example when I use a png icon in one of my (custom) ui nodes, and I put it in this path:
image

Then it is being found by Node-RED and displayed in the flow editor, but from this path:

image

Not sure where those paths are being stored in the Node-RED flow editor, for every node type.
If anybody has an idea, don't hesitate to let me know!

bartbutenaers:
Closing this PR because it only took into account fontawesome icons...

PR #1165, opened 2024-08-06
DRAFT - Node icons in sidebar

bartbutenaers:
## Description

Fix for [pull request](https://github.com/FlowFuse/node-red-dashboard/pull/1151).
The other PR assumed that all ui nodes use a FontAwesome icon, however some nodes have a custom (png, svg, ...) icon.

I had a look in the Network tab of the developer tools, and seems that all node icons are being loaded like this:
`.../icons/<node_module_name>/<icon_resource_name>`.  So I use the same mechanism, and my png test icon seems to be loaded correctly.

**_Known issue_**: the icons of nodes are always white on a transparant background, so they are nicely visible on top of their node in the flow editor.  However since the dashboard sidebar (with my current theme) is white, you won't see the icon although it is available:

image

Not sure how to solve this.  CSS isn't my best friend...

## Related Issue(s)

None

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.html
+++ b/nodes/config/ui_base.html
@@ -1038,7 +1038,12 @@
             addItem: function (container, i, /** @type {DashboardItem} */ widget) {
                 const titleRow = $('<div>', { class: 'nrdb2-sb-list-header nrdb2-sb-widgets-list-header' }).appendTo(container)
                 $('<i class="nrdb2-sb-list-handle nrdb2-sb-widget-list-handle fa fa-bars"></i>').appendTo(titleRow)
-                let widgetIcon = 'fa fa-image'
+
+                let widgetIcon = widget.node._def.icon || 'fa fa-question'
+                if (widget.node._def.icon.startsWith('font-awesome/')) {
+                    widgetIcon = 'fa ' + widget.node._def.icon.replace('font-awesome/', '')
+                }
+
                 if (widget.isSubflowInstance) {
                     // In this MVP, subflow instances are constrained to stay within own group.
                     container.parent().addClass('red-ui-editableList-item-constrained')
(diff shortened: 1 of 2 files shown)


joepavitt:
The image is there, but the CSS isn't behaving too well as the icon has `0 x 0` sizes. We can fix the icons to be `svg` in a later PR which would correct this, but the odd alignment problems are caused by the CSS atm. I'll see if I can propose a quick fix

joepavitt:
Now the icons are rendering (within the constraints) I set at `1600px` in width and height - the joys of SVG

joepavitt:
Gauge and Markdown are the only core nodes that have SVG icons, for `svg` to use the `currentColor` for `fill`, which is what we'd need here, we'll need to load the SVG content itself and render that - not use an `<image />` - that's for a follow up PR.

This is great step forward - thanks Bart

joepavitt (approved):

nodes/config/ui_base.html
```diff
@@ -1038,7 +1042,12 @@
             addItem: function (container, i, /** @type {DashboardItem} */ widget) {
                 const titleRow = $('<div>', { class: 'nrdb2-sb-list-header nrdb2-sb-widgets-list-header' }).appendTo(container)
                 $('<i class="nrdb2-sb-list-handle nrdb2-sb-widget-list-handle fa fa-bars"></i>').appendTo(titleRow)
-                let widgetIcon = 'fa fa-image'
+
+                let widgetIcon = widget.node._def.icon || 'fa fa-question'
+                if (widget.node._def.icon.startsWith('font-awesome/')) {
```
Steve-Mcl, line 1047: This breaks the sidebar. subflows icon can be a function.

image

PR #1166, opened 2024-08-06
Theme: Add Row Height option (including "Compact")

### LINKED ISSUE #1095: Theme: Compact View
### Description

It's coming up too regulsrly for us to ignore the need to offer a "compact" view.

Vuetify offers this as an option, and we need to ensure our own, custom-built widgets offer the same. 

It should be an option on the "Theme", and all widgets need to be able to inherit this options (whether that's via CSS or JS) and render accordingly.

joepavitt:
## Description

- Adds a new "Row Height" field to the UI Themes
- Defaults to "Default (48px)"
    - Also provides "Comfortable (38px)" and "Compact (32px)
- Originally tried to do this solely through Vuetify's "Density" property, but it sizes heights very inconsistently for each widget, plus it didn't cover our own custom widgets
    - Ended up still setting this in `Baseline.vue`, but had to add in overrides for `ui-button` as that behaves differently to all others.
- Consistency in use of `widget-row-height` CSS variable to drive row height (and inheritance into widgets)
- Add in documentation for the new "Row Height" option

## Related Issue(s)

Closes #1095

### Default

https://github.com/user-attachments/assets/c8ac414b-6262-4408-8972-3833b53fcc68

### Comfortable

https://github.com/user-attachments/assets/ee1f2d28-8bd4-4ed2-b971-ed2f3c5ffe65

### Compact

https://github.com/user-attachments/assets/78c451fc-922a-425d-8a5b-babdd01b1b42


joepavitt:
Perhaps the most noticeable differences is in the UI Table:

### Compact
Screenshot 2024-08-06 at 17 28 00

### Comfortable
Screenshot 2024-08-06 at 17 27 48

### Default
Screenshot 2024-08-06 at 17 28 15

thebaldgeek:
Thanks for the tag, it does look interesting. 
Have you tested it on mobile? I wonder how the (card) view looks on the smaller screen.
BTW, I could not use this ui_table node on the website since it only displays 10 rows before the uncontrollable pagination cuts in and makes the table take up a ton more height and sort of breaks mobile view.
Great to see some progress to reducing the default jumbo size.

joepavitt:
> before the uncontrollable pagination cuts in

It is very controllable via the "Max Rows" option:

https://dashboard.flowfuse.com/nodes/widgets/ui-table.html#pagination

gayanSandamal (approved):
This looks good to me

PR #1170, opened 2024-08-07
DRAFT - Table send clicked column

bartbutenaers:
## Description

Hi guys,
When you select a row in a ui-table widget, the clicked column will be send by this PR in the `msg.column` property:

table_column

I have this PR marked as ***draft*** because I wasn't sure whether the `msg.column` property is ok?  Therefore I also haven't yet adapted any documentation...

## Related Issue(s)

None

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-table/UITable.vue
+++ b/ui/src/widgets/ui-table/UITable.vue
@@ -21,12 +21,12 @@
         <template #item="{ item, index, internalItem, isSelected, toggleSelect }">
             <tr
                 :class="{'nrdb-table-row-selectable': props.selectionType === 'click', 'nrdb-table-row-selected': selected === item}"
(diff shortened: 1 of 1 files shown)


joepavitt:
I assume there is little value in this case now, but the linked PR re: button click would add further value?

bartbutenaers:
Yes true.  I created this PR because I wanted this morning to show a recycle-bin icon in the last column.  When clicked I could remove the corresponding data.  But after some testing it is hard to see where you click, because the table shows no vertical lines...
That is why I created the next PR with the buttons.
So if you don't see added value, you can close this one...

joepavitt:
I think your buttons PR covers what we'd like to see for this behaviour

PR #1185, opened 2024-08-08
Fix: Defaults set before Vuetify loading

joepavitt:
## Description

Noticed this error starting to appear:

Screenshot 2024-08-08 at 16 49 39

Seems to have been introduced in https://github.com/FlowFuse/node-red-dashboard/pull/1166 whereby the defaults are trying to be set, before Vuetify has fully loaded them, wbhich also meant #1166 never fully worked when delivered too.


gayanSandamal (approved):
LGTM

PR #1186, opened 2024-08-08
UI Template - Run generic JS onMounted

joepavitt:
## Description

- We group any generic JS we find into a `beforeCreate` function, which was then run in the `beforeCreate` lifecycle hook of VueJS. This was causing problems though if a user was trying to reference HTML elements in their component e.g. https://discourse.nodered.org/t/graph-shapes-and-single-point/90041/4 as the component hadn't yet been mounted.
- This PR moves the `beforeCreate` content (i.e. generic JS) into a new hok called `js`, and runs this code in `mounted()` instead so that the code _can_ reference HTML content

### DIFF
--- a/ui/src/util/vue.acorn.js
+++ b/ui/src/util/vue.acorn.js
@@ -115,7 +115,7 @@ function parseSFC (js) {
     } if (tree.type === 'Program' && tree.body.length > 0) {
         // we have raw JavaScript.
         // - this is fine, but we need to wrap it in a Vue Component to render in Dashboard
-        // assusming this runs straight away, let's add it to the beforeCreate of the component
+        // assuming this runs straight away, let's document this in a "beforeCreate" lifecycle hook
         const component = {}
         component.beforeCreate = ''
         tree.body.forEach((block) => {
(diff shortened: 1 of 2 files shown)


ui/src/util/vue.acorn.js
```diff
@@ -115,11 +115,11 @@ function parseSFC (js) {
     } if (tree.type === 'Program' && tree.body.length > 0) {
         // we have raw JavaScript.
         // - this is fine, but we need to wrap it in a Vue Component to render in Dashboard
-        // assusming this runs straight away, let's add it to the beforeCreate of the component
+        // assuming this runs straight away, let's document this in a "beforeCreate" lifecycle hook
```
Steve-Mcl, line 118:
```suggestion
        // assuming this runs straight away, let's add this to a new lifecycle hook named "js" that is executed on the "mounted" hook
```

suggestions

Steve-Mcl (approved):
Approved with an optional comment update.

One (probably naive) question though, why `mounted` and not `beforeMount`?

would `beforeMount` not give users an opportunity to "do stuff" before its displayed? (or would it simply not work due to not being present in the DOM)?

joepavitt:
>  (or would it simply not work due to not being present in the DOM)?

This - `beforeMount` would still not permit code to reference elements in the DOM

PR #1196, opened 2024-08-10
Fix blank side bar

### LINKED ISSUE #1195: Blank Sidebar when def icon is a function (subflows)
This breaks the sidebar. subflows icon can be a function.

image

_Originally posted by @Steve-Mcl in https://github.com/FlowFuse/node-red-dashboard/pull/1165#discussion_r1712632661_

Steve-Mcl:
closes #1195

## Description

* use RED.utils to get correctly resolved icon

NOTE: some icons from contrib nodes do not use transparency and cannot be faithfully represented. If we want to support that then the only reliable way would be to generate the tree entries like nodes (as they appear in the left hand palette.)

### Demo
image

## Related Issue(s)

#1195

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.html
+++ b/nodes/config/ui_base.html
@@ -177,16 +177,17 @@
         width: 14px;
         text-align: center;
     }
-    /* apply subflow icon */
-    .nrdb2-sb-widgets-list-header .nrdb2-sb-widget-icon.nrdb2-sb-subflow-icon {
+    /* common styles for images */
+    .nrdb2-sb-widgets-list-header .nrdb2-sb-widget-icon.nrdb2-sb-widget-icon-img {
         background-color: currentColor;
         display: inline-block;
-        mask-image: url(red/images/subflow_tab.svg);
         mask-size: contain;
         mask-position: center;
+        mask-repeat: no-repeat;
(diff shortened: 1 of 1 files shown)


bartbutenaers:
Hi @Steve-Mcl,
Thanks for the PR!  Really appreciated!!
What a bummer that you had to spend your saturday on this one 😞

It looks like your implementation support a wide range of icon types.

The ´RED.utils.getNodeIcon´ was the one I was looking for last week.  The missing link...

> some icons from contrib nodes do not use transparency and cannot be faithfully represented

Is that a real problem?  Currently there are only a few ui nodes for D2.  The developers or the users will quickly see that something is wrong with the icon and fix it.

Or am I mistaken?   Because from your screenshot it looks like you have also non-ui nodes in your sidebar tree?

joepavitt (approved):

joepavitt:
> Because from your screenshot it looks like you have also non-ui nodes in your sidebar tree?

They're [subflows](https://dashboard.flowfuse.com/user/subflows.html) :)

PR #1206, opened 2024-08-13
UI Table - Configurable "Mobile" view

joepavitt:
## Description

- Configurable via the new `mobileBreakpoint` property
- Applies typedInput which provides:
    - `None` - always render full table
    - 'px' - define a custom `px` value to act at the breakdpoint
- Need to update documentation

## Related Issue(s)

Closing #76 

## Checklist


joepavitt:
Ended up being more hands on than I thought, although the new `:mobile=""` did help, at least with the skeleton structure. 
I've then had to implement the row styling ourselves as we use a custom `<tr>` element.

ui/src/widgets/ui-table/UITable.vue
```diff
@@ -183,7 +192,41 @@ export default {
...
+        updateIsMobile () {
+            this.isMobile = this.checkBreakpoint()
+        },
+        checkBreakpoint () {
+            let table = this.$refs.table
+            if (!table) {
+                return false
+            }
+            table = table.$el
+            const width = table.clientWidth
```
gayanSandamal, line 205: @joepavitt I've noticed that the actual device screen sizes are not aligning with defined breakpoints.

Therefore I think this should be `window.innerWidth` or `screen.width` instead as this breakpoint checks the table's width even the ui table properties show mobile, table and desktop.

Also when it comes to css media queries, it checks browser window width as usual.

Available breakpoints 

Screenshot 2024-08-14 at 14 19 28

Selected breakpoint is 768px

Screenshot 2024-08-14 at 14 18 35

But the table got re-arranged at a different breakpoint

Screenshot 2024-08-14 at 14 22 16

joepavitt: > I've noticed that the actual device screen sizes are not aligning with defined breakpoints.

This is because it's based on the table width, and is intentional, maybe the labelling/help needs to be clearer. If you have a table that is 50% the width of the screen, then I think we should be switching to card view if that table is too wide for it's space

gayanSandamal: Got it, I understand the use case now. Switching to card view when the table is too wide for its space makes sense. Thanks for clarifying!

joepavitt: I agree thought, that the docs/help should make this cleaer, so, I'll add that in before we merge

gayanSandamal: Thanks, I'll hold off on this for now then

Issue #1215, opened 2024-08-17
Apparent layout change on small screens since 1.14.0


colinl:
### Current Behavior

I am not sure if this is a regression or an intended change.  I have a page with Layout Fixed and Theme Sizings settings:

![image](https://github.com/user-attachments/assets/3e00f52a-05d7-4ac7-9967-167bafed71b4)

Using version 1.14.0 of the dashboard, in a small window, and on a tablet, with three groups each 3 wide, it looks like 

![Screenshot from 2024-08-17 08-48-51](https://github.com/user-attachments/assets/2b071b5e-029f-41c2-97cc-85ecd787a896)

but with the current `main` branch (16th Aug 2024), in the same window size, it only shows two groups across, pushing the third off the screen.

![image](https://github.com/user-attachments/assets/14934ce0-2349-4498-a8e6-cd1bc0949400)

Is this an intended change, and if it is how do I restore the previous functionality please?

### Environment

- Dashboard version: Current main branch, commit cd5d1ce42a
- Node-RED version: 4.0.2
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

joepavitt:
Hmmm, certainly not intentional - the padding/sizings look consistent, it appears the gauges themselves are rendering wider?

colinl:
Running two instances of node-red, both with only the dashboard installed, one with 1.14.0 and one with main from github, both running only the flow below, which is three text widgets, each 3x1, in three groups of size 3x1, in a PC I see

![image](https://github.com/user-attachments/assets/1a7093e8-2c39-46fe-8a73-6f1225cec5fc)

That is both running at the same time with a page open on each in Edge browser.  It can be seen that the latest main has significantly larger groups than 1.14.0, which means they do not fit on the tablet.  There are no CSS overrides, only those three nodes and default theme.

I notice that the colour of the text has changed too, was that intentional?

```
[...]

PR #1218, opened 2024-08-20
UI Text - Ensure we check props.style before applying custom styling

joepavitt:
## Description

Fixes https://github.com/FlowFuse/node-red-dashboard/issues/1215#issuecomment-2298729673 specifically, part of #215, where the rest still needs investigation.


gayanSandamal (approved):
This fix looks good to me.

Tested locally with and without custom styles.
No console errors ✅
Fix worked ✅

Screenshot 2024-08-21 at 12 48 17

PR #1222, opened 2024-08-21
v1.15.0

gayanSandamal:
## Description

v1.15.0 version

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@flowfuse/node-red-dashboard",
-    "version": "1.14.0",
+    "version": "1.15.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@flowfuse/node-red-dashboard",
-            "version": "1.14.0",
+            "version": "1.15.0",
             "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


joepavitt (approved):

colinl:
Is it a good idea to release before the regression in #1215 is fixed?  That means the layout will change again when it is fixed.

joepavitt:
> Is it a good idea to release before the regression in https://github.com/FlowFuse/node-red-dashboard/issues/1215 is fixed? That means the layout will change again when it is fixed.

We can't hold off releases until every bug is fixed. We iterate, we've made big improvements and fixes elsewhere, and they shouldn't be held up for this.

I think I know the problem with that, and will address it on the issue directly.

colinl:
OK, I thought the fact that this a regression might make a difference to this policy.

joepavitt:
It's definitely a priority - but not a blocker per-say. I've opened #1223 which covers this anyway, and we will be holding off merging the `1.15.0` release until it's resolved.

colinl:
Great, thanks.

PR #1223, opened 2024-08-21
Ensure groups in Fixed layouts are driven by the correct width property

### LINKED ISSUE #1215: Apparent layout change on small screens since 1.14.0

joepavitt:
## Description

Fix #1215 

Groups were rendered at a static width because of a `width` property - this should not be the case, they should be driven by their group's width.

### DIFF
--- a/ui/src/layouts/Flex.vue
+++ b/ui/src/layouts/Flex.vue
@@ -125,7 +125,6 @@ export default {
 }
 
 .nrdb-layout--flex > div {
-    width: var(--layout-card-width);
     max-width: 100%;
 }
 


gayanSandamal (approved):
Fixed group width is working as expected.
Looks good to me

https://github.com/user-attachments/assets/79da0c5a-c86f-4ea8-973d-e5852f2e0fdb

PR #1234, opened 2024-08-25
Layouts: Ensure we're parsing float on the "px" row height

### LINKED ISSUE #1215: Apparent layout change on small screens since 1.14.0

joepavitt:
## Description

When reconciling the `rowHeight` into the CSS vars, made an error on Fixed layouts which meant that the `48px` was then included ins a JS calculation for the width, failing (because of the inclusion of `px`) and then not setting the width properly.

This PR ensures that `rowHeight` is a number, before injecting it into the width calculations.

## Related Issue(s)

Fixes #1215

### DIFF
--- a/ui/src/layouts/Group.vue
+++ b/ui/src/layouts/Group.vue
@@ -30,7 +30,6 @@ export default {
     },
     computed: {
         columns () {
-            console.log(this.group.width)
             return this.group.width
         }
     },
--- a/ui/src/layouts/Group.vue
+++ b/ui/src/layouts/Group.vue
@@ -30,6 +30,7 @@ export default {
     },
     computed: {
         columns () {
+            console.log(this.group.width)
             return this.group.width
         }
     },
--- a/ui/src/layouts/Flex.vue
+++ b/ui/src/layouts/Flex.vue
@@ -49,7 +49,7 @@ export default {
     data () {
         const rowHeight = getComputedStyle(document.body).getPropertyValue('--widget-row-height')
         return {
-            rowHeight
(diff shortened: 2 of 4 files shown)


ui/src/layouts/Group.vue
```diff
@@ -30,6 +30,7 @@ export default {
     },
     computed: {
         columns () {
+            console.log(this.group.width)
```
Steve-Mcl, line 33:
```suggestion
```

Steve-Mcl (approved):
Approved with suggestion

ui/src/layouts/Group.vue
gayanSandamal: Fixed the above as Joe is not available today

gayanSandamal (approved):
Removed the console statement as Joe is not available today.

Looks good to me.

PR #1235, opened 2024-08-25
UI Template: Improves in-Editor Docs & Initial Example

joepavitt:
## Description

Coming up in the forums a lot about how to bind attributes with variables. Whilst we detail this extensively in the online-docs, we do not include an example out of the box, nor in the in-Editor help documentation.

### DIFF
--- a/nodes/widgets/locales/en-US/ui_template.html
+++ b/nodes/widgets/locales/en-US/ui_template.html
@@ -11,58 +11,49 @@ Provides a means to render custom content or a set of node-styles depending on s
 
 ### Type: Widget
 When the scope is set to **Widget**, the template will be rendered in the Dashboard as HTML.
-You can render dynamic content using any VueJS data-binding expressions, (e.g. `v-if`, `v-for`)
-and access incoming data to the node via the `msg` object, e.g:
 
-<pre>
-&lt;template&gt;
-  &lt;div&gt;
-      &lt;h2&gt;Counter&lt;/h2&gt;
-      &lt;p&gt;Current Count: {{ count }}&lt;/p&gt;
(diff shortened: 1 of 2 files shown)


nodes/widgets/locales/en-US/ui_template.html
```diff
@@ -11,58 +11,49 @@
...
+and access incoming data to the node via the `msg` object.
+
+#### Working with Variables
+
+Any variables that you want to render into your `<template />` are done so in one of two ways:
+
+- **Attribute Binding** - Use `:` to bind a variable to an attribute. Anything inside the `""` here is treated as JavaScript, for example: 
+
+```html
+&lt;p :class="msg.payload">Hello World&lt;/p>
```
Steve-Mcl, line 25:
```suggestion
\<p :class="msg.payload">Hello World\</p>
```

Joe, these can all be escaped simply with a `\`

Do you mind if I edit the file and replace them all in one go instead of individual suggestions?

joepavitt: > Do you mind if I edit the file and replace them all in one go instead of individual suggestions?

Please do

Steve-Mcl: bah, unfortunately the markdown processor used in NR doesnt correctly use `\` as an escape 😢 (it should)

Steve-Mcl:
I have 
* removed the unnecessary extra `&gt;` escapes to tidy it up a little 
* changed a couple of missed `<` entries in the doc
* added one more section i felt was an important example:


Here is how it renders:

image

image

gayanSandamal:
> I approve but since a few changes were made by me, i should probably wait until @gayanSandamal approves before merge

Thanks @Steve-Mcl  I will review and approve this

Issue #92, opened 2023-07-30
All nodes that accept input to have a typedInput


Steve-Mcl:
### Description

To avoid having to split up and shift around properties in my object payload, please support a standard "property" input as a `typedInput`.  Same as the built in switch node, file, range, etc, nodes do

![image](https://github.com/flowforge/flowforge-nr-dashboard/assets/44235289/184f2856-e40c-4d0a-b40d-8b394f479251)
![image](https://github.com/flowforge/flowforge-nr-dashboard/assets/44235289/adf64214-d41f-47d3-ab88-3f46c2e7fe43)
![image](https://github.com/flowforge/flowforge-nr-dashboard/assets/44235289/697e4b78-ef67-45a8-9578-1f9c66e262fc)

I would recommend, to keep this simple, it would default to `msg.payload` and would only support `msg` and `JSONata`.  

Anyone who knows JSONata can use `performance * availability * quality` directly in the input property
Anyone who doesn't know or like JSONata, can use a math node or other nodes to compute the OEE value then pass it into the ui-text node

This means given this payload:
```
{
  name: "Cooling Stage 1"
  time: '16:00',
  cycleCount: 800,
  availability: 0.938,
  performance: 0.933,
  quality: 0.983
  OEE, "86%"
  productionHours: 8,
  partCount: 700,
  targetCount: 800,
  scrap: 12
}
```

I can change this...
![image](https://github.com/flowforge/flowforge-nr-dashboard/assets/44235289/e75875dc-2951-4850-83f0-b704fdc5e76a)

into this...

![image](https://github.com/flowforge/flowforge-nr-dashboard/assets/44235289/63c8fc29-ac09-4a66-b163-6bafb522f504)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

phedders:
This is a genius solution - would love to see it progressed!

Steve-Mcl:
Pushed to back log due to design oversight. Will be picked up again at a later time

Issue #30, opened 2023-06-22
Visual Layout Editor


rcarmo:
### Description

One to the unappreciated (and somewhat obscure) features of the current solution is the integration with the main Node-RED environment and the ability to do a baseline visual layout without fully leaving the flow editor. The widget grid and sizing system is a bit clunky (the emphasis on units rather than percentages/free resizable elements, etc.), but it works pretty well for multiple scenarios (like the example below):

![](https://taoofmac.com/media/blog/2021/12/01/1920/CS5Qvd_0UfnXxJr3qkvOh33wP-I=/dashboard.jpg)

### Have you provided an initial effort estimate for this issue?

I am no FlowForge team member

joepavitt:
Just thinking through this (out loud) a little.

We have a new set of [events](https://flowforge.github.io/flowforge-nr-dashboard/contributing/guides/events.html) that we could add a `widget-layout` event too, or something along those lines. 

The ideal (end goal) UX would be as follows:

1. Enable "edit" mode (maybe via a shortcut, or option in the UI, or via config on the page in Node-RED, the latter feels clunky?)
2. All groups & widgets then allow for dragging & dropping (for ordering) & corner/side grabbing for resizing.
3. Upon re-ordering and/or resizing, the `widget-layout` event send a message back to Node-RED for the config to be updated (i.e. the options int he editor now reflect the new layout already).

The challenge is that how to re handle the nature of re-deploying? The NR editor would have changes not deployed, but that are actually reflected in the UI/Dashboard, because we've created those values locally in the client.

bartbutenaers:
@joepavitt,
Perhaps you might also find some useful ideas in Flexdash about this, like [this](https://discourse.nodered.org/t/flexdash-alpha-release-round-3/72562/56?u=bartbutenaers) demo from Thorsten to create a layout via the config screen...
Bart

joepavitt:
Thanks for sharing @bartbutenaers - exactly what I had in mind.

Hyperlooper:
First off, let me say thank you for this amazing project. I am blown away at the progress. 

I wanted to put my hat in the ring to say that a visual layout editor would definitely be a feature I would miss if it were absent. For context, here's an example of a dashboard I am working on in dashboard 1.0 that is being used on a television kiosk. Would this layout be possible in 2.0 without a visual layout editor? (I'm asking out of ignorance, I don't currently understand how I would accomplish this layout sans the visual editor). 

[...]

PR #1237, opened 2024-08-25
UI notification improvements

Steve-Mcl:
## Description

This is not tied directly to any one ISSUE. 

This was based on recent forum feedback and my desire to support TypedInputs throughout DB2 widgets.

The changes implemented are opinionated and may not all may be welcomed but since I had some playtime, and I like where it got to, I thought I would raise a PR for feedback.

At present, this PR includes no doc updates or tests because if the changes are rejected I wont have wasted any time ;) 

### Whats changed

* **Support `TypedInput`**
  * What: This PR introduces a pattern for supporting typed inputs for the payload.
  * Why: Because is super annoying adding change nodes everywhere
  * How: It works by evaluating node config `property`/`propertyType` and adding a temporary prop `ui_payload` to the msg that is used client side as the `value`. This should work for other widgets too.
* **Inhibit default passthru**
  * Why: Currently, the notification node sends multiple messages - 1 every time it gets and input msg and 1 when it closes. 
  * What: This PR only sends the final msg upon timeout/dismiss/confirm
  * How: This is done by setting the config prop to `false` (hard coded)
* **Adds `ui_reason` to `msg`
  * `msg.ui_reason` will be added to the msg containing the relevant close reason of `timeout`, `input_msg`, `dismiss_clicked`, or `confirm_clicked`
* ** Only sends msg to Node-RED if `show` is true**
  * Why: I could quite easily get notification node to emit both "confirm" or "dismiss" AND "timeout" (when clicking a button near the end of progress bar)
 

### Demo
chrome_zRLbx6gi7J

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -977,6 +977,10 @@ module.exports = function (RED) {
                                 if (widgetConfig.topic || widgetConfig.topicType) {
                                     msg = await appendTopic(RED, widgetConfig, wNode, msg)
                                 }
+                                if (widgetConfig.property || widgetConfig.propertyType) {
+                                    // append ui_payload to msg if property/propertyType is set
+                                    msg = await updatePayload(RED, widgetConfig, wNode, msg)
+                                }
                                 if (hasProperty(widgetConfig, 'passthru')) {
                                     if (widgetConfig.passthru) {
                                         send(msg)
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -4,7 +4,7 @@ const path = require('path')
 const v = require('../../package.json').version
 const datastore = require('../store/data.js')
 const statestore = require('../store/state.js')
-const { appendTopic, addConnectionCredentials } = require('../utils/index.js')
+const { appendTopic, addConnectionCredentials, updatePayload } = require('../utils/index.js')
 
 // from: https://stackoverflow.com/a/28592528/3016654
 function join (...paths) {
--- a/nodes/utils/index.js
+++ b/nodes/utils/index.js
@@ -30,6 +30,28 @@ async function appendTopic (RED, config, wNode, msg) {
 }
 
 /**
+ * Evaluates the property/propertyType and sets ui_payload in the message object
+ * This leaves the original payload untouched
+ * This permits an TypedInput widget to be used to set the payload
+ * @param {*} RED - The RED object
+ * @param {Object} config - The node configuration
+ * @param {Object} wNode - The node object
+ * @param {Object} msg - The message object
+ * @returns {Object} - The updated message object
+ */
+async function updatePayload (RED, config, wNode, msg) {
+    if (config.propertyType && config.property) {
+        try {
+            msg.ui_payload = await asyncEvaluateNodeProperty(RED, config.property, config.propertyType || 'msg', wNode, msg) || ''
+        } catch (_err) {
+            // do nothing
+            console.error(_err)
+        }
+    }
+    return msg
+}
+
+/**
  * Adds socket/client data to a msg payload, if enabled
  *
  */
@@ -57,5 +79,6 @@ function addConnectionCredentials (RED, msg, conn, config) {
 module.exports = {
     asyncEvaluateNodeProperty,
     appendTopic,
-    addConnectionCredentials
+    addConnectionCredentials,
+    updatePayload
 }
--- a/nodes/widgets/ui_notification.html
+++ b/nodes/widgets/ui_notification.html
@@ -27,7 +27,9 @@
                 confirmText: { value: 'Confirm' },
                 raw: { value: false },
                 className: { value: '' },
-                name: { value: '' }
+                name: { value: '' },
+                property: { value: 'payload' },
+                propertyType: { value: 'msg' }
             },
             inputs: 1,
             outputs: 1,
@@ -37,6 +39,12 @@
             label: function () { return this.name || (this.position === 'prompt' ? 'show dialog' : (this.position === 'dialog' ? 'show dialog' : 'show notification')) },
             labelStyle: function () { return this.name ? 'node_label_italic' : '' },
             oneditprepare: function () {
+                $('#node-input-property').typedInput({
+                    default: 'msg',
+                    typeField: $('#node-input-propertyType'),
+                    types: ['str', 'msg', 'jsonata']
+                })
+
                 $('#node-input-topic').typedInput({
                     default: 'str',
                     typeField: $('#node-input-topicType'),
@@ -86,6 +94,11 @@
     <div class="form-row">
         <label for="node-input-name"><i class="fa fa-tag"></i> Name</label>
         <input type="text" id="node-input-name" placeholder="Name">
+    </div>
+    <div class="form-row">
+        <label for="node-input-property"><i class="fa fa-ellipsis-h"></i> Message</label>
+        <input type="hidden" id="node-input-propertyType">
+        <input type="text" id="node-input-property" placeholder="payload" style="width: 70%;">
     </div>
     <div class="form-row">
         <label for="node-input-ui"><i class="fa fa-bookmark"></i> UI</label>
--- a/nodes/widgets/ui_notification.js
+++ b/nodes/widgets/ui_notification.js
@@ -5,6 +5,7 @@ module.exports = function (RED) {
         const node = this
 
         RED.nodes.createNode(this, config)
+        config.passthru = false // prevent default passthru by setting it explicity to `false`. The notification itself will send msg on timeout, dismissal or confirmation!
 
         // Which ui are we rendering this widget.
         // In contradiction to other ui nodes (which belong to a group), the notification node belongs to a ui instead.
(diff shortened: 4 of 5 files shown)


Steve-Mcl:
> **Support `TypedInput`**
> * What: This PR introduces a pattern for supporting typed inputs for the payload.
> * Why: Because it is super annoying adding change nodes everywhere!
>   
>   * [All nodes that accept input to have a typedInput #92](https://github.com/FlowFuse/node-red-dashboard/issues/92)
> * How: It works by evaluating node config `property`/`propertyType` and adding a temporary prop `ui_payload` to the msg that is used client side as the `value`. This should work for other widgets too.
> * NOTE: Adding JSONata might be an alt solution to the whole "support mustache for label / text / title / etc" debate
>   
>   * which makes me thing the temp `ui_payload` should probably be an object like `ui_computed` so that we can do the custom label stuff too e.g. `{ ui_computed: { payload: "evaluated payload", "label": "evaluated label" }`

Another avenue may be to utilise the existing mechanism of `ui_update` and add the evaluated `payload` into that object 🤔

bartbutenaers:
@Steve-Mcl,
My few cents about your proposal, for what it is worth:

1. I really like the idea of using Jsonata to mimic the Mustache expressions from dashboard D1.  It is very powerful and then indeed you don't have to pollute your flows with Template nodes.  Cool idea!

2. Personally I am not a fan to move the current ouput `msg.payload` to `msg.ui_reason` (in order to be able to keep the original input `msg.payload`).  I know that it is advised that nodes should reuse and manipulate existing input messages (by adding data to it), instead of creating new output messages from scratch.  In order to keep the (important) info from the input message.  However the state of a ui nodes is mostly build upon multiple input messages.  The last message might only contain some (unimportant) `ui_update` stuff, so not really to reuse such messages.  Or do you mean that you reuse the last message stored in the data store perhaps?  Because most ui nodes currently only store messages which have a payload, it might indeed be useful to reuse that one...

   Moreover when you use `msg.ui_reason` your flow again needs extra nodes.  Because currently I can attach any kind of node to the output of my notification node, because the payload contains all the required info.  When you start using `msg.ui_reason` I need to add Change nodes to move that reason into the payload again...

3. Only sends msg to Node-RED if show is true, makes sense to me.  But could it perhaps be solved by moving the 
[clear timer and interval](https://github.com/FlowFuse/node-red-dashboard/blob/main/ui/src/widgets/ui-notification/UINotification.vue#L171) code snippet BEFORE the `socket.emit` (because the emitting perhaps takes a bit too long)?

4. The default passthrough of messages is indeed something that was confusing for me while implementing my PR for the notification node.  So indeed I would disable it.

Steve-Mcl:
> 2\. Moreover when you use `msg.ui_reason` your flow again needs extra nodes.  Because currently I can attach any kind of node to the output of my notification node, because the payload contains all the required info.  When you start using `msg.ui_reason` I need to add Change nodes to move that reason into the payload again

The current implementation of the UI node only sends "timeout" or "confirm" etc. What I did here was to move the reason for it closing into `ui_reason` so that the original payload could stay in payload and therefore be passed back for downstream use. And yes the msg passed back will be the last message sent in to that particular notification node.

Steve-Mcl:
> 3\. Only sends msg to Node-RED if show is true, makes sense to me.  But could it perhaps be solved by moving the
>     [clear timer and interval](https://github.com/FlowFuse/node-red-dashboard/blob/main/ui/src/widgets/ui-notification/UINotification.vue#L171) code snippet BEFORE the `socket.emit` (because the emitting perhaps takes a bit too long)?

I've been bitten by timeouts in the past. 

I did also move the clear calls above the emit but i could still cause dual messages.  The interlock was a final catch all that stopped it happening.

joepavitt:
I need to spend time on this before committing either way, and give it the attention it deserves, will quickly comment on:

> Another suggestion on the forum was multiple outputs. That would remove the need for any property at all.

Whilst it's clean from the notification node perspective, joining these in order to compute on them is then a pain as you have two async messages.

colinl:
> I need to spend time on this before committing either way, and give it the attention it deserves, will quickly comment on:
> 
> > Another suggestion on the forum was multiple outputs. That would remove the need for any property at all.
> 
> Whilst it's clean from the notification node perspective, joining these in order to compute on them is then a pain as you have two async messages.

I think the confusion is that early on there was a suggestion to move just the result to a separate output, but I think that has been superseded by the suggestion of sending just to one output of the three, dependent on what is clicked. Effectively including a Switch node inside the Notification node.

Steve-Mcl:
> Does this PR already include adding a note to the docs and help text to say that empty or 0 timeout disables the timeout?

No Colin.  No docs done at all - very much PoC at the moment.

Will make a mental note to include that kind of info if/when it evolves into something.

colinl:
In fact that is actually just something missing from the docs for the existing node, it isn't anything to do with the suggested enhancements.

ui/src/widgets/data-tracker.mjs
```diff
@@ -11,6 +11,7 @@ export function useDataTracker (widgetId, onInput, onLoad, onDynamicProperties)
     const socket = inject('$socket')
 
     function checkDynamicProperties (msg) {
+        console.log('checkDynamicProperties', msg)
```
Steve-Mcl, line 14:
```suggestion
```

remove before merge

ui/src/widgets/data-tracker.mjs
```diff
@@ -52,16 +53,19 @@ export function useDataTracker (widgetId, onInput, onLoad, onDynamicProperties)
             socket.on('widget-load:' + widgetId, (msg, state) => {
                 // automatic handle state/dynamic  updates for ALL widgets
                 if (state) {
+                    console.log('widget-load:  store.commit "ui/widgetState"', widgetId, state)
```
Steve-Mcl, line 56:
```suggestion
```

remove before merge

ui/src/widgets/data-tracker.mjs
```diff
@@ -52,16 +53,19 @@ export function useDataTracker (widgetId, onInput, onLoad, onDynamicProperties)
...
                     store.commit('ui/widgetState', {
                         widgetId,
                         config: state
                     })
                 }
                 // then see if there is custom onLoad functionality to deal with the latest data payloads
                 if (onLoad) {
+                    console.log('widget-load: onLoad', msg)
```
Steve-Mcl, line 64:
```suggestion
```

remove before merge

ui/src/widgets/data-tracker.mjs
```diff
@@ -52,16 +53,19 @@ export function useDataTracker (widgetId, onInput, onLoad, onDynamicProperties)
...
                     onLoad(msg)
                 } else {
                     if (msg) {
+                        console.log('widget-load: store.commit "data/bind"', widgetId, msg)
```
Steve-Mcl, line 68:
```suggestion
```

remove before merge

ui/src/widgets/data-tracker.mjs
```diff
@@ -71,6 +75,7 @@ export function useDataTracker (widgetId, onInput, onLoad, onDynamicProperties)
             })
             // This will on in msg input for ALL components
             socket.on('msg-input:' + widgetId, (msg) => {
+                console.log('socket on msg-input', widgetId, msg)
```
Steve-Mcl, line 78:
```suggestion
```

remove before merge

ui/src/widgets/ui-text-input/UITextInput.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <v-tooltip :text="tooltip" :disabled="!tooltip?.length" location="bottom">
+    <v-tooltip :key="key" :text="tooltip" :disabled="!tooltip?.length" location="bottom">
```
Steve-Mcl, line 2: not entirely sure this is required (I forget when/why I felt the need to add the key)

Steve-Mcl (commented):
for consideration

joepavitt:
@Steve-Mcl merging https://github.com/FlowFuse/node-red-dashboard/pull/1343 has triggered a merge conflict - you okay to sanity check it please?

Steve-Mcl:
> @Steve-Mcl merging #1343 has triggered a merge conflict - you okay to sanity check it please?

All done. Found and fixed an issue too so e2e tests should pass now.
Also, broke then fixed unit tests - bah!

Steve-Mcl:
so this (rather large) piece of work now has conflicts again - bah!

There are some things in this PR that fix things like breaking of widgets after being dynamically updated etc - we should get this is.  It is likely going to conflict with #1401 / #1428 and #30 / #1369

With these 3 large PRs hanging around, I am loathe to start some of the other tasks that will probably compound this further.

joepavitt:
Thanks Steve, it's a huge PR which is going to take a lot of my time to review properly. I can't promise when I can get to it as there are a lot of fundamental changes here.

joepavitt:
> Happy to do on a call tomorrow to make this swift(er)

Let's do that - mind putting something in that suits you please?

PR #1241, opened 2024-08-27
Moved "Motivation" section right above "License" section

gayanSandamal:
## Description

Moved "Motivation" section right above "License" section.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,6 @@
 
 Documentation can be found here: https://dashboard.flowfuse.com
 
-## Motivation
-
-The original [Node-RED Dashboard](https://github.com/node-red/node-red-dashboard) has served us well for many years, providing an intuitive way to create live dashboards for Node-RED flows. However, the original Dashboard is based on Angular v1, which is no longer actively maintained. We identified the need for a secure, updated, and innovative successor.
-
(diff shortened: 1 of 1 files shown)


joepavitt (approved):
Thanks Gayan - Would also be good to add a section above `Installation` that shows off some example dashbaords/visualisations.

gayanSandamal:
> Thanks Gayan - Would also be good to add a section above `Installation` that shows off some example dashbaords/visualisations.

That would be a good idea. How about a section called "Live Demo" with some screenshots of the interactive documentation and a link to the interactive dashboard charts?

joepavitt:
A link to interactive dashboard for sure, I think we could make a single Dashboard that was a bit cleaner in it's layout though and showed more diverse widgets on a single Dashboard.

gayanSandamal:
> A link to interactive dashboard for sure, I think we could make a single Dashboard that was a bit cleaner in it's layout though and showed more diverse widgets on a single Dashboard.

In that case, I will create a new page with more diverse widgets in the demo instance and move it as the first page.

PR #1246, opened 2024-08-29
UI Radio Group - server side data storing issue fix

gayanSandamal:
## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/widgets/ui_radio_group.js
+++ b/nodes/widgets/ui_radio_group.js
@@ -1,4 +1,5 @@
 const statestore = require('../store/state.js')
+const { appendTopic } = require('../utils/index.js')
 
 module.exports = function (RED) {
     function RadioGroupNode (config) {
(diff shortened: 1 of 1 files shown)


nodes/widgets/ui_radio_group.js
```diff
@@ -24,7 +24,7 @@ module.exports = function (RED) {
                     // dynamically set "columns" property
                     statestore.set(group.getBase(), node, msg, 'columns', updates.columns)
                 }
-                if (updates?.options) {
+                if (updates.options) {
```
joepavitt, line 27: why are you making this change?

gayanSandamal: It didn't throw any errors when initialising. So I removed because of that. But you're correct better to handle it. 

I will add it back

joepavitt: It's perfectly reasonable for a message to be sent without `msg._update` though, so the `?` is a guard against that and should be kept.

gayanSandamal: Reverted the removal of null checker

gayanSandamal: > It's perfectly reasonable for a message to be sent without `msg.ui_update` though, so the `?` is a guard against that and should be kept.

Understood the scenario. I added it back. Thanks for the clarification

nodes/widgets/ui_radio_group.js
```diff
@@ -12,11 +13,22 @@ module.exports = function (RED) {
 
         const evts = {
             onChange: true,
-            beforeSend: function (msg) {
-                if (msg.options) {
+            beforeSend: async function (msg) {
+                const updates = msg.ui_update
+                console.log('ui_radio_group: beforeSend', msg)
```
joepavitt, line 18:
```suggestion
```

joepavitt (changes requested):

PR #1252, opened 2024-08-31
Added opt in pointerdown and pointerup event capture on button widget and fixed button theming bug

### LINKED ISSUE #1276: Button - Not filling height > 1 with "Compact" or "Comfortable" sizings
### Current Behavior

Buttons do not fill the container when "compact" or "comfortable" is used as theme

### Expected Behavior

Buttons should fill their container everytime

### Steps To Reproduce

Create Button
Set Page theme to "compact" or "comfortable"

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### LINKED ISSUE #1277: More eventhandlers for Buttons
### Description

It would be nice for many HMI applications if we could register a button down event in order to move machinery or whatever while a button is pressed 

JulianStremel:
## Description

I made a HMI for a machine where the user wants to jog part of the machine to another position.
In order for the machine to start on pointerdown and stop on pointerup the button needed to be able to capture these events.

Because the button would only capture the release when pointerup occured on the button itself, the user could have slipped his finger ftom the button while operating and the pointerup event would not have been captured.
To prevent this i added the setPointerCapture to the pointerdown event to make releasing the button more predictable.
On touch devices "touch-action: none" should also be set in order for the browser to not act on gestures like zoom etc. which could also make the release of the button unreliable but this is left as an excercise to the user because i didn´t want to interact with global css.

## Related Issue(s)

Fixes #1277
Fixes #1276 

## Checklist

 -> i do not currently understand the cypress test framework and would need some guidance in order to implement the tests
    - [x] Upgrade instructions (Not needed)
    - [x] Configuration details
    - [x] Concepts
 - [x] Changes `flowforge.yml`?
    - NO

## Labels

 - [x] Includes a DB migration? -> NO


nodes/widgets/ui_button.js
```diff
@@ -43,7 +57,7 @@ module.exports = function (RED) {
                     }
                 }
             }
-
+
```
Steve-Mcl, line 60:
```suggestion

```

nodes/widgets/ui_button.html
```diff
@@ -151,6 +172,31 @@
         <label for="node-input-tooltip"><i class="fa fa-info-circle"></i> <span data-i18n="ui-button.label.tooltip"></label>
         <input type="text" id="node-input-tooltip" data-i18n="[placeholder]ui-button.label.optionalTooltip">
     </div>-->
+
+    <!-- ------------------------------------------------------------------------------------------------------------------------------------- -->
```
Steve-Mcl, line 176:
```suggestion
```

nodes/widgets/ui_button.html
```diff
@@ -151,6 +172,31 @@
...
+        <label for="node-input-enablePointerup"><i class="fa fa-hand-pointer-o"></i> Enable pointerup</label>
+        <input type="checkbox" id="node-input-enablePointerup">
+    </div>
+    <div class="form-row">
+        <label for="node-input-pointerupPayload" style="padding-left: 25px; margin-right: -25px"><span data-i18n="ui-button.label.payload"></label>
+        <input type="text" id="node-input-pointerupPayload" style="width:70%">
+        <input type="hidden" id="node-input-pointerupPayloadType">
+    </div>
+
+    <!-- ------------------------------------------------------------------------------------------------------------------------------------- -->
```
Steve-Mcl, line 198:
```suggestion
```

Steve-Mcl (changes requested):
Hi @JulianStremel thanks for this contribution.

Having generated many HMIs in my time I can 100% see the benefit of pointer down and pointer up events.

I am doing a preliminary review to move it along but ultimately it will be up to Joe to accept this.

there are few tidy up suggestions and some usability suggestions/questions...

1. Tidy up empty lines and empty comment blocks
1. Normally we hide the typed input if the checkbox is disabled. However this get messy.  A potetially better solution would would be to lose the checkboxes altogether and simply add "None" or "Disabled" type to the typed inputs. Also they should be the default so that existing user flows are backwards compatible.
2. I have not tested this but i _think_ you need to handle touchscreen touches differently to how mouse operations raise events (happy to be corrected here)

JulianStremel:
> Having generated many HMIs in my time I can 100% see the benefit of pointer down and pointer up events.
> 
Thanks i really think this is a key feature this node was missing :)

> there are few tidy up suggestions and some usability suggestions/questions...
> 
> 1. Tidy up empty lines and empty comment blocks
>
Thanks i commited the the changes you suggeested

> 2. Normally we hide the typed input if the checkbox is disabled. However this get messy.  A potetially better solution would would be to lose the checkboxes altogether and simply add "None" or "Disabled" type to the typed inputs. Also they should be the default so that existing user flows are backwards compatible.
>
I will have a look into that 
> 3. I have not tested this but i _think_ you need to handle touchscreen touches differently to how mouse operations raise events (happy to be corrected here)
>
As far as i am aware the [pointerevents](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) are hardware agnostic and as such should be used here.

nodes/widgets/ui_button.html
```diff
@@ -36,7 +36,26 @@
...
+                iconColor: { value: '' },
+                
+                // pointerup/down event support
+                enablePointerdown: { value: false },
+                pointerdownPayload: { value: '', validate: (hasProperty(RED.validators, 'typedInput') ? RED.validators.typedInput('pointerdownPayloadType') : function (v) { return true }) },
+                pointerdownPayloadType: { value: 'str' },
+                enablePointerup: { value: false },
+                pointerupPayload: { value: '', validate: (hasProperty(RED.validators, 'typedInput') ? RED.validators.typedInput('pointerupPayloadType') : function (v) { return true }) },
+                pointerupPayloadType: { value: 'str' },
+
```
Steve-Mcl, line 58:
```suggestion
                iconColor: { value: '' }
```

duplicate code

Steve-Mcl (changes requested):
You seem to have duplicate code in the node defaults.

Also, thanks for checking the pointer events work on both mouse and touchscreen.

Considering the action could be a mouse pointer or touch press or even keyboard spacebar, does it make sense to name these events `pointerXxx`?  Would it not make more sense to name these `buttonDown` and `buttonUp` events?

Lastly, was there a particular reason you stuck with extra checkboxes for enable disable?  Adding a `none` option to the typedInput would remove the need for all the change handlers showing hiding stuff, remove 2 extra boolean props, removes a bunch of HTML etc.

nodes/widgets/ui_button.html
JulianStremel: Oh thanks that has to have happened when merging somathing ... i am fairly unexperienced as you can see.

I just used the name because it is refering to the actual browser event nothing more nothing less but the user would propably be more used to calling it button down and up. So i will rename them.

I thought disabling the events early as possible would be the best or maybe even disabling the eventhandler inside the component in order for the system to not have to compute on something tht gets dropped when it tries to send.

joepavitt (approved):
A quite excellent addition - thanks @JulianStremel

joepavitt:
> Lastly, was there a particular reason you stuck with extra checkboxes for enable disable? Adding a none option to the typedInput would remove the need for all the change handlers showing hiding stuff, remove 2 extra boolean props, removes a bunch of HTML etc.

I quite like the UX for the checkbox as it hides fields for those that aren't using it.

Steve-Mcl (approved):

joepavitt:
The test failure is an odd one, can't see why the flow would now be broken, and not seeing any obvious errors in the flow. I need to investigate this further.

JulianStremel:
> The test failure is an odd one, can't see why the flow would now be broken, and not seeing any obvious errors in the flow. I need to investigate this further.

The onInput event in the [nodes/widgets/ui_button.js](https://github.com/FlowFuse/node-red-dashboard/pull/1252/files#diff-db2ff9d19a736161f8ea5684f142ef64dbc83477404e0c5d5632162783d3ae03) sends a msg without the _event proprty to the beforeSend function which tries to determine the event type ... now we check if the _event property is set before we go into the switch case and we set the proprty to "inject" if not present ... i know this is kind of rough but i do not know anything about async js in order to set the property inside the onInput handler

Steve-Mcl:
Hi @JulianStremel @joepavitt 

I have pulled this locally and took it out for a spin.

It is working nicely (thanks for the hard work) but I have some reservations that I think should at minimum be discussed before we lock this in and users become depnendant on its operation.


1. The On click `topic` and `payload` are grouped together in the UI suggesting they "belong together" but the `topic` for the "click" event gets sent to the other events too.   I am not necessarily saying this is a bad thing but it looks very odd.
   1. A potential solution to avoid re-jigging the UI would be to add "on down" and "on up" topic fields. 
   2. The other solution is to move topic to its own part of the form (but how that ties up is not immediately obvious)
3. The timing of the messages for `up` and `click` are IDENTICAL (to the `ms`) and TBH, feel almost like something the user will have to handle/decern whenever they want to use UP and DOWN only. 
   1. Users (me too) are bound to ask, why can't I just disable the "click" event so that I only get "down" and "up" - I think that is a valid mode of operation.
5. Keyboard operations are not supported. i.e. tab onto the button so it is highlighted, press the space bar - the "down" event is not triggered. Releasing the space bar does not fire the "up" event either (the "click" event does trigger)

image

<br>

As a side note, I have noticed the buttons do not adhere to the normal "dont send click if mouse is slid away before release".  This is super important in HMIs (e.g. operator presses the "FIRE THE LASERS" button then before releasing finger spots something wrong/dangerous and slides finger off the button to avoid the click event being issued) (ref: https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) 

For these reasons, I will hold off on merging this into main (it is close, but just want the above considered before merging and releasing)

Steve-Mcl:
> > As a side note, I have noticed the buttons do not adhere to the normal "dont send click if mouse is slid away before release". This is super important in HMIs (e.g. operator presses the "FIRE THE LASERS" button then before releasing finger spots something wrong/dangerous and slides finger off the button to avoid the click event being issued) (ref: https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event)
> 
> Your last note is the exact part why i added the pointerup event as well. I personally use this button to turn a cnc axis on when pressed and off if released which is CRUCIAL to turn off reliably even if the operator slides is finger from the button. That is why i register every release event after the pointerdown to trigger on the element which was pressed down to begin with.

Ok, so I checked with vuetify (e.g. [this demo](https://play.vuetifyjs.com/#eNqdVU1z2yAQ/StU05m0M8GJHadpM0mmnfbeQ3uLc8BoZREj0ACSnXr837sISQbbvdQHD7CP/Xj7WD3vsm91PWkbyO6zBwdVLZmDp4Ui5KGltgRwfk0Il8zax0VWayuc0IoaQKRoYZEFQCUULUGsSoeo+e11OO88oa9ctKcu2NJq2TggOS0kbAmTYqUoB+XAkNfGOlG8DdsNnV5fk9L/L7Lea5fj0qlhh1lqqQ2GyAFqWjemlkBzZtag6GxI1P+s+AMI21LJzGoswf++cin42rsQTOoVeSTvwuoAGqMT8rMGRX4E+5DSVZfTUPgVVo7rftfSguVAnWEqcEBKgXvkQgJrA+89kDOTHwK1VBRjUnG+rMYUciq4Vmh/z6W2ST0D6dWWssbp2AQSWmxhd3H6KbZUbEs3InclWm494QeTE0566n5hVMKIAQ6idmfJGeWE2Vup3X3INUIMDSRH6ZOWGcGUV5KDLXo/05aCSQ99SvkOnEc6jo5bir1Auk13Z1gnkESmb3Q6Iz5+r8FId70/n3Z8FNG9pAmjsThtwzlYe2zuOahyQXkJfE25MBz1qxsnhUqaGkl4OvucWrrivK+ksqPauqrKOSm0cnTTvVq61DLHEn+Xwg5tJRtmicXiBxkfKI5l/V/sMjoP5KIkTplNXnVMbHdF6VM+BnYryEVTUVRAyaw4YdmPqUHaX5KZ0D2zUXc965haijC6UTkeJ4f/VGfSmPTO907qcdVnhByz7u1+KIyT5WSYhLGN593cxt3RS3iw3PiuWnBN3YFFVWvjyA4bXpA9KYyuyAV+DC5C31BF1pGxKAR9cKaBj95z8PWUXWbI09pOXq1W+A3ZLTLutX3//LJHWwhAK1ZHgHDoQbv9PttfZjeT+WR2h/Cbyd1kll36IC9/AQl09RI=) and buttons DO work as expected (sliding finger off prevents the click event) and I checked with dashboard-2 v1.17.0, it is the same (sliding finger off the button inhibits click - as expected/hoped) 

I understand your desire to fire the `up` always (if a `down` was fired) - I have fought this battle too on HMI development. However, changing the default and expected behaviour of the click event is not a side effect I am personally happy to incorporate.  This has wider implications that should be discussed, understood, and ultimately documented. For example, it is not unreasonable for the DB button widget to always send an `up` after sending a `down` - it _feels_ expected (even if not the "normal HTML/forms behaviour) (and we can document that as a behaviour) however the sending of `click` events after deliberately sliding by finger off is not something I would want or expect. I would always expect sliding finger off the button NOT to send a click event.

@joepavitt what are your thoughts here?

JulianStremel:
> Ok, so I checked with vuetify (e.g. [this demo](https://play.vuetifyjs.com/#eNqdVU1z2yAQ/StU05m0M8GJHadpM0mmnfbeQ3uLc8BoZREj0ACSnXr837sISQbbvdQHD7CP/Xj7WD3vsm91PWkbyO6zBwdVLZmDp4Ui5KGltgRwfk0Il8zax0VWayuc0IoaQKRoYZEFQCUULUGsSoeo+e11OO88oa9ctKcu2NJq2TggOS0kbAmTYqUoB+XAkNfGOlG8DdsNnV5fk9L/L7Lea5fj0qlhh1lqqQ2GyAFqWjemlkBzZtag6GxI1P+s+AMI21LJzGoswf++cin42rsQTOoVeSTvwuoAGqMT8rMGRX4E+5DSVZfTUPgVVo7rftfSguVAnWEqcEBKgXvkQgJrA+89kDOTHwK1VBRjUnG+rMYUciq4Vmh/z6W2ST0D6dWWssbp2AQSWmxhd3H6KbZUbEs3InclWm494QeTE0566n5hVMKIAQ6idmfJGeWE2Vup3X3INUIMDSRH6ZOWGcGUV5KDLXo/05aCSQ99SvkOnEc6jo5bir1Auk13Z1gnkESmb3Q6Iz5+r8FId70/n3Z8FNG9pAmjsThtwzlYe2zuOahyQXkJfE25MBz1qxsnhUqaGkl4OvucWrrivK+ksqPauqrKOSm0cnTTvVq61DLHEn+Xwg5tJRtmicXiBxkfKI5l/V/sMjoP5KIkTplNXnVMbHdF6VM+BnYryEVTUVRAyaw4YdmPqUHaX5KZ0D2zUXc965haijC6UTkeJ4f/VGfSmPTO907qcdVnhByz7u1+KIyT5WSYhLGN593cxt3RS3iw3PiuWnBN3YFFVWvjyA4bXpA9KYyuyAV+DC5C31BF1pGxKAR9cKaBj95z8PWUXWbI09pOXq1W+A3ZLTLutX3//LJHWwhAK1ZHgHDoQbv9PttfZjeT+WR2h/Cbyd1kll36IC9/AQl09RI=) and buttons DO work as expected (sliding finger off prevents the click event) and I checked with dashboard-2 v1.17.0, it is the same (sliding finger off the button inhibits click - as expected/hoped)
> 
> I understand your desire to fire the `up` always (if a `down` was fired) - I have fought this battle too on HMI development. However, changing the default and expected behaviour of the click event is not a side effect I am personally happy to incorporate. This has wider implications that should be discussed, understood, and ultimately documented. For example, it is not unreasonable for the DB button widget to always send an `up` after sending a `down` - it _feels_ expected (even if not the "normal HTML/forms behaviour) (and we can document that as a behaviour) however the sending of `click` events after deliberately sliding by finger off is not something I would want or expect. I would always expect sliding finger off the button NOT to send a click event.
>

To be completely honest i had not tested the normal click behavior ...
i had simply infered that "$evt.target.setPointerCapture($evt.pointerId)" would only bind the pointer type events and not also bind the click event.

Honestly i completely share your opinion on not interferring with the standard button behavior 100%.

Maybe the functionality should be implemented as a kind of "mode switch" that clearly changes the complete button behavior to only emit on up and down events or maybe the "setPointerCapture" can be modified or replaced to only bind the pointerup event to the initial element.

JulianStremel:
[e6ddd9c](https://github.com/FlowFuse/node-red-dashboard/pull/1252/commits/e6ddd9c7f95c236c35f9f43b78a556dd0917c4cf)

Now we check if the click event occurs on the button itself making the button behave like expected ... aka not fire when the finger is moved off the button while holding it clicked.

Steve-Mcl:
@joepavitt are you happy with approving this for the CI tests to run (i.e. not PR approval, just test runners)

JulianStremel:
The last Version  17.0 seems to fail on the navigation tests even without my code so i guess there is another issue at play ? @Steve-Mcl  @joepavitt

joepavitt:
Yep, right you are @JulianStremel - those are independent from your work.

joepavitt:
This is great, thanks @JulianStremel - sorry it took so long to re-review. Happy to merge. I'll add a little more documentation about the event type being in `msg._event`, but other than that, I think we have everything we need.

I may also add in the `topic` per event type too, as I agree with Steve that it seems like the `topic` only gets added on click based on the config layout, when in fact, it persists across all events

JulianStremel:
No problem!
Thanks for the merge!
Happy to contribute!!!

joepavitt:
https://github.com/user-attachments/assets/48aa8f58-5661-4e32-87fb-20522c48259a

Quick demo that I'll be putting into the documentation, alongside incoming PR with some small changes too

Issue #1249, opened 2024-08-30
Button Group and Radio Group widget label horizontal position


jdotbdot:
### Current Behavior

These two widgets' labels are indented differently, marring the dashboard appearance if it has both 
![d55723be16bd73e55dfbed96638378d0cbf18e88](https://github.com/user-attachments/assets/1a755557-f741-4a69-b662-7e8818706d59)

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #1268, opened 2024-09-04
Radio group label alignment issue fix

gayanSandamal:
### Before
image

### After
Screenshot 2024-09-04 at 23 43 03

_Note that this doesn't need E2E tests to be written as this contains only one line css change_

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/1249

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-radio-group/UIRadioGroup.vue
+++ b/ui/src/widgets/ui-radio-group/UIRadioGroup.vue
@@ -143,3 +143,9 @@ export default {
     }
 }
 </script>
+
+<style>
+.nrdb-ui-radio-group .v-radio-group>.v-input__control > .v-label {
+    margin-inline-start: 0;
+}
+</style>
--- a/ui/src/widgets/ui-radio-group/UIRadioGroup.vue
+++ b/ui/src/widgets/ui-radio-group/UIRadioGroup.vue
@@ -145,7 +145,7 @@ export default {
 </script>
 
 <style>
-.nrdb-ui-radio-group .v-radio-group .v-label {
+.nrdb-ui-radio-group .v-radio-group .v-input__control .v-label {
     margin-inline-start: 0;
 }
 </style>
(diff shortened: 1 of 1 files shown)


ui/src/widgets/ui-radio-group/UIRadioGroup.vue
```diff
@@ -143,3 +143,9 @@ export default {
     }
 }
 </script>
+
+<style>
+.nrdb-ui-radio-group .v-radio-group>.v-input__control > .v-label {
```
Steve-Mcl, line 148: Gayan, does this need to be do specific? 

I'm worried users won't be able to style this so easily with their own CSS. 

Is there a chance the selector can be reduced to the minimum? 

It's ok if not (I am no CSS expert, this is just an observation)

gayanSandamal: Better to make more specific as the Vuetify has used the same path to apply CSS otherwise we may have to force apply the CSS with `!important`

Screenshot 2024-09-05 at 00 14 39

The best we can do is to reduce the selector path to this `.nrdb-ui-radio-group .v-radio-group .v-label`

Steve-Mcl: Thanks gayan

gayanSandamal: @Steve-Mcl it doesn't override when we skipped `v-input__control` class. We have to specify it but it's not necessary to catch next immediate child element using `>` operator

The updated selector is `.nrdb-ui-radio-group .v-radio-group .v-input__control .v-label` [here](https://github.com/FlowFuse/node-red-dashboard/pull/1268/commits/68a70a4d3382400245bdcac0b4d10882475625c1)

Steve-Mcl (approved):

Issue #1159, opened 2024-08-05
Cannot load widgets that are part of pre-installed nodes


theotherwillembotha:
### Current Behavior

Currently I package an instance of nodered as a docker image with some nodes that are installed during the build step.
The base image is `nodered/node-red:3.1.11` and the preinstalled nodes are added to `/usr/src/node-red/package.json`. The reason they are added to `/usr/src/node-red/package.json` is because `package.json` in the userDir is overwritten when the image is run as a container and the `RED.settings.userDir` (in this case `/data`) is mounted externally.

- The preinstalled nodes work correctly in the flows.
- Nodes installed through the pallet and installed via the commandline work correctly (as they are added '/data/package.json')
- Thirdparty Widgets of nodes installed at runtime work correctly.
- Thirdparty Widgets of nodes installed at build time show a  `'No Vue component found for ', widget.type, ' - falling back to ui-template'` message in the browser console.

I chased the issue down to `ui_base.js` line 90 (working off branch `main` commit id `97de583`) where the third party widgets are loaded. In addition to the `RED.settings.userDir`, it looks like you also have to scan the package.json file in `process.cwd()` for widgets.

### Expected Behavior

Thirdparty widgets of nodes installed in the main `package.json` file should work the same as packages widgets of nodes installed at runtime.

### Steps To Reproduce

1. in the root nodered folder, run `npm install ........` to install a node that contains a third party widget.
2. verify that the node was installed in the `package.json file`
3. run nodered and specify the `--userDir' to another folder.
4. verify that node with the thirdparty plugin is not installed in the `package.json` file in the --userDir folder.
5. create a dashboard with the third party widget.

### Environment

- Dashboard version: 1.14.0
- Node-RED version: 3.1.11
- Node.js version: v18.20.2
- npm version: 10.5.0
- Platform/OS: docker
- Browser: frefox 123.0

### Have you provided an initial effort estimate for this issue?

I can not provide an initial effort estimate

theotherwillembotha:
Additional Information:

The current version of our docker build script looks something like this. As mentioned previously, we build an image that contains roundabout 70 of our own plugins and about 10 public plugins.

``` dockerfile

FROM nodered/node-red:4.0.2-20
USER root:root
RUN apk update
RUN npm install -g typescript
RUN npm install -g vite
RUN chown -R 1000:1000 "/data/.npm"
USER node-red:node-red

# install the "patched" node-red-dashboard.
[...]

PR #1274, opened 2024-09-06
load preinstalled 3rd party widgets

theotherwillembotha:
## Description
We build a docker image based on the official nodered docker image and and bundle it with a couple of our own plugins and widgets. The current 3rd party widget loader code only loads widgets from the package.json file in the user settings folder and not from the package.json file in the nodered root /base where preinstalled nodered plugins reside. This changeset allows widgets to be installed from both locations in addition to loading plugins relative to a node_modules folder or package.json file.

## Related Issue(s)
[Issue 1159](https://github.com/FlowFuse/node-red-dashboard/issues/1159)

## Checklist

> Testing requires loading 3rd party plugins at runtime which the current testing framework doesn't support. 
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
> No documentation changes needed
> No changes to `flowforge.yml`

## Labels

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -88,38 +88,66 @@ module.exports = function (RED) {
             }
 
             /**
-             * Load in third party widgets
-             */
-            let packagePath, packageJson
-            if (RED.settings?.userDir) {
-                packagePath = path.join(RED.settings.userDir, 'package.json')
-                packageJson = JSON.parse(fs.readFileSync(packagePath, 'utf8'))
-            } else {
-                node.log('Cannot import third party widgets. No access to Node-RED package.json')
-            }
+            * Load in third party widgets
+            */
 
-            if (packageJson && packageJson.dependencies) {
-                Object.entries(packageJson.dependencies)?.filter(([packageName, _packageVersion]) => {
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
hi @theotherwillembotha 

Thank you very much for this contribution. It is very much appreciated.

You may notice that we have raised a new PR that supersedes this: https://github.com/FlowFuse/node-red-dashboard/pull/1286

Your contribution was fine however I felt it important that unit tests were present with this PR and that meant refactoring. So to avoid a lot of too-and-fro, I ended up just raising a new PR. I hope you dont mind.

FYI, there were a couple of issues that would have needed to be fixed before merging as it stood:
1. linting errors (use of semi-colons, spacing, etc) were present (we use eslint in our GH integrations and local env)
2. Use of `process.cwd()` to locate main package would not be suitable for all types of node-red installations (i.e. launching it from a different directory)
3. Did not support `nodesDir` 
   4. Since we are touching this code, we might as well support loading nodes that Node-RED loads via `nodesDir`
5. No unit tests present - this one felt important in this area (so as to avoid future regressions)

I hope this doesnt put you off contributing further features 🤞 

thanks again, Steve.

PR #1300, opened 2024-09-17
Enable configuration of the responsive breakpoints at the page-level

### LINKED ISSUE #1225: Responsiveness: Config to control responsive column counts
### Description

Currently the columns rendered varies with screen width as described in https://dashboard.flowfuse.com/layouts/types/grid.html and the equivalent for Fixed layout.
These can be overridden by CSS but it would be convenient to have this configurable in the ui-page setup.

I am not a flowfuse team member

joepavitt:
## Description

- Adds a new configuration option `breakpoints` at the `ui-page` configuration-level.
- Permits users to define the varying breakpoints, both _when (in px) the breakpoint is active, and how many columns that breakpoint renders.
- Not available for "Fixed" layout as that doesn't use the concept of columns in this way, until it reaches mobile, where we force 3 columns anyway. This can be improved in the future.

### Follow Ups

It would be beneficial to introduce a "Copy to Pages" button inline with the "Responsiveness" header whereby users can copy a configuration to other pages in their Dashboard. This will be opened as a separate issue, as doesn't have to be in for an MVP.

## Related Issue(s)

Closes #1225


ui/src/mixins/responsiveness.js
```diff
@@ -0,0 +1,39 @@
...
+                const breakpoints = b.sort((a, b) => a.px - b.px)
+
+                breakpoints.forEach((bp) => {
+                    if (window.innerWidth >= bp.px) {
+                        cols = bp.cols
+                    }
+                })
+            }
+            this.columns = cols
+            console.log('columns', this.columns)
```
Steve-Mcl, line 36:
```suggestion
```

ui/src/layouts/Grid.vue
```diff
@@ -48,6 +49,12 @@ export default {
         BaselineLayout,
         WidgetGroup
     },
+    mixins: [Responsiveness],
+    data () {
+        return {
+            columns: 0
```
Steve-Mcl, line 55: should this still be here since columns now comes from the mixin?

ui/src/layouts/Grid.vue
```diff
@@ -113,7 +120,7 @@ export default {
     --layout-gap: 12px;
 }
 .nrdb-layout--grid {
-    --layout-columns: 12;
+    --layout-columns: v-bind(columns);
```
Steve-Mcl, line 123: should this have a fallback?

ui/src/layouts/Notebook.vue
```diff
@@ -122,7 +125,7 @@ export default {
     min-height: 100%;
     flex-wrap: wrap;
     padding: var(--page-padding);
-    --layout-columns: 12;
+    --layout-columns: v-bind(columns);
```
Steve-Mcl, line 128: should this have a fallback?

ui/src/layouts/Tabs.vue
```diff
@@ -118,30 +121,11 @@ export default {
 .nrdb-layout--tabs {
     --layout-gap: 12px;
     --widget-row-height: 48px;
-    --layout-columns: 12;
+    --layout-columns: v-bind(columns);
```
Steve-Mcl, line 124: should this have a fallback?

joepavitt:
Feedback resolved with 9785d3bb and 47c75076

Steve-Mcl:
In our node js files we declare `function c_ (name) { ... }` at global scope. This means the last node loaded remains the active `c_` function and as such, the labels are being loaded from a different node:

image


for now, I will move the `c_` function into the `oneditprepare` (to reduce its scope to the closure) but if we need `c_` in other parts of the node, we should wrap the whole thin in an IIFE

Steve-Mcl:
Small issue with breakpoints not updating when switching pages:

chrome_yNSE0J8X3c


Looking into it

nodes/config/ui_page.html
```diff
@@ -1,107 +1,256 @@
...
+                // breakpoints
+                const breakpoints = $('#node-config-list-breakpoints').editableList('items')
+    
+                node.breakpoints = []
+                breakpoints.each(function (i) {
+                    const row = $(this)
+                    const breakpoint = {
+                        name: row.find('.node-input-breakpoint-name').val(),
+                        px: parseFloat(row.find('.node-input-breakpoint-px').val()),
+                        cols: parseFloat(row.find('.node-input-breakpoint-cols').val())
```
Steve-Mcl, line 241:
```suggestion
                        px: row.find('.node-input-breakpoint-px').val(),
                        cols: row.find('.node-input-breakpoint-cols').val()
```

`parseFloat` will try to evaluate env var overrrides like `${MY_BRK_POINT}` and will result in `NaN` being saved as the value.

In short, sanitisation and defaults should probably be made server side (if at all)

Steve-Mcl (changes requested):
1 issue

nodes/config/ui_page.html
Steve-Mcl: exec decision: removing parseFloat in this 1st iteration

Steve-Mcl (approved):

PR #1303, opened 2024-09-18
Update Node-RED to force dependency update

hardillb:
## Description

Updates Node-RED to force express update to clear a bunch of CVEs


hardillb:
I've re-run the tests a few times, number of failures changes, but looks to be on consistent failure, but I can't see how this can be down to Node-RED or Express  version

joepavitt:
Yep, that's our normal control problem - not on you

joepavitt (approved):

Issue #1266, opened 2024-09-04
Switching gauge types clears users entered values


Steve-Mcl:
### OP

> 3. Next I added a gauge, and spent a while setting up the "segments" (these are the different colours for different thresholds of the gauge). Also I set up the range for the gauge. I enjoyed spending time choosing the segment colours and thresholds. Unfortunately when I deployed, the segment colours just didn't show at first. Then I changed another field and re-deployed, and suddenly they showed. Then the problem happened again, so I changed the gauge type, and immediately lost the threshold colours, threshold levels, and range, all of which I had carefully set and planned to duplicate later 

### Response

> When switching gauge types, we have different defaults, it had been built to switch to the relevant defaults when changing gauge type. A better solution here would be for us to check your values vs. Our known defaults, if it doesnt match anything, keep your values...

See https://discourse.nodered.org/t/latest-dashboard-2-0-release-tabs-number-input-new-gauges/90506/28?u=steve-mcl for 

### Expected Behavior

If user has entered values, dont discard them

### Steps To Reproduce

enter your own colour settings, switch gauge type

### Environment

- Dashboard version: current
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS: all
- Browser: all

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

gayanSandamal:
@joepavitt this seems it has been already fixed with https://github.com/FlowFuse/node-red-dashboard/pull/1264

```
[...]

PR #1309, opened 2024-09-19
Preserve Range and Segments on Field Value Changes + Added Reset to Default Button

gayanSandamal:
## Description

This PR introduces two key changes to enhance the user experience when interacting with range and segment fields:

#### 1. Preserve Field Values:
The range (min / max) and segment (color / from) values will now be preserved when the user modifies the field inputs (`min`, `max`, `segment color`,  `segment from`, `add new segment`, `remove an existing segment`).
Previously, the values of these fields got reset to defaults on gauge change.

#### 2. Reset to Default Button:
A "Defaults" button has been added to allow users to quickly revert any changes made to the range and segment fields back to their original default values based on the selected gauge type.
Screenshot 2024-09-20 at 10 53 42

_This doesn't need e2e test update but node-red built-in help updated_

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/1266

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/widgets/locales/en-US/ui_gauge.html
+++ b/nodes/widgets/locales/en-US/ui_gauge.html
@@ -12,6 +12,8 @@
         <dd>Collection of objects that define the relevant color for a given value</dd>
         <dt>Label <span class="property-type">list</span></dt>
         <dd>The title text shown above the gauge</dd>
+        <dt>Defaults <span class="property-type">list</span></dt>
+        <dd>Reset to default min/max and segments based on the selected gauge type</dd>
     </dl>
     <h3>Properties (Half &amp; 3/4 Gauges Only)</h3>
     <dl class="message-properties">
--- a/nodes/widgets/locales/en-US/ui_gauge.json
+++ b/nodes/widgets/locales/en-US/ui_gauge.json
@@ -28,7 +28,8 @@
             "styling": "Styling",
             "class": "Class",
             "units": "Units",
-            "icon": "Icon"
+            "icon": "Icon",
+            "defaults": "Defaults"
         },
         "errors": {
             "unique": "All 'from' values must be unique."
--- a/nodes/widgets/ui_gauge.html
+++ b/nodes/widgets/ui_gauge.html
@@ -143,20 +143,11 @@
                 prefix: { value: '' },
                 suffix: { value: '' },
                 segments: {
-                    value: [{
-                        color: '#5CD65C',
-                        from: 0
-                    }, {
-                        color: '#FFC800',
-                        from: 4
-                    }, {
-                        color: '#EA5353',
-                        from: 7
-                    }],
+                    value: [...DEFAULTS['gauge-half'].segments],
                     validate: validateGaugeSegments
                 },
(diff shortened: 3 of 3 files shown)


Steve-Mcl:
@gayanSandamal as discussed:

The default segments should never be reset on any gauge type (not just half & 3/4).  


The OP stated:

>  Then the problem happened again, so I changed the gauge type, and immediately lost the threshold colours, threshold levels, and range, all of which I had carefully set and planned to duplicate later.

If necessary, we could add a "Reset to default" button under the table that applies the defaults (I would personnally prefer to that but it I concede it is a _little bit_ out of scope - so will leave it up to you to decide if it makes sence to add that in this iteration) but the most important thing is not to lose the users hard work and time by deleting their thresholds that they have painstakingly entered.

gayanSandamal:
> @gayanSandamal as discussed:
> 
> The default segments should never be reset on any gauge type (not just half & 3/4).
> 
> The OP stated:
> 
> > Then the problem happened again, so I changed the gauge type, and immediately lost the threshold colours, threshold levels, and range, all of which I had carefully set and planned to duplicate later.
> 
> If necessary, we could add a "Reset to default" button under the table that applies the defaults (I would personnally prefer to that but it I concede it is a _little bit_ out of scope - so will leave it up to you to decide if it makes sence to add that in this iteration) but the most important thing is not to lose the users hard work and time by deleting their thresholds that they have painstakingly entered.

@Steve-Mcl PR title and description of this PR is updated with the above mentioned fixes including the reset to default button

Steve-Mcl (commented):
Apart from `isDirty` flag polluting global scope, it will not work as you expect. 

for example, if there are more than one gauge, setting or resetting it in one node affects the other.

chrome_3Iuvbe5swO

Steve-Mcl:
Gayan, I have pushed a refactoring to get this over the line.

* Wraps the code in an IIFE (because there is global code in the script tag)
* Extracts the defaults to a constant `DEFAULTS` (for use in setup and comparison)
* Adds a helper function `isDefault(type)` to aid determining if the UI matches the `constant` default value
* Removes a lot of duplicate boilerplate code for setting up the min/max/segments


Here is a working demo:

chrome_JyG4oxn1YW

Steve-Mcl:
@gayanSandamal please check my work and let me know if you are happy with the changes

Steve-Mcl (approved):
Approved - but want an overcheck from @gayanSandamal on my refactoring

Steve-Mcl:
e2e fails are the typical usual `widgets/control.spec.js` issue (i.e. no regressions created) 👍

gayanSandamal:
@Steve-Mcl  Thank you for the refactoring. I will check the fixes and let you know

gayanSandamal:
> @gayanSandamal please check my work and let me know if you are happy with the changes
> Approved - but want an overcheck from @gayanSandamal on my refactoring

Thanks, @Steve-Mcl ! I really appreciate the refactoring, specially simplifying the defaults and adding the `isDefault` helper. The cleaner structure looks great.

Issue #1284, opened 2024-09-10
Inconsistent use of msg.ui_update.title to update the label for gauge node, msg.ui_update.label is used for other nodes.


colinl:
### Current Behavior

The [docs for dynamic properties of the gauge node](https://dashboard.flowfuse.com/nodes/widgets/ui-gauge.html#dynamic-properties) show `msg.ui_update.title` being used to update the label.  Other nodes I have looked at all use `msg.ui_update.label`. 

### Expected Behavior

There should be consistency across nodes.

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

joepavitt:
This was inherited from Dashboard 1.0, but we should support _both_ `title` and `label` here.

PR #1310, opened 2024-09-19
Standardise the property label instead title

gayanSandamal:
## Description

@joepavitt  @Steve-Mcl 
As this [related issue](https://github.com/FlowFuse/node-red-dashboard/issues/1284) is only mentioned about the docs update but as per my understanding it should update all the areas of gauge node related with `title` but this will lead to a problem with the existing gauges that currently use `title` property in NR flows.

Screenshot 2024-09-19 at 15 37 52

Screenshot 2024-09-19 at 15 47 07

Therefore the dashboard users have to update their flows otherwise I should limit this PR just for a one line docs update

Screenshot 2024-09-19 at 15 47 31

Screenshot 2024-09-19 at 15 47 35

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/1284

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
The only request was to support `ui_update.label` in the dynamic properties functionality, and for thag to then, under the covers, update the `title` property.

The `title` field doesn't need to change, this should require no changes from the user.

joepavitt:
The other option here is that we keep these changes, do move forward with `label`, but support injection of `config.label` if a `config.title` is present in the `js` file when the node is registered. This will prevent the need for users to manually update.

@Steve-Mcl can talk you through that process if required.

Steve-Mcl:
> The `title` field doesn't need to change, this should require no changes from the user.

This is the preferred direction IMO.  It will be seamless to the user. We will continue to use `title` in the flows and continue to support `ui_update.title` (but depreciate it) and support `ui_update.label` going forwards.  Under the hood, `ui_update.title` and `ui_update.label` should be stored in the datastore as the same prop (I propose that prop is `label`). I also propose we do not mention `title` in the docs (even though it will continue to work as before)

Gayan and I have discussed this. If I get time today i will make the changes to this branch/PR otherwise, Gayan is to pick it up in the morning.

@gayanSandamal let me know is anything needs further clarification please.  Ta.

Steve-Mcl (approved):

Steve-Mcl:
Approved. Merged only after tests pass

Steve-Mcl:
looks good - only the usual `✖  widgets/control.spec.js   ` fail

PR #1286, opened 2024-09-11
Loading 3rd party widgets

Steve-Mcl:
## Description

* Refactors 3rd party widgets loading to support locations that Node-RED already supports
  * main `package.json`
  * user `package.json`
  * `userNodes` locations
* Adds unit tests and fixtures

### Tests 
REF: https://github.com/FlowFuse/node-red-dashboard/actions/runs/10814674509/job/30001697116?pr=1286#step:6:17
```
  utils
    getThirdPartyWidgets
      ✔ should load single node package
      ✔ should load nodes from a package dependencies
```

#### Src showing code finds 3 packages in different locations
image

#### Actual running 
image

This PR supersedes https://github.com/FlowFuse/node-red-dashboard/pull/1274 

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #1316, opened 2024-09-21
Load types installed at runtime

### LINKED ISSUE #938: A node red restart is necessary after installing third party ui nodes
### Current Behavior

If a non-core node such as @flowfuse/node-red-dashboard-2-ui-led is installed then it is necessary to restart node-red in order for the widget to appear on the dashboard. Before the restart the widget appears just as an empty space on the page. It seems that this is also the case for ui-frame and ui-gauge-classic.

### Expected Behavior

Either a restart should not be required, or the user should be informed that a restart is required.

### Steps To Reproduce

_No response_

### Environment

- Dashboard version: 1.11.1
- Node-RED version: 4.0.0 beta 4
- Node.js version: 20.5.1
- npm version:
- Platform/OS:
- Browser: Edge

Steve-Mcl:
closes #938

## Description

Load widget type info upon `type-registered` event to remove the need to restart node-red after a contrib node is installed at runtime.

### Demo

https://github.com/user-attachments/assets/726907c6-c597-4de0-ad31-f3740b3f97b0

## Related Issue(s)
#938

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -91,6 +91,8 @@ module.exports = function (RED) {
              */
 
             uiShared.contribs = loadContribs(node)
+            node.onTypeRegistered = onTypeRegistered.bind(null, node)
+            RED.events.on('type-registered', node.onTypeRegistered)
 
             /**
              * Configure Web Server to handle UI traffic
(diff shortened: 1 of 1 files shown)


nodes/config/ui_base.js
```diff
@@ -188,6 +190,20 @@ module.exports = function (RED) {
         }
     }
 
+    function onTypeRegistered (node, type) {
+        // reload nodes from user directory package.json
+        if (RED.settings?.userDir) {
+            try {
+                const contribs = getThirdPartyWidgets(RED.settings.userDir)
```
joepavitt, line 197: Does this not restrict such that https://github.com/FlowFuse/node-red-dashboard/pull/1286/files would not work anymore?

joepavitt (approved):

Issue #1236, opened 2024-08-25
Switch node improvement


mugginsjm:
### Description

I think the switch node, especially with multiple entries (in my case a 3x3 layout for my phone) would look better with the label to the right and left justified.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
The feature request here is very valid - we should offer a "Layout" option (similar to the "Text" node) which allows users to choose the text-alignment with respect to the switch itself.

joepavitt:
To clarify, the layout options would be:

- `switch label       `
- `label switch       `
- `switch        label`
- `label        switch`

joepavitt:
We'd want something similar to this, which is available in `ui-text`'s configuration editor:

![Image](https://github.com/user-attachments/assets/ea04378c-835e-4c81-af30-f18b462cdfc3)

PR #1320, opened 2024-09-23
Layout switching configuration with dynamic property support

gayanSandamal:
## Description

This feature includes,

1. Switching between 5 different layouts `row-left`,  `row-left-swapped`, `row-spread`,  `row-spread-swapped` which is similar with `ui-text` node
2. Dynamic property support for the newly added layout property
3. Docs updated for the newly added property `Layout` and marked `dynamic: true` in the docs for the existing dynamic properties which haven't marked previously
4. Necessary E2E tests are added

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/1236

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/widgets/ui_switch.html
+++ b/nodes/widgets/ui_switch.html
@@ -130,8 +130,10 @@
                     }
                 })
 
-                $('.nr-db-switch-layout-' + (this.layout || 'row-spread')).addClass('selected');
-                ['.nr-db-switch-layout-row-left', '.nr-db-switch-layout-row-left-swapped', '.nr-db-switch-layout-row-spread', '.nr-db-switch-layout-row-spread-swapped']
+                // select default option
+                $('#nr-db-switch-layout-' + (this.layout || 'row-spread')).addClass('selected');
+                // add click handlers
+                ['#nr-db-switch-layout-row-left', '#nr-db-switch-layout-row-left-swapped', '#nr-db-switch-layout-row-spread', '#nr-db-switch-layout-row-spread-swapped']
                     .forEach(function (id) {
                         $(id).click(function (e) {
                             $('.nr-db-switch-layout').removeClass('selected')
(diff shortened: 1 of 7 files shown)


gayanSandamal:
> Ready for review @gayanSandamal?

Yes, I was waiting for tests to pass. I'm marking this as ready for review

gayanSandamal:
The PR is updated. only the usual `❌  widgets/control.spec.js` fails

nodes/widgets/ui_switch.html
```diff
@@ -194,6 +206,37 @@
...
+                    <span class="nr-db-switch-layout-label">label</span>
+                    <span class="nr-db-switch-layout-value">switch</span>
+                    <div class="nr-db-switch-layout-checkbox"></div>
+                </a>
+                <a href="#" class="nr-db-switch-layout nr-db-switch-layout-row nr-db-switch-layout-row-spread-swapped">
+                    <span class="nr-db-switch-layout-value">switch</span>
+                    <span class="nr-db-switch-layout-label">label</span>
+                    <div class="nr-db-switch-layout-checkbox"></div>
+                </a>
+            </div>
```
joepavitt, line 237: I don't understand _why_, but for some reason, even though this code is correct, it renders in the Node-RED Editor with the last option reversed in it's rendering:

Screenshot 2024-09-24 at 16 46 32

So there are two options that _look_ the same. The HTML rendered to the DOM is as rendered, but not as defined in the underlying HTMl file. It's incredibly strange.

Functionally, the PR is great. Docs are good, but this is a strange one

joepavitt: I can see the problem, CSS issues. Fixing  myself now, and will use `display: flex` as it's cleaner

gayanSandamal: > I can see the problem, CSS issues. Fixing myself now, and will use `display: flex` as it's cleaner

Thanks fixing this Joe

joepavitt (approved):

Issue #1322, opened 2024-09-23
Padding of <v-text-field> input field is overlapping with label


liulysk:
### Current Behavior

Version v1.17. Input field of <v-text-field> is overlapping label and additional css is needed to pad it. 
![image](https://github.com/user-attachments/assets/a932f6c4-7679-41e9-bca4-24531d7dd139)

Last tested correct version of dashboard which didn't have this issue: v1.13

### Expected Behavior

Label and input not overlapping.
![image](https://github.com/user-attachments/assets/6ce5cbad-abdb-41e8-a4a8-d3f1a722e9c5)

### Environment

- Dashboard version:1.17
- Node-RED version:4.0.2
- Node.js version:20
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

Issue #1201, opened 2024-08-13
Texts getting cropped on multi-line mode in UI Text Input


gayanSandamal:
### Current Behavior

Texts are getting cropped in the UI Text Input widget on multi-line (textarea) mode

<img width="616" alt="Screenshot 2024-08-13 at 13 41 33" src="https://github.com/user-attachments/assets/6d37d7ce-24ce-4c53-bfcf-57ee61f548ed">

### Expected Behavior

Textual contents should be visible without being cropped

### Steps To Reproduce

- Add a UI Text Input node in the NR dashboard 2.0
- Change the mode from text to textarea
- Add some textual contents and check the dashboard widget

<img width="343" alt="Screenshot 2024-08-13 at 13 33 23" src="https://github.com/user-attachments/assets/a348904a-b3a2-4431-b2b7-79c433598ab0">

### Environment

- Dashboard version: 2.0
- Node-RED version: v3.1.10
- Node.js version: v20.11.1
- npm version: 10.2.4
- Platform/OS: MacOS Sanoma (14.5)
- Browser: Chrome / Edge / Brave / Firefox / Safari

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Issue #989, opened 2024-06-19
ui-text-input, multiline text area does not fill available space.


Steve-Mcl:
### Current Behavior

text area does not fill available space.

![chrome_QI22j5GcSK](https://github.com/FlowFuse/node-red-dashboard/assets/44235289/26c25e8d-20ec-4d32-bef7-ba36001aa091)

### Expected Behavior

multi-line text box should fill space

### Steps To Reproduce

add a ui-text-input and set the size height greater than 4

### Environment

- Dashboard version: 1.11.1
- Node-RED version: 3.19
- Node.js version:
- npm version:
- Platform/OS: Windows
- Browser: Chromw

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

hotNipi:
CSS override to make it as it should:

```
.nrdb-ui-text-input .v-input--horizontal {
    grid-template-rows: auto 0;
}

.nrdb-ui-text-input textarea {
    height:100%;
    resize: none;
}
```
It is not solution to use directly as fix. The component has gridarea for messages, If it will be used for something - there's 0 pixels for it to live.

PR #1323, opened 2024-09-24
Fix: Empty space filling for multiline text input 

gayanSandamal:
## Description

https://github.com/user-attachments/assets/9f1ad555-ac34-41c5-818c-e4760602082e

_This fix doesn't need any documentation or E2E test updates_

## Related Issue(s)

Closes https://github.com/FlowFuse/node-red-dashboard/issues/989
Closes https://github.com/FlowFuse/node-red-dashboard/issues/1201
Closes https://github.com/FlowFuse/node-red-dashboard/issues/1322

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-text-input/UITextInput.vue
+++ b/ui/src/widgets/ui-text-input/UITextInput.vue
@@ -1,10 +1,10 @@
 <template>
     <v-tooltip :text="tooltip" :disabled="!tooltip?.length" location="bottom">
         <!-- eslint-disable-next-line vue/no-template-shadow -->
-        <template #activator="{ props }">
+        <template #activator="{ props: activatorProps }">
             <v-text-field
                 v-if="type !== 'textarea'" v-model="value"
-                v-bind="props"
+                v-bind="activatorProps"
                 :disabled="!state.enabled" class="nrdb-ui-text-field"
                 :label="label" :type="type" :rules="validation" :clearable="clearable" variant="outlined" hide-details="auto"
                 :prepend-icon="prependIcon" :append-icon="appendIcon" :append-inner-icon="appendInnerIcon" :prepend-inner-icon="prependInnerIcon" @update:model-value="onChange"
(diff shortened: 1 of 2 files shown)


ui/src/widgets/ui-text-input/UITextInput.vue
```diff
@@ -1,19 +1,19 @@
 <template>
     <v-tooltip :text="tooltip" :disabled="!tooltip?.length" location="bottom">
         <!-- eslint-disable-next-line vue/no-template-shadow -->
-        <template #activator="{ props }">
+        <template #activator="{ props: activatorProps }">
```
joepavitt, line 4: `activatorProps` isn't a variable here?

joepavitt (changes requested):
Screenshot 2024-09-24 at 10 59 23

I actually think this issue is caused by this CSS I injected into `common.css` a couple of months ago.

I think a reversal of that `padding-top`/`padding-bottom` is in fact the better option here. I say that because https://github.com/FlowFuse/node-red-dashboard/issues/1322 has also been opened which shows it's a problem even when using `v-text-field` inside a `ui-template`, so the fix in this PR wouldn't cover that.

ui/src/widgets/ui-text-input/UITextInput.vue
gayanSandamal: I've to set an alias for the `template activator's props` because I'm using `props.height` coming from the parent component to set the rows and it's getting override by the `template activator's props` 
Screenshot 2024-09-24 at 15 54 57

gayanSandamal:
> Screenshot 2024-09-24 at 10 59 23
> I actually think this issue is caused by this CSS I injected into `common.css` a couple of months ago.
> 
> I think a reversal of that `padding-top`/`padding-bottom` is in fact the better option here. I say that because #1322 has also been opened which shows it's a problem even when using `v-text-field` inside a `ui-template`, so the fix in this PR wouldn't cover that.

This is fixed with the commit https://github.com/FlowFuse/node-red-dashboard/pull/1323/commits/5dee981189d86baab1f75a0064b0b8701a87a21e and it resolves https://github.com/FlowFuse/node-red-dashboard/issues/1322 and https://github.com/FlowFuse/node-red-dashboard/issues/1201

joepavitt (approved):
Confirms this works, and also closes #1322

Issue #1254, opened 2024-09-01
Hover to show all topics in x property value when hovering


E1cid:
### Description

With multi line chart if x property has multiple y values when you hover only one is shown, this differs to dashboard one behaviour. 
Would be nice to see all y topics  with the same x property value
![Screenshot_20240901-101351](https://github.com/user-attachments/assets/9714454e-69b9-4e63-a99b-42bc5ccec5e7)

### Have you provided an initial effort estimate for this issue?

I am no FlowFuse team member

joepavitt:
Have found the ChartJS setting for this: https://www.chartjs.org/docs/latest/samples/tooltip/interactions.html

But it doesn't compare by `x` value, it compares by `index` in the dataset, so if the data is in a different order, it's not matching up correctly.

Haven't spent more than 10 mins on this, but want ed to report it here as it's a feature we should support sooner rather than later.

PR #1331, opened 2024-09-25
Charts e2e tests (mvp)

Steve-Mcl:
## Description

Adds end to end tests for charts.  This is in preparation for larger charts PR (e.g. https://github.com/FlowFuse/node-red-dashboard/pull/1325)

* Adds a mixed 3 chart test that uses a version of the example flows
* Adds a stacked bar test
* Adds a line chart limited to 3 points test

the intention is to get tests in place and merged, then merge main into #1325 to minimise the chances of any regressions.

## Related Issue(s)

#1325

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #1336, opened 2024-09-26
Group tooltips for line chart

Steve-Mcl:
## Description

Groups tooltips on `line` type charts using the [interactions](https://www.chartjs.org/docs/latest/samples/tooltip/interactions.html) options 

NOTES:
* I wrapped the call to `this.chart.update` into a `chartUpdate` method so that we can debounce the multiple calls generated by property updates (the props are `watch`ed individually) 
  * I then noted we dont actually support dynamic prop updates 🤦‍♂️ BUT I left it in as dynamic updates are listed as a TODO and this will avoid unnecessary chart update calls.
* I added some debugging where I add the `chart` object to the `window` object so that during debugging, I can see the actual config and data applied to the chart. This was super useful. So instead of deleting it, I let it in commented with a note for future devs to take advantage of.  fee free to remove it.

## Related Issue(s)

#1254

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/cypress/tests/widgets/chart-options.spec.js
+++ b/cypress/tests/widgets/chart-options.spec.js
@@ -96,15 +96,15 @@ describe('Node/-RED Dashboard 2.0 - Chart Widget - chart options', () => {
         })
     })
 
-    it('sets tooltip `mode` to "index" for donut charts', () => {
-        const chartName = 'donut-chart-1'
+    it('sets tooltip `mode` to "index" for doughnut charts', () => {
+        const chartName = 'doughnut-chart-1'
         const overrides = [
             {
                 id: 'chart-1', // the ID in the base flow fixture (required)
                 mode: 'merge',
                 data: {
                     label: chartName, // set the label
-                    chartType: 'donut' // change the chart type
+                    chartType: 'doughnut' // change the chart type
                 }
             }
         ]
@@ -120,7 +120,7 @@ describe('Node/-RED Dashboard 2.0 - Chart Widget - chart options', () => {
             should(win.uiCharts).is.not.empty()
             const chartObject = win.uiCharts['chart-1']
             should(chartObject).is.not.empty()
-            should(chartObject.chart.options.type).be.equal('donut')
+            should(chartObject.chart.options.type).be.equal('doughnut')
             should(chartObject.chart.options.interaction).be.an.Object()
             should(chartObject.chart.options.interaction.mode).be.equal('index')
             should(chartObject.chart.options.interaction).not.have.property('axis')
--- a/cypress/tests/widgets/chart-options.spec.js
+++ b/cypress/tests/widgets/chart-options.spec.js
@@ -1,0 +1,121 @@
+/// <reference types="cypress" />
+import should from 'should'
+
+/* eslint-disable cypress/no-unnecessary-waiting */
+describe('Node/-RED Dashboard 2.0 - Chart Widget - chart options', () => {
+    it('resets tooltip mode to "nearest" for bar charts', () => {
+        const chartName = 'bar-chart-1'
+        const overrides = [
+            {
+                id: 'chart-1', // the ID in the base flow fixture (required)
+                mode: 'merge',
+                data: {
+                    label: chartName, // set the label
+                    chartType: 'bar' // change the chart type to bar
+                }
+            }
+        ]
+        cy.deployFixture('dashboard-chart-options', overrides)
+        cy.visit('/dashboard/page1')
+        cy.reload()
+
+        cy.get('#nrdb-ui-widget-chart-1 > div > canvas').should('exist')
+        cy.clickAndWait(cy.get('button').contains('random'), 10)
+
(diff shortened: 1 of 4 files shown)


Steve-Mcl:
NOTE: I will add tests for this once #1331 is merged.

cypress/tests/widgets/chart-options.spec.js
```diff
@@ -0,0 +1,121 @@
...
+        cy.deployFixture('dashboard-chart-options', overrides)
+        cy.visit('/dashboard/page1')
+        cy.reload()
+
+        cy.get('#nrdb-ui-widget-chart-1 > div > canvas').should('exist')
+        cy.clickAndWait(cy.get('button').contains('random'), 10)
+
+        // eslint-disable-next-line promise/catch-or-return, promise/always-return
+        cy.window().then(win => {
+            should(win.uiCharts).is.not.empty()
```
joepavitt, line 57: If `win.uiCharts` is already available, then why also the need for `window.uiChart` which is commented out?

ui/src/widgets/ui-chart/UIChart.vue
```diff
@@ -191,10 +194,52 @@ export default {
...
+        updateInteraction () {
+            switch (this.chart.config.type) {
+            case 'line':
+                this.chart.options.interaction.axis = 'x'
+                this.chart.options.interaction.mode = 'index'
+                break
+            default:
+                delete this.chart.options.interaction.axis
+                this.chart.options.interaction.mode = 'nearest' // default
+                break
```
joepavitt, line 218: I mentioned [here](https://github.com/FlowFuse/node-red-dashboard/issues/1254#issuecomment-2371774649) that `mode` should be `x` for the line charts, as if you have different lines with a different number of data points, then the tooltips get mis-aligned.

```suggestion
            case 'line':
                this.chart.options.interaction.mode = 'x'
                break
            case 'scatter':
                this.chart.options.interaction.axis = 'x'
                this.chart.options.interaction.mode = 'nearest'
                break
            default:
                delete this.chart.options.interaction.axis
                this.chart.options.interaction.mode = 'index' // default
                break
```

Having played around too, the bar and donut charts benefit from `index` type, with the scatter reverting to the `nearest` tooltip type.

cypress/tests/widgets/chart-options.spec.js
Steve-Mcl: Not sure I understand the question Joe.

Where is window.uiChart commented out?

For reference: `win.uiCharts` is added explicitly in `UIChart.vue` when `Cypress` is detected (only). 
This is pretty much the only way of getting to the end for E2E tests. It was debated and researched with Serban at the time of impleneting the first round of e2e tests for charts.

ui/src/widgets/ui-chart/UIChart.vue
Steve-Mcl: > I mentioned https://github.com/FlowFuse/node-red-dashboard/issues/1254#issuecomment-2371774649 that mode should be x for the line charts


I misread the API docs https://www.chartjs.org/docs/latest/configuration/interactions.html#modes (didnt scroll down enough to see there was a mode 'x' and thought what you entered was a typo. My bad!)

cypress/tests/widgets/chart-options.spec.js
Steve-Mcl: > window.uiChart which is commented out?

ah, you mean in the other file - in UIChart.vue:
https://github.com/FlowFuse/node-red-dashboard/pull/1336/files/6040f1ca6c988150cb901e0d1292552e7e925614

Those commented out lines were super helpful during development of tests (so I could see what was set in the chart  object).

the line you have commented on here is related directly to existing code here
https://github.com/FlowFuse/node-red-dashboard/pull/1336/files/6040f1ca6c988150cb901e0d1292552e7e925614#diff-0ab94e22f1954c12fd38be2754bb91b4c72e3ad50b23742c57a12ff9c3e9e194R60-R65

As mentioned above, `uiChart` is conditionally added to the window object ONLY when cypress is detected. This highlighted code (in the test runner) picks up the chart object to get as close to the end as possible.

Steve-Mcl:
Merged your correction to the `mode` and addition for scatter. Ta.
I will need to update tests to pick up additional grouping chart type for the above.

Addressed you comment about `window.uiChart` in comment here: https://github.com/FlowFuse/node-red-dashboard/pull/1336#discussion_r1783199460

Steve-Mcl:
Tests updated/added to cover changes for scatter and bar.

All passing bar one of the control spec tests (usual suspect) (really would love to get that reliable!)

joepavitt:
> (really would love to get that reliable!)

@Steve-Mcl - @cstns and I have collectively spent days trying - we've given up.

joepavitt:
Works great - thanks @Steve-Mcl

joepavitt (approved):

PR #1349, opened 2024-09-30
Table: Support key type option for entering fixed strings as item labels

### LINKED ISSUE #1188: Typedinput for buttons in ui-table
As discussed [here](https://github.com/FlowFuse/node-red-dashboard/pull/1171#issuecomment-2277380023) a typedinput for the buttons in a ii-table would be more user friendly.

I will take care of this as soon as I have time.

Steve-Mcl:
closes #1188

## Description

* Alters "key" field to "value" and changes it to a typed input
  * fall back is type "key" (for backwards compatibility
* Adds e2e tests to ensure value rendered obeys `keyType` 

chrome_uthk6yoxbC

## Related Issue(s)

#1188

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Functionally, great, but I'm not convinced on the "key" icon. Few reasons:

1. `key` icon to me speaks security key
2. `key` as an icon to refer to the key of an object may not translate well into other languages.
3. If we're to be consistent with the `chart` node, it should just be the text "key"

joepavitt (commented):
Functionally, great, but I'm not convinced on the "key" icon. Few reasons:

1. key icon to me speaks security key
2. key as an icon to refer to the key of an object may not translate well into other languages.
3. If we're to be consistent with the chart node, it should just be the text "key"

Steve-Mcl:
Removed key icon

Steve-Mcl:
Tests are passing. Failure is the usual:

```
  1) Node-RED Dashboard 2.0 - Control - Enable/Disable
       can enable/disable a particular page from the navigation options:
     AssertionError: Timed out retrying after 4000ms: expected '<a.v-list-item.v-list-item--link.v-list-item--nav.v-theme--nrdb.v-list-item--density-default.v-list-item--one-line.v-list-item--rounded.v-list-item--variant-text>' to have class 'v-list-item--disabled'
      at Context.eval (webpack://@flowfuse/node-red-dashboard/./cypress/tests/widgets/control.spec.js:137:51)
```

NOTE: Ideally we would fix this but for the time being, can we either disable this test or move the control tests to their own npm script and GH task so that we get clear passes on the other tests?

nodes/widgets/ui_table.html
```diff
@@ -178,14 +179,27 @@
...
+                        placeholder: RED._('@flowfuse/node-red-dashboard/ui-table:ui-table.label.value'),
                         value: col.key
                     }).appendTo(keyRow)
+                    const keyType = ['str', 'key'].includes(col.keyType) ? col.keyType : 'key'
+                    $('<input/>', {
+                        class: 'node-input-column-keyType',
+                        type: 'hidden',
+                        value: keyType
+                    }).appendTo(keyRow)
+                    console.log('ui-table: setting up typedInput for key')
```
joepavitt, line 196:
```suggestion
```

joepavitt (changes requested):

joepavitt (approved):

PR #1355, opened 2024-10-01
Button: Add minor amendments to button config for click events & docs

joepavitt:
## Description

Small follow up for https://github.com/FlowFuse/node-red-dashboard/pull/1252 which adds an option for whether or not the `click` event is emitted, and ensures it's clear that the "topic" defined is _always_ sent.

It also fixes a bug where the `pointerup` event was taking the `pointerdown` `payloadType`.

Screenshot 2024-10-01 at 10 57 35

These changes then enable us to easily monitor how long a button is held down for, such as:

https://github.com/user-attachments/assets/e416a023-af9d-4ba9-9542-185a9cd10a6c


nodes/widgets/ui_button.js
```diff
@@ -40,6 +45,8 @@ module.exports = function (RED) {
                 break
             }
 
+            console.log('payload', payload, payloadType)
+
```
joepavitt, line 49:
```suggestion
```

Steve-Mcl:
> Screenshot 2024-10-01 at 10 57 35


IMHO; this is confusing. Based on the operations we are now offering, I feel the arrangement of events would probably suit being something like:

image

or 

image

or 

image



Also, I see you have merged the original PR but as far as I can see, the OP did not address the lack of keyboard operation as raised in the comments: https://github.com/FlowFuse/node-red-dashboard/pull/1252#issuecomment-2361239656 

> Keyboard operations are not supported. i.e. tab onto the button so it is highlighted, press the space bar - the "down" event is not triggered. Releasing the space bar does not fire the "up" event either (the "click" event does trigger)

joepavitt:
> Also, I see you have merged the original PR but as far as I can see, the OP did not address the lack of keyboard operation as raised in the comments:

Not a blocker for why this shouldn't have been merged though. Raise a new issue please.

> IMHO; this is confusing

Curious to know what is confusing about it? Given the layout we have, what are you expecting it to send vs. what it actually sends?

By having the "topic" option just as a standalone row, I'm not convinced it's clear to the user what is being done with that topic.

I did toy with the idea of a topic per button event, but we already have `msg._event.type` to differentiate the type, so didn't think it necessary.

Steve-Mcl:
> > IMHO; this is confusing
> 
> Curious to know what is confusing about it? Given the layout we have, what are you expecting it to send vs. what it actually sends?
> 
> By having the "topic" option just as a standalone row, I'm not convinced it's clear to the user what is being done with that topic.
> 
> I did toy with the idea of a topic per button event, but we already have `msg._event.type` to differentiate the type, so didn't think it necessary.

Looking closer, I think I missed a subtly - the `payload` field for the "click" event is hidden due to the click [x] being unticked.

I _assumed_ there was a payload field below the topic field (cut short in the screenshot) because it was required for the "click through" support.  but on second thought, the "click through" doesnt use the entered payload right?

joepavitt:
> but on second thought, the "click through" doesn't use the entered payload right?

This is a good point... I think it does, as it's an "Emulate Click"? I'll need to check

Steve-Mcl:
> > but on second thought, the "click through" doesn't use the entered payload right?
> 
> This is a good point... I think it does, as it's an "Emulate Click"? I'll need to check

Any conclusion Joe?

Steve-Mcl:
> > > but on second thought, the "click through" doesn't use the entered payload right?
> > 
> > 
> > This is a good point... I think it does, as it's an "Emulate Click"? I'll need to check
> 
> Any conclusion Joe?

I've checked and we are correct (unfortunately).  As of now (before this PR), the payload value set for "click" on the node edit form IS what is used for "emulate click". That means hiding it is not ideal I'm afraid.

joepavitt:
Screenshot 2024-10-12 at 18 41 32

Okay, I've moved the "Emulate Button Click" option to be nested in the optional button click section, so that yo can't get to the emulate, without also enabling the click event in the first place, which feels sensible.

joepavitt:
Can confirm, when button emulated is enabled, but the `enableClick` is `false`, then emulated event _will not_ happen.

nodes/widgets/locales/en-US/ui_button.html
```diff
@@ -23,11 +23,13 @@ <h3>Properties</h3>
...
         <dt>Enable pointerup event <span class="property-type">bool</span></dt>
-        <dd>If enabled, the button will emit a message on pointerup event with the given payload</dd>
+        <dd>If enabled, the button will emit a message on pointerup event with the given payload. The output will contain <code>msg._event</code> which details the <i>type</i> of interaction causing the event.</dd>
+        <dt>Enable Click event <span class="property-type">bool</span></dt>
+        <dd>If enabled, the button will emit a click event, which fires when both the up and down events occur over the button. The output will contain <code>msg._event</code> which details the <i>type</i> of interaction causing the event.</dd>
         <dt>Emulate Button Click <span class="property-type">bool</span></dt>
-        <dd>If enabled, any received message will trigger a button click, emitting the relevant payload and topic</dd>
+        <dd>If enabled, any received message will trigger a button click, emitting the relevant payload and topic.</dd>
```
Steve-Mcl, line 32:
```suggestion
        <dd>If enabled, any received message will trigger a button click, emitting the relevant payload and topic. Note, requires the "Enable Click event" to be enabled</dd>
```

Steve-Mcl (approved):

PR #1357, opened 2024-10-02
Chart - New Type: Histogram

### LINKED ISSUE #1339: Chart Type: Histogram
### Description

I just went to plot a chart of data that I had, detailing FlowFuse self-hosted telemetry. In it, we have entries like the following:

```json
{
    "age": 90,
    "ff_version: "2.9.0"
}
```

I want an easy way to plot a histogram/frequency chart of this combination, with the `age` on the x-axis, and the `ff-version` being the `series`, then the y- should just count the unique appearances of these combinations.

In it's simplest form, the Histogram Chart could have `Series: None` and `x: age`, and that would give me a histogram or just the frequencies each `age` appears.

joepavitt:
## Description

- Adds a new chart type "Histogram"
- Adds new x-axis type, "Bins"
- Any data ingested is tallied based on the frequency that the unique combination of given x and series values are seen
- Possible to render categorical histograms, where the "bin" is a string, but also works with numerical data whereby a user can specify a min/max/bins in order to quantify the x-value into the relevant bin.

#### Stacked Categorical Data

Random 100 sample data points from our FlowFuse self-hosted Ping data showing age of the instance (capped at 90 days) and broken down by FF Version

Screenshot 2024-10-02 at 15 05 08

#### Button Payload Injection

https://github.com/user-attachments/assets/a4331823-d858-43d3-b294-48ec291d51ce

#### Slider Examples

https://github.com/user-attachments/assets/e59dad9f-4d2a-4002-9c05-6e21b1fba1cd

## Related Issue(s)

Closes #1339 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
Joe, if the users enters a bin configuration that results in fractions, we get this:

image

should we do some validation AND/OR perhaps round to a few decimal places?

I realise it is tricky - for example min bin of 0.0000, 10 bins max bin 0.0010 would not be possible if we enforce integer resolution and likewise, rounding to 2 decimal places would not present the small fractional bins!  

any thoughts?

joepavitt:
Rounding for sure, not sure we should add the validation though for now.

Steve-Mcl:
> Rounding for sure, not sure we should add the validation though for now.

Ok, I'll add it as a follow up task once review is done.  Not be too long now. Almost done.

nodes/widgets/ui_chart.html
```diff
@@ -80,11 +80,41 @@
                 xAxisType: { value: 'time' },
                 xAxisFormat: { value: '' },
                 xAxisFormatType: { value: 'auto' },
+                xmin: {
+                    value: '',
+                    validate: function (value) {
+                        if (this.chartType === 'histogram' && this.xAxisType === 'bins') {
+                            return typeof value !== 'undefined' && value !== '' && RED.validators.number()
```
Steve-Mcl, line 87: this does not validate the field correctly. 

e.g. I can enter  a bunch of spaces `  ` or even  `  aaa  ` and the field is not marked invalid, deploys without warning and the chart fails to show any data (and no errors are generated client side either)

chrome_RlWHJ5kLlD

Steve-Mcl: I have prepared a change locally and will push shortly to address the ineffective validations

In short, the RED.validator.number incorrectly returns `true` for certain scenarios.


### fixed locally demo
chrome_pCZbYR2TPP

Steve-Mcl: fixed number validation in [fix number validation](https://github.com/FlowFuse/node-red-dashboard/pull/1357/commits/39ec9544a8d176885471ec61cf1d420331e756b1)  [39ec954](https://github.com/FlowFuse/node-red-dashboard/pull/1357/commits/39ec9544a8d176885471ec61cf1d420331e756b1)

ui/src/widgets/ui-chart/UIChart.vue
```diff
@@ -24,33 +24,54 @@ export default {
...
+        },
         radialChart () {
             // radial charts have no placeholder in ChartJS - we need to add one
             return this.props.xAxisType === 'radial'
         }
     },
     watch: {
+        chartType: function (value) {
+            this.chart.config.type = value
+            this.update()
```
Steve-Mcl, line 57:
```suggestion
            this.update(false)
```
As this is a prop update, should it not also use `false` (i.e. debounce multiple quick successive prop changes)

ui/src/widgets/ui-chart/UIChart.vue
```diff
@@ -24,33 +24,54 @@ export default {
...
+        },
         'props.label': function (value) {
             this.chart.options.plugins.title.text = value
-            this.chartUpdate(false)
+            this.update()
```
Steve-Mcl, line 61:
```suggestion
            this.update(false)
```
As this is a prop update, should it not also use `false` (i.e. debounce multiple quick successive prop changes)

Steve-Mcl (commented):
couple of comments, otherwise good

ui/src/widgets/ui-chart/UIChart.vue
gayanSandamal: Discussed with @Steve-Mcl and applying suggestions

ui/src/widgets/ui-chart/UIChart.vue
gayanSandamal: As discussed with @Steve-Mcl applying the suggestions

Steve-Mcl (approved):

Issue #1370, opened 2024-10-08
running old flows against new src causes errors


Steve-Mcl:
### Current Behavior

it is possible to get a situation where breakpoints are not defined

![image](https://github.com/user-attachments/assets/326716c0-2184-4ced-b228-5651b8eb5dff)

### Expected Behavior

should guard or provide defaults

### Environment

- Dashboard version: git source
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

gayanSandamal:
> @gayanSandamal can you take this is a high priority please, we'll need this addressed before a release.

I will prioritise this

PR #1374, opened 2024-10-10
Provided default breakpoints when there's nothing defined

gayanSandamal:
### Problem
When using old flow which doesn't have screen size breakpoints if throws `undefined` as the breakpoints are not defined under `const breakpoints = b.sort((a, b) => a.px - b.px)`

### Solution
Provided default breakpoints when there's no breakpoints defined as per the [documentation](https://dashboard.flowfuse.com/nodes/config/ui-page.html#mobile)

_This doesn't require any updates of the documentation of E2E tests_

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/1370

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/mixins/responsiveness.js
+++ b/ui/src/mixins/responsiveness.js
@@ -26,12 +26,6 @@ export default {
     methods: {
         countColumns () {
             let cols = 0
-            const defaultBreakpoints = [
-                { name: 'Default', px: 0, cols: 3 },
-                { name: 'Tablet', px: 576, cols: 6 },
-                { name: 'Small Desktop', px: 768, cols: 9 },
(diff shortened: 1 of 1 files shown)


gayanSandamal:
> default

@Steve-Mcl thank you for the suggestions provided. I've tested them locally with all the scenarios we can have in the `this.page.breakpoints` such as `undefined`, `null`, `{}`, `[]` and `""`.
Guard you suggested worked perfectly all the scenarios

Steve-Mcl (approved):

PR #1353, opened 2024-10-01
Fixed UI Gauge Dial rendering issue

gayanSandamal:
### Problem
UI-Gauge (dial with needle) was loosing the track of its position when resizing the container.

### Solution
updated dial gauge and its needle on dial gauge container resize

https://github.com/user-attachments/assets/54c516fb-7abd-448b-9c8e-de48c1aa02b0

## Related Issue(s)

[ui-gauge 2.0 1.17.1 - The needle loses track of it's position](https://github.com/FlowFuse/node-red-dashboard/issues/1337)

This doesn't need documentation or E2E test update

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #1390, opened 2024-10-12
Gauge - Revert to using only width to calculate value font-size

### LINKED ISSUE #1385: V1.18.0: Rounded Half Gauge: tiny KPI value
### Current Behavior

Half Gauge - Rounded:  The KPI value (5.4) is tiny and out of proportion compared with the previous version.

image

### Expected Behavior

The value/unit (24.6) is big and visible
image

### Steps To Reproduce

_No response_

### Environment

Dashboard version: 1.18.0
Node-RED version: 4.0.5
Node.js version: v20.6.0
npm version: 9.8.1
Platform/OS: MacOS
Browser: Chrome Version 128.0.6613.138 (Official Build) (arm64)
 

joepavitt:
## Description

- Problem introduced in https://github.com/FlowFuse/node-red-dashboard/pull/1353/files#diff-db418d12c562abfca7a5d735428a2e24ac2cd54a3795550187a71f698335474fR396-R409 
- It used the height _or_ width to calculate the value font-size, was causing the `height` to win too often and therefore rendering very small.
- This reverts it back to how it was, where it just considers the gauge's width.

## Related Issue(s)

Fixes #1385

### DIFF
--- a/ui/src/widgets/ui-gauge/types/UIGaugeDial.vue
+++ b/ui/src/widgets/ui-gauge/types/UIGaugeDial.vue
@@ -394,18 +394,14 @@ export default {
         },
         resizeText () {
             const clientWidth = this.$refs.value?.clientWidth
-            const clientHeight = this.$refs.value?.clientHeight
-
-            // work out how much space we have within which to render the value/icon/range
-            const minDimension = Math.min(clientWidth || 0, clientHeight || 0)
             this.size = 'default'
-            if (minDimension < 80) {
+            if (clientWidth < 80) {
                 this.size = 'xxs'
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):

selwynjackson:
@joepavitt Hi Joe, can you also dynamically change the font size of the min and max text for the half gauge.  Can one overide the min and max text to use say 15k instead of 15000 to use less space?

joepavitt:
That's a very good idea @selwynjackson - we don't currently support that, but would really appreciate it if you opened a new issue/feature request to cover that please?

PR #1399, opened 2024-10-14
Wysiwyg editor patch

Steve-Mcl:
## Description

Implement discussed changes (with some executive decisions) to enable on demand editing of supported page types (Grid + Flex)

### Backend
* original PR did not work with node-red security (1/2 day figuring that out)
  * Needed to pick up the access token from the local storage
* Changed call to deploy flows from V1 to V2 API
  * this helps with merge conflict resolution
* Changed deploy mode from "full" to "nodes"
  * "full" edits trigger any nodes set to run on start. Also, with "full" mode, **all nodes get destroyed and recreated** that includes non-dashboard stuff (like PLC comms, MQTT comms, in-memory context etc) would all being taken down and rebuilt due to someone _just tweaking a dashboard layout_.
* changed dashboard route to include `api/v1/`  path segment
  * this is to clarify intent (this is an API)
* Added validation checks (server side)
  * return 400 + message if there was nothing to deploy (message is shown to user)
    *  this was important since a bug or issue would make it _appear_ flows have been deployed (but nothing was!)
  * check dashboard base id matches, returns 403 if not (plus message)
  * adds an ephemeral "edit key" that is verified upon applying edits, returns 403 if not (plus message)
  * Adds page and base IDs to the PATCH req and checks they match on the server, returns 403 if not (plus message)
  * Ensures every changed group still exists on server before posting flows
    * this is for integrity - before we do a deploy - ensures server flows have not changed while user edited dashboard
  * Ensures every changed groups page ID matches the original flow groups page ID
    * this is for integrity - before we do a deploy  - ensures server flows have not changed while user edited dashboard
  * Only perform POST of new flows IF **changes** are detected.
    * avoids interrupting other users if there is nothing actually changed.
    * although in theory this is handled at the client side, it is possible (due to string -vs- number differences to have differences that dont apply or make any change to the flow)
* Added 2nd dashboard API endpoint for enabling edit mode
  * This is key to enabling edit mode and ONLY for one page. 
  * If another user tries to edit a different page, the single edit-key will be changed and will invalidate/prevent the other edits. The aim is to avoid multiple edits
  
### Front end (dashboard)
* fixed tooltips (kinda*)
   * Could not get v-tooltip component with slots to work!!!
   * Changed the buttons to use v-tooltip attribute instead.
* Implemented "Exit Edit Mode" functionality
* Block changes being applied on every mouse move event
  * was getting really REALLY slow
  * was causing back pressure and browser lock ups / artifacts
* Moved toolbar controls to own component
* Added re-useable `ConfirmDialog` so that we could:
  * Prompt to Confirm losing changes when exiting edit mode (if unsaved/dirty)
  * Prompt to Confirm user is certain they want to deploy flows
    * avoid accidental deploys. clarifyies what is really happening (so no suprises when they see the "merge your changes" in node-red!
  * Feed back issues
    * like when server has changed or permissions are not met or there is network failure etc - user is informed that "save" did not work & thus get the ability to try again.
* Moved lot of logic to the mixin to minimise repeated code on layout pages
* Implemented necessary changes on the Flex and Grid vue components
* Deleted the temporary WYSIWYG vue POC file

### Front end (node-red)
* added on-hover "Layout Editor" buttons
  * Removed text on buttons to make them all fit (used the original titles as tooltips)
    * This new button is conditioned to only appear on supported pages (Grid and Flex)

#### Toolbar issues
- [x] Larger groups push the toolbar off bottom of screen. This makes it hard to distinguish edit mode is active. Also, not great UX having to scroll down to get toolbar.  should be change to position fixed
- [x] No clear indication of saving / saved / save failure
      * Added error dialog for non 200 response
      * Added spinner to save button
      * Disable toolbar while spinning
- [x] Users page theme affects toolbar buttons visibility. 
    * e.g. setting black background and light foreground makes the undo button effectively invisible
    * DONE - added color="secondary" and color="success" to undo and save. Also added `filter: grayscale(1);` to disabled buttons via CSS

#### Resize issues
- [x] random issue - group collapsing to size 1 while resizing
    * resize my a few pixels, let go of mouse (i.e. don't resize) - it skips straight to size: 1
- [x]  resize issue - cannot be dragged back to original size
- [x] seizure inducing flip-flop when resizing and the position is between snap points
- [ ]  resize skipping some snap points.  
     * To recreate, start with a single size 6 group on a 12 grid. Drag it to size 1 and keep hold - now size up - it skips snap points 2,3,4 then starts working again. 
- [ ]  resize snap points are inconsistent
    * Currently, the group jumps ahead to the next snap point, other times it doesn't snap until you reach the snap point.
    * in short, it changes depending on how you operate and is quite disconcerting
    * the improvement could be to modulo the dx on stepX and snap at the halfway point - OR - calculate snap zones
      * snap zones _should_ fix the flopflop issue
- [ ]  Mouse cursor pointer gets changed to the halt pointer while resizing (should stay as the ew-resize pointer)

#### Ordering issues
- [x]  Not obvious groups can be moved
  * DONE Set cursor to grab on :hover
- [x] seizure inducing flip-flop when moving a group up over different sized group
    * DONE - don't snap into position until the moving component is within the area it would actually fit into. At this point, the groups get actually re-ordered but since we are holding the grid that just landed, it has the same index and so doesn't flip-flop
- [ ]  Mouse cursor pointer gets changed to the halt pointer while moving (should be the "grabbing" pointer)

#### General issues
- [ ]  when widgets associated with a group have a set size (non auto), they appear in error after refresh of node-red
- [ ]  resizing sets the width to a `number` but flow source is a `string`. This means the undo "lights up" (is dirty) even when it is not. 
- [ ]  possible refactoring and de-duplication in grid+flex vue pages around edit controls events
      * sortable.js not utilised (delete or implement)

#### For consideration:
  * set mouse pointer to grab open when on a place than can drag group - OR/AND - add a visible grabbable element 
  * move the save/discard/cancel logic to mixin or EditControls to avoid duplicate code?
* sometimes when user enters edit page it gets refreshed a few seconds after fully loaded
  * this interrupts the workflow and loses any sizing or moving user has done
  * This emanates from file:@vite-plugin-pwa/virtual:pwa-register/vue 
  * It is related to workbox wb.addEventListener("activated"... and `event.isUpdate` is true causes `window.location.reload();` to be called. 

#### UX considerations
* I really dislike that this posts to the FLOWS endpoint.
  An alternative (that flexdash did) is to simply send edits to the editor & then let the user press deploy on Node-RED.
* While editing, I would like an overlay on Node-RED that says "currently editing" and have a "cancel edits"
* Add a pencil icon to the side bar entry when editing a page to make it clear page is being edited
* Consider updating title of browser TAB to indicate edit mode.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Discussed the following outstanding items on a call:

> resize skipping some snap points.

> resize snap points are inconsistent

Both okay for now

> Mouse cursor pointer gets changed to the halt pointer while moving (should be the "grabbing" pointer)

This is a strange one, needs investigation before merging.

> when widgets associated with a group have a set size (non auto), they appear in error after refresh of node-red

I see us having two options here:

1. We change the widget size validation, so that widths higher than the parent group are allowed, given that the UI handles there correctly anyway.
2. As part of the API deploy, we _also_ modify any widths of widget's larger than the parenting group that was explicitly edited.

joepavitt:
> Add a pencil icon to the side bar entry when editing a page to make it clear page is being edited

I agree with this, how about a secondary-styled pill with "Edit Mode" text inside next to the page name? This won't be available when the side menu is collapsed, but it's better than nothing

joepavitt:
Summarising Actions:

## Cursor Icon

> Mouse cursor pointer gets changed to the halt pointer while moving (should be the "grabbing" pointer)

Needs investigating

## Widget Size Issue

Implement option 1 as stated in my comment above, disregard widget width > group width validation

## Code Cleaning

Remove `sortable.js`

## Action Icons

Clean the iconography & colours in the "Action" bar. @joepavitt to resolve this.

Steve-Mcl:
Merged into feature branch as discussed with @joepavitt 

I'll take the action points and comments over to the original issue.

Issue #1396, opened 2024-10-13
v1.18.0: ui-chart group tooltips in highy populated charts


m-schaeffler:
### Current Behavior

the tooltips look like this:
![Bildschirmfoto vom 2024-10-12 12-49-33](https://github.com/user-attachments/assets/c87b465b-2217-49c5-bc04-c0194ce0edf2)
![Bildschirmfoto vom 2024-10-12 12-50-09](https://github.com/user-attachments/assets/da536efc-7cf0-44e6-bebb-d337e94a86c4)
![Bildschirmfoto vom 2024-10-13 12-55-46](https://github.com/user-attachments/assets/61c11171-0930-47b2-b378-8227f336d37a)

### Expected Behavior

perhaps it could only show one data point of each line?

### Steps To Reproduce

fill line chart with lots of data points

### Environment

- Dashboard version: 1.18.0
- Node-RED version: 4.0.5
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

joepavitt:
> perhaps it could only show one data point of each line?

Hmm, that's our expected behaviour too. 

Not sure why ChartJS is thinking otherwise. We configured it to read a single point from each series for the matching x-coordinates.

We'll try to reproduce this and get a fix in. Thanks for raising.

PR #1402, opened 2024-10-15
Added interaction mode to 'nearest' to fixed cluttered tooltips

gayanSandamal:
#### Before

https://github.com/user-attachments/assets/6f1ab90d-aae3-4662-86cc-1d5b75240467

#### After

https://github.com/user-attachments/assets/a8bbd9e6-ec33-42ed-b44e-a531f12b7825

_This doesn't require any E2E tests or document updates_

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/1396

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/cypress/tests/widgets/chart-options.spec.js
+++ b/cypress/tests/widgets/chart-options.spec.js
@@ -3,7 +3,7 @@ import should from 'should'
 
 /* eslint-disable cypress/no-unnecessary-waiting */
 describe('Node/-RED Dashboard 2.0 - Chart Widget - chart options', () => {
-    it('sets tooltip `mode` to "nearest" for line charts', () => {
+    it('sets tooltip `mode` to "x" for line charts', () => {
         const chartName = 'line-chart-1'
         const overrides = [
             {
--- a/cypress/tests/widgets/chart-options.spec.js
+++ b/cypress/tests/widgets/chart-options.spec.js
@@ -29,7 +29,7 @@ describe('Node/-RED Dashboard 2.0 - Chart Widget - chart options', () => {
             should(chartObject).is.not.empty()
             should(chartObject.chart.options.type).be.equal('line')
             should(chartObject.chart.options.interaction).be.an.Object()
-            should(chartObject.chart.options.interaction.mode).be.equal('nearest')
+            should(chartObject.chart.options.interaction.mode).be.equal('x')
             should(chartObject.chart.options.interaction).not.have.property('axis')
         })
     })
--- a/cypress/tests/widgets/chart-options.spec.js
+++ b/cypress/tests/widgets/chart-options.spec.js
@@ -3,7 +3,7 @@ import should from 'should'
 
 /* eslint-disable cypress/no-unnecessary-waiting */
 describe('Node/-RED Dashboard 2.0 - Chart Widget - chart options', () => {
-    it('sets tooltip `mode` to "x" for line charts', () => {
+    it('sets tooltip `mode` to "nearest" for line charts', () => {
         const chartName = 'line-chart-1'
         const overrides = [
             {
@@ -29,7 +29,7 @@ describe('Node/-RED Dashboard 2.0 - Chart Widget - chart options', () => {
             should(chartObject).is.not.empty()
             should(chartObject.chart.options.type).be.equal('line')
             should(chartObject.chart.options.interaction).be.an.Object()
-            should(chartObject.chart.options.interaction.mode).be.equal('x')
+            should(chartObject.chart.options.interaction.mode).be.equal('nearest')
             should(chartObject.chart.options.interaction).not.have.property('axis')
         })
     })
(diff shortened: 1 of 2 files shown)


joepavitt:
So, with the _after_ here, I'd still expect to see, in the tooltip, a value for _each_ series, that matches the `x` value. This change means that we only see the single line?

gayanSandamal:
> So, with the _after_ here, I'd still expect to see, in the tooltip, a value for _each_ series, that matches the `x` value. This change means that we only see the single line?

@joepavitt  Thanks for the clarification! That makes sense now. I noticed that the existing implementation is correct and the issue is with the data points; the data point values are keep stacking when taking the cursor through the chart line 

https://github.com/user-attachments/assets/6fb6a2d1-dd36-4866-8328-a1c188ac1ac3

joepavitt:
> the data point values are keep stacking when taking the cursor through the chart line

Ooh, good find, so an underlying ChartJS bug possibly? Can you check through their existing bug list to see if it's something that has been reported.

gayanSandamal:
> > the data point values are keep stacking when taking the cursor through the chart line
> 
> Ooh, good find, so an underlying ChartJS bug possibly? Can you check through their existing bug list to see if it's something that has been reported.

Thanks for the suggestion! I'm checking the ChartJS bug list now. I'll update you soon

joepavitt:
> I'm checking the ChartJS bug list now.

Any update please @gayanSandamal?

gayanSandamal:
> > I'm checking the ChartJS bug list now.
> 
> Any update please @gayanSandamal?

@joepavitt I found the exact same issue reported https://github.com/chartjs/Chart.js/issues/8293

After carefully going through the solutions provided they have are recommending to use 'nearest' (which was my initial solution https://github.com/FlowFuse/node-red-dashboard/pull/1402/commits/2608ef107691a7ea4d9120fff69187fad7bc7e83) option instead `x` and all the examples they have included are single line charts.

But for our scenario we cannot use that when the data points are overlapping as the tooltip gets the values of from the intersection.

I've tried the below options

- [x] Setting `intersect` to `false`
- [x] Set `pointHitRadius` to `0`
- [x] Set `pointRadius` to `0`
- [x] Trying different "Custom Interaction Modes"

The only thing we can do here is reduce the `pointerRadius` and the it doesn't fix this problem entirely, supports up to a certain level of data points and it's not easy to get the tooltip because user needs to carefully place the cursor right on the data points get the tooltip.

Screenshot 2024-10-16 at 01 51 42

joepavitt:
How about using a filter per: https://github.com/chartjs/Chart.js/issues/5231

gayanSandamal:
> How about using a filter per: [chartjs/Chart.js#5231](https://github.com/chartjs/Chart.js/issues/5231)

Thanks a lot. I'm checking this

gayanSandamal:
> How about using a filter per: [chartjs/Chart.js#5231](https://github.com/chartjs/Chart.js/issues/5231)

@joepavitt PR is updated with this commit https://github.com/FlowFuse/node-red-dashboard/pull/1402/commits/21076d11e2aa2de40dcd1ad828760c9972abbf31

PR description is also updated https://github.com/FlowFuse/node-red-dashboard/pull/1402#issue-2587844383

bartbutenaers:
> How about using a filter per: https://github.com/chartjs/Chart.js/issues/5231

Hi @joepavitt,
Without the intension of becoming a party-stopper...
That all seems to be very ChartJs specific stuff.  
Doesn't this kind of code will make it very hard afterwards to switch to another charting library? 
Unless the replacement of ChartJs by another library is not an established fact anymore, then please forget my comment!

joepavitt:
> Doesn't this kind of code will make it very hard afterwards to switch to another charting library?

Yes, but it's part of the ChartJS config, if/when we switch to Apache, we would need to make an equivalent. I can't justify getting Gayan to spend 4-6 weeks on migrating to Apache eCharts at the moment.

> Unless the replacement of ChartJs by another library is not an established fact anymore, then please forget my comment!

It's something we hope to do, but it's a monstrous piece of work, and we're regularly distracted by the many issues and smaller feature requests being open

cypress/tests/widgets/chart-options.spec.js
```diff
@@ -3,7 +3,7 @@ import should from 'should'
 
 /* eslint-disable cypress/no-unnecessary-waiting */
 describe('Node/-RED Dashboard 2.0 - Chart Widget - chart options', () => {
-    it('sets tooltip `mode` to "x" for line charts', () => {
+    it('sets tooltip `mode` to "nearest" for line charts', () => {
```
joepavitt, line 6:
```suggestion
    it('sets tooltip `mode` to "x" for line charts', () => {
```

joepavitt (approved):

Issue #1411, opened 2024-10-17
ui-chart stopped working as expected after update v1.18.1


DiogoGoncalves35:
After the new update, the chart stopped working as expected. I have already tried to include property X as "timestamp", but the problem continues. Do you have any ideas on how to solve this issue?

colinl:
Have you tried restarting node red?
Show us how you have configured the node and a sample of the data you are feeding to it.  Also tell us what is not working.

joepavitt:
Going to need more information here, what data were you passing in? What properties did you have configured? What chart type were you using?

DiogoGoncalves35:
I already tried to restart Node-Red but the problem persists.

Function Node (Before connecting to chart):
`msg.topic = "Input"
msg.payload = {
    "Y_Axis": msg.payload
}
return msg;`

Ui-Chart-Configuration
`    {
        "id": "02f1bf493c1b64ce",
        "type": "ui-chart",
        "z": "5b733c2dd6b81e4c",
        "g": "20b44803b0943e84",
        "group": "0ce29f646ca1844a",
        "name": "Voltage",
        "label": "Voltage",
        "order": 4,
        "chartType": "line",
        "category": "topic",
        "categoryType": "msg",
        "xAxisLabel": "Time",
        "xAxisProperty": "",
        "xAxisPropertyType": "timestamp",
        "xAxisType": "time",
        "xAxisFormat": "ss",
        "xAxisFormatType": "custom",
        "xmin": "",
        "xmax": "",
        "yAxisLabel": "Voltage (V)",
        "yAxisProperty": "Y_Axis",
        "yAxisPropertyType": "property",
        "ymin": "0",
        "ymax": "",
        "bins": "",
        "action": "append",
        "stackSeries": false,
        "pointShape": "circle",
        "pointRadius": 4,
        "showLegend": true,
        "removeOlder": "5",
        "removeOlderUnit": "1",
        "removeOlderPoints": "5",
        "colors": [
            "#0095ff",
            "#ff0000",
            "#ff7f0e",
            "#2ca02c",
            "#a347e1",
            "#d62728",
            "#ff9896",
            "#9467bd",
            "#c5b0d5"
        ],
        "textColor": [
            "#666666"
        ],
        "textColorDefault": true,
        "gridColor": [
            "#e5e5e5"
        ],
        "gridColorDefault": true,
        "width": 6,
        "height": "9",
        "className": "",
        "x": 780,
        "y": 2160,
[...]

PR #1417, opened 2024-10-22
Axis updating issue fix for limited points while maintaining the fix the for #1255

gayanSandamal:
## Description

This fixes the issue #1411 while maintaining previous fix done for #1255 

https://github.com/user-attachments/assets/0591f71c-98b9-4ea1-90a2-2e18e999bf38

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/1411

_Docs update is not necessary_

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/cypress/fixtures/flows/dashboard-chart-mulitline.json
+++ b/cypress/fixtures/flows/dashboard-chart-mulitline.json
@@ -1,0 +1,226 @@
+[
+    {
+        "id": "1bc5a69e1358d15b",
+        "type": "split",
+        "z": "d03c59d3654ae31a",
+        "name": "",
+        "splt": "\\n",
+        "spltType": "str",
+        "arraySplt": 1,
+        "arraySpltType": "len",
+        "stream": false,
+        "addname": "",
+        "x": 290,
+        "y": 480,
+        "wires": [
+            [
+                "126b3ed93bfb87c6"
+            ]
+        ]
+    },
+    {
+        "id": "126b3ed93bfb87c6",
+        "type": "change",
+        "z": "d03c59d3654ae31a",
+        "name": "",
+        "rules": [
+            {
+                "t": "move",
+                "p": "payload.topic",
+                "pt": "msg",
+                "to": "topic",
+                "tot": "msg"
+            },
+            {
+                "t": "set",
+                "p": "payload",
+                "pt": "msg",
+                "to": "payload.payload",
+                "tot": "msg"
+            }
+        ],
+        "action": "",
+        "property": "",
+        "from": "",
+        "to": "",
+        "reg": false,
+        "x": 460,
+        "y": 480,
+        "wires": [
+            [
+                "dashboard-ui-chart-multiline"
+            ]
+        ]
+    },
+    {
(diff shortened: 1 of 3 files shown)


joepavitt (commented):
Whilst I can see that this works, I'm not sure I understand why the code changes that were made, were made. I was expecting something in `limitDataSize()` to be changed, but it hasn't.

Can you talk me through the changes you've made here, and the logic behind why those changes fix the problem please?

gayanSandamal:
> Whilst I can see that this works, I'm not sure I understand why the code changes that were made, were made. I was expecting something in `limitDataSize()` to be changed, but it hasn't.
> 
> Can you talk me through the changes you've made here, and the logic behind why those changes fix the problem please?

Hi @joepavitt ,

Thanks for your feedback! Let me walk you through the changes and the reasoning behind them.

The issue we were addressing had two aspects:

1. **Preserving point data across updates**: In the previous implementation, when a new x-value was encountered, the logic always initialized an array of known x-values (`xLabels`) with `undefined` objects. This could lead to problems where the first data point wouldn’t draw a connecting line properly (especially for new series or topics). This happened because gaps in the dataset resulted in incomplete arrays, affecting rendering.

2. **Handling the new x-values when `removeOlderPoints` is enabled**: I modified the initialization of the `data` array. Now, if `removeOlderPoints` is enabled, we extend the array length by one to accommodate the newest point before discarding old ones. This ensures that we’re correctly aligning the incoming data points to the corresponding x-values without any skips.

```javascript
let data = Array(xLabels.length).fill({});
if (this.props.removeOlderPoints) {
  data = Array(xLabels.length + 1).fill({});
}
```

3. **Fixing the handling of new x-values dynamically**: If a new x-value arrives and `removeOlderPoints` is active, we store it at the tail of the array rather than introducing undefined gaps:

```javascript
if (this.props.removeOlderPoints) {
  data[xLabels.length] = datapoint;
} else {
  xLabels.push(datapoint.x);
  data.push(datapoint);
}
```

This change smoothen the transitions in the chart without skipping the first data point when switching topics or adding new series dynamically.

joepavitt:
@gayanSandamal the logic you've detailed here doesn't really address the _why_ the problem was occurring. The key function for trimming data is `limitDataSize`, the logic for which data points to retain is stored here.

In actual fact, the reason your code changes fix the problem is because you're not adding the `datapoint.x` to the array of `xLabels` stored on the chart. In preventing that action, the line-charts work again. 

The actual fix here is in `limitDataSize`, and is to ensure that the length of `xLabels` on the chart is the same as the data points when rendering scatter/line charts. Alternatively, I actually think we have no need to be storing `xLabels` at all for scatter/line charts, I'm playing locally, and see no need/benefit for them at all.

This can be quickly checked by removing the `xLabels.push()` line from the code base, without any of the changes you've included here, which fixes the cut-off points.

The line was rendering too long as we were keep storage of _all_ `xLabel` values, but not a corresponding `x,y` pair to associate with the label.

PR #1429, opened 2024-10-29
Drive app name via Base's "Name" field

### LINKED ISSUE #1375: Sync the dashboard name in the UI with the base node Name field
### Description

I'm raising this followup task as per the below PR review suggestion made by @Steve-Mcl 

Original comment - https://github.com/FlowFuse/node-red-dashboard/pull/1366#pullrequestreview-2359426184

> Looks good to me.

> Question, shouldnt the `name` come from the base node `Name` field?

> image

> As it stands, all of my dashboards are called "Node-RED Dashboard 2.0" - not a great UX IMHO.

> This can of course be a follow up task.

### Epic/Story

_No response_

joepavitt:
## Description

- Removes the hardcoded `name` in the `manifest`.
- In theory, this should be served up dynamically on each request, but when using dev tools, the `name` wasn't updated until after a Node-RED restart which is odd, but it does update/work.
- Updated documentation too

## Related Issue(s)

Closes #1375

### DIFF
--- a/docs/user/pwa.md
+++ b/docs/user/pwa.md
@@ -35,6 +35,12 @@ You'll then be prompted to name your application:
 
 ## Customization
 
+### Name
+
+You can personalize your application's name to enhance branding and make it easily recognizable. 
+
+To customize the name, select "Edit Settings" in the Dashboard 2.0 sidebar. Then modify the "Name" field. You may need to restart Node-RED in order for the change to take effect in the application when installed.
+
 ### Icon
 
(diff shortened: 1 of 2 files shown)


Steve-Mcl (approved):
I was thinking these should have fallback values since a config node (or any node) typically does NOT require a `name` to be set.

However, this permits users to set the name deliberately empty.

joepavitt:
Confused by the feedback, are you saying it's okay to proceed as-is?

PR #1440, opened 2024-11-04
WYSIWYG: Tidy up styling/positioning of the Edit Controls

joepavitt:
## Description

Screenshot 2024-11-04 at 11 18 42

- Improve the colouring of the buttons in the Edit Controls and confirmation Dialog.
- Move the "Actions" to the top of the screen, as they felt a little lost floating in the middle at the bottom.
- Remove the "Pencil" icon from the page title, and instead have an "In Edit Mode" note under the actions bar

PR #1441, opened 2024-11-04
WYSIWYG: Fix edit controls taking full width fo the page

joepavitt:
## Description
Fixes bug introduced in #1440

### DIFF
--- a/ui/src/layouts/Grid.vue
+++ b/ui/src/layouts/Grid.vue
@@ -239,11 +239,6 @@ export default {
     gap: var(--group-gap);
 }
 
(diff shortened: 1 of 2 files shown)


Steve-Mcl (changes requested):
Can we remove the old (duplicate) css values (left, top, width) please Joe

https://github.com/FlowFuse/node-red-dashboard/pull/1441/files#diff-b4bd132ef6a4f90612e90d3efaa366e22250f590b72744cb59ca6b02a19a7b40R43-R46

Steve-Mcl (approved):

PR #1460, opened 2024-11-12
Chart - Clear stale data from front-end chart

### LINKED ISSUE #755: Stale data not being weeded from chart
### Current Behavior

I have a chart being fed data every 5 seconds, and have set the chart config to append data points and x-axis limit to 1 hour.
When I first load the chart, or refresh the browser I get exactly that, 1 hour of data. But if I leave the browser connected, the stale data (more than 1 hour old) is not removed from the chart, as per the screenshot.
I'm concerned that if the browser was kept connected, that the amount of data (points) being stored would affect the stability of the system due to memory constraints.

chart

config

### Expected Behavior

I expected that chart data older than 1 hour would be removed from the chart after 1 hour, however that is not the case.

### Steps To Reproduce

Create a chart configured as above, and leave the browser connected to the chart.
Forum link to discussion - https://discourse.nodered.org/t/stale-data-not-being-weeded-from-chart/87039

### Environment

- Dashboard version: 1.7.0
- Node-RED version: 3.1.8
- Node.js version: 20.12.0
- npm version: 10.5.0
- Platform/OS: Ubuntu on Oracle cloud server
- Browser: Chrome Version 123.0.6312.106 (Official Build) (64-bit) (Although same result with Edge)

joepavitt:
## Description

- Ensure we consider the `cutoff` _and_ `points` in the front-end Chart filtering of stale data.
- Also extracted out the `clearOldPoints` function in the server-side to improve readibility/discoverability.

## Related Issue(s)

Closes #755


joepavitt:
Yeha, that E2E isn't looking healthy, I'll take a look when I have a chance

joepavitt:
Test should now be fixed - rerunning in CI to be sure

PR #1461, opened 2024-11-12
Add Light & Dark Theme

cgjgh:
#### Adds light and dark mode to D2. 

- Colors are configurable in NodeRED editor. 
- Mode can be toggled from D2 nav sidebar. 
- Last setting is saved to and loaded from cache on reloads.

Still WIP...

https://github.com/user-attachments/assets/3416ae2a-e9d1-420a-87dd-d71daf2508f9
## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


nodes/config/ui_base.html
```diff
@@ -1895,14 +1895,22 @@
...
-                const palette = $('<div>', { class: 'nrdb2-sb-palette' }).appendTo(titleRow)
+                const paletteLight = $('<div>', { class: 'nrdb2-sb-palette' }).appendTo(titleRow)
                 const colors = theme.colors
 
-                palette.append($('<div>', { class: 'nrdb2-sb-palette-color', style: `background-color: ${colors.surface}` }))
-                palette.append($('<div>', { class: 'nrdb2-sb-palette-color', style: `background-color: ${colors.primary}` }))
-                palette.append($('<div>', { class: 'nrdb2-sb-palette-color', style: `background-color: ${colors.bgPage}` }))
-                palette.append($('<div>', { class: 'nrdb2-sb-palette-color', style: `background-color: ${colors.groupBg}` }))
-                palette.append($('<div>', { class: 'nrdb2-sb-palette-color', style: `background-color: ${colors.groupOutline}` }))
+                paletteLight.append($('<div>', { class: 'nrdb2-sb-palette-color', style: `background-color: ${colors.surfaceLight}` }))
```
cgjgh, line 1901: Still deciding between `colors.light.surface` or `colors.surfaceLight`...

ui/src/layouts/Baseline.vue
```diff
@@ -66,14 +78,34 @@
     </v-app>
 </template>
 
-<script>
-import { mapGetters, mapState } from 'vuex'
+<script setup>
+import { useTheme } from 'vuetify'
```
cgjgh, line 82: Not sure if these would be better placed in `App.vue`...

nodes/config/ui_theme.html
```diff
@@ -93,6 +110,28 @@
...
+            const tabs = RED.tabs.create({
+                id: 'color-tabs',
+                onchange: function (tab) {
+                    $('#color-tabs-content').children().hide()
+                    $('#' + tab.id).show()
+                }
+            })
+            tabs.addTab({
+                id: 'color-tab-light',
+                iconClass: 'fa fa-sun',
```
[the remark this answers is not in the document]
joepavitt: Node-RED uses Font-Awesome 4, the only sun-based icon there is `fa-sun-o`

ui/src/layouts/Baseline.vue
joepavitt: I think here is sensible as all layouts extend this `Baseline` anyway

nodes/config/ui_theme.html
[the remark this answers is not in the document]
cgjgh: Got it, thanks!

ui/src/layouts/Baseline.vue
cgjgh: Ok, great.

joepavitt:
I need some time to think about this one, so won't put it in with this current release (sorry for the delay in getting to this in the first place)

I understand the value in having a dark/light theme, but I'm thinking, as a first point of call, we could just offer to two themes as default in the configuration of the Dashboard, rather than expose a light/dark switch more in the side navigation directly. The positioning of the toggle is also tricky, I have dashboards, and have seen many others that have a full sidebar of pages, so this will float above a page and get in the way of navigation - that said, I can't necessarily think of better positioning.

cgjgh:
@joepavitt No problem, take your time. I realize there would be significant changes, and some clean-up might be required. 
I also wasn't too sure about the light/dark switch positioning. On the subject, what are your thoughts on using [media query (CSS)](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) to switch themes based on system theme settings?

Alternatively, we could use the JS approach:
```js
if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
    // dark mode
}

// watch for changes
window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', event => {
    const newColorScheme = event.matches ? "dark" : "light";
});
```

Aridzki:
@cgjgh Can I do this using ui_template node

cgjgh:
@Aridzki No, this PR needs to be merged first.

gemini86:
This looks great and I love the idea of being able to manipulate the theme, but may I suggest we simply expose the ability to control theme changes through ui_control node? We can define multiple themes and then direct the ui_node which theme to load programmatically through the flows like anything else.

PR #1394, opened 2024-10-12
Sanitize HTML Injections

bartbutenaers:
## Description

There are 7 code snippets in this repo where `v-html` is being used, without sanetizing the input html string.  Which can be exploit by hackers to execute XSS attacks:

+ [ ] [PropsTable.vue](https://github.com/FlowFuse/node-red-dashboard/blob/main/docs/components/PropsTable.vue#L14)
+ [ ] [EventList.vue](https://github.com/FlowFuse/node-red-dashboard/blob/main/docs/components/EventsList.vue)
+ [X] [UIText.vue](https://github.com/FlowFuse/node-red-dashboard/blob/main/ui/src/widgets/ui-text/UIText.vue#L5)
+ [ ] [ControlsTable.vue](https://github.com/FlowFuse/node-red-dashboard/blob/main/docs/components/ControlsTable.vue)
+ [ ] [MigrationPropertyTable.vue](https://github.com/FlowFuse/node-red-dashboard/blob/main/docs/components/MigrationPropertyTable.vue#L30C66-L32C17)
+ [X] [UIButton.vue](https://github.com/FlowFuse/node-red-dashboard/blob/main/ui/src/widgets/ui-button/UIButton.vue#L17)
+ [X] [UINotification.vue](https://github.com/FlowFuse/node-red-dashboard/blob/main/ui/src/widgets/ui-notification/UINotification.vue#L16)

This is a ***DRAFT*** PR until all 7 code snippets have been refactored.  The fixed code snippets are represented by a checked checkbox in the list above.  Below I will show a demo of how it looked like before I applied the sanetization fix.  Afterwards it was solved, and the popups did not appear anymore.

### ui-text: html text
malicious_ui_text_payload
```
[...]

PR #1466, opened 2024-11-13
Sanetize 'target' attr fix

### LINKED ISSUE #1464: Text: target="_blank" gets stripped from injected HTML values
### Current Behavior

- The rendered `<a />` has the `target="_blank"` removed.
- Adding other attributes is fine, it's just the `target` that seems to get stripped.

### Expected Behavior

- If the HTML has attributes, such as `target` defined, these should not be stripped.

### Steps To Reproduce

Copy the example from our [docs](https://dashboard.flowfuse.com/nodes/widgets/ui-text.html#static-html), inject it into a UI Text under `msg.payload`

### Environment

- Dashboard version: 1.19.1
- Node-RED version: 4.x

bartbutenaers:
## Description

The sanetization of injected html into the ui-text node (see [other](https://github.com/FlowFuse/node-red-dashboard/pull/1394) PR) seems to remove a bit too much attributes, because the `target` attribute of hyperlinks is also removed.

The DOMPurify library offers two ways to customize which attributes should be allowed:
+ `ALLOWED_ATTR` can be used to replace the entire list of default allowed attributes.  Which imho seems to be a bit too invasive.
+ `ADD_ATTR` can be used to retain the default list of allowed attributes and to add a few more.

I have used the latter option to add allow the `link` attribute.  Of course extra attributes can be added in the future if necessary.

The result:

image

## Related Issue(s)

Closes #1464

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt (approved):

joepavitt:
This conflict should get resolved once https://github.com/FlowFuse/node-red-dashboard/pull/1463 is merged, so will wait for Steve's eyes on it before moving forward with this one.

PR #1471, opened 2024-11-14
Improve DX - dont minify dev build

Steve-Mcl:
## Description

Dont minify for dev mode
Also, ditch the source maps (since its not minified!)

Discussed here: https://discourse.nodered.org/t/when-are-vuejs-methods-binded/93142/11?u=steve-mcl

### DIFF
--- a/docs/contributing/index.md
+++ b/docs/contributing/index.md
@@ -56,6 +56,8 @@ A [VitePress](https://vitepress.dev/) documentation site that is used to generat
     ```bash
     npm run build
     ```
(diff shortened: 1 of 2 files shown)


bartbutenaers:
Thanks @Steve-Mcl !
It would also be useful to put your `npm run build:dev` command somewhere.  I see now - for the first time - [this](https://github.com/FlowFuse/node-red-dashboard/blob/main/ui/README.md) page.  Not sure whether it should but put there, because that page is not part of the contribution documentation.  Even not sure why I have never found that readme page, because it contains a.o. the lint command (which I only found recently myself somewhere else).

Steve-Mcl (commented):
update contributing docs

joepavitt (approved):

PR #1483, opened 2024-11-19
Form - Fix form node not passing on msg.topic on submit

### LINKED ISSUE #1453: Form node does not pass on topic when configured to do so
### Current Behavior

When a form is configured with the Topic field set to msg.topic, then if a message with a topic is passed in then that topic should be passed on when the form is submitted.  In this example flow that is not happening, the passed on topic is always an empty string.
```
[{"id":"782b20ba75db1d98","type":"inject","z":"bdd7be38.d3b55","name":"topic: 7","props":[{"p":"topic","v":"7","vt":"num"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":250,"y":3680,"wires":[["93af822611e1a3bd"]]},{"id":"93af822611e1a3bd","type":"ui-form","z":"bdd7be38.d3b55","name":"","group":"4f87bd59a15b847e","label":"Test form","order":0,"width":0,"height":0,"options":[{"label":"","key":"formValue","type":"text","required":true,"rows":null}],"formValue":{"formValue":""},"payload":"","submit":"submit","cancel":"clear","resetOnSubmit":true,"topic":"topic","topicType":"msg","splitLayout":"","className":"","passthru":false,"dropdownOptions":[],"x":470,"y":3680,"wires":[["c47d1b4e28c96bb3"]]},{"id":"c47d1b4e28c96bb3","type":"debug","z":"bdd7be38.d3b55","name":"debug 87","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","statusVal":"","statusType":"auto","x":630,"y":3680,"wires":[]},{"id":"f0ca8088245f8e89","type":"inject","z":"bdd7be38.d3b55","name":"topic: 42","props":[{"p":"topic","v":"42","vt":"num"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":260,"y":3740,"wires":[["93af822611e1a3bd"]]},{"id":"4f87bd59a15b847e","type":"ui-group","name":"test","page":"c6ff182a4185f2f2","width":"6","height":"1","order":1,"showTitle":false,"className":"","visible":"true","disabled":"false"},{"id":"c6ff182a4185f2f2","type":"ui-page","name":"Test page","ui":"ID-BASE-1","path":"/testpage","icon":"home","layout":"grid","theme":"f9b6670b127dc219","order":2,"className":"","visible":"true","disabled":"false"},{"id":"ID-BASE-1","type":"ui-base","name":"Dashboard","path":"/dashboard","includeClientData":true,"acceptsClientConfig":["ui-control","ui-notification","ui-gauge-classic"],"showPageTitle":true,"titleBarStyle":"default"},{"id":"f9b6670b127dc219","type":"ui-theme","name":"FlowForge Theme","colors":{"surface":"#152a47","primary":"#005aff","bgPage":"#ffffff","groupBg":"#ffffff","groupOutline":"#cc3e3e"},"sizes":{"pagePadding":"12px","groupGap":"12px","groupBorderRadius":"4px","widgetGap":"12px"}}]
```

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- Dashboard version: 1.19.1
- Node-RED version: 4.0.5
- Node.js version: 20.5.1
- npm version:
- Platform/OS: Ubuntu
- Browser: Edge

cgjgh:
## Description

Fixes `msg.topic` not being applied on form submits. 
Not completely sure if this meets any previously agreed upon design pattern for applying topics.
## Related Issue(s)
Closes #1453

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/widgets/ui_form.js
+++ b/nodes/widgets/ui_form.js
@@ -1,3 +1,4 @@
+const datastore = require('../store/data.js')
 const statestore = require('../store/state.js')
 
 module.exports = function (RED) {
@@ -13,6 +14,11 @@ module.exports = function (RED) {
         const evts = {
             onAction: true,
             beforeSend: function (msg) {
+                if (msg?._event === 'submit') {
+                    if (config.topicType === 'msg') {
+                        msg.topic = datastore.get(node.id)?.[config.topic] || ''
+                    }
+                }
                 if (msg.ui_update) {
                     const update = msg.ui_update
                     if (typeof update.label !== 'undefined') {
--- a/ui/src/widgets/ui-form/UIForm.vue
+++ b/ui/src/widgets/ui-form/UIForm.vue
@@ -127,7 +127,8 @@ export default {
             })
 
             this.$socket.emit('widget-action', this.id, {
-                payload: this.input
+                payload: this.input,
+                _event: 'submit'
             })
             if (this.props.resetOnSubmit) {
                 this.reset()


joepavitt:
Thanks @cgjgh - this is an odd one, as the `appendTopic` function runs as a universal line in https://github.com/FlowFuse/node-red-dashboard/blob/13c32b005d1d116e2f429418fccbc531ad578f75/nodes/config/ui_base.js#L581-L590

Which is part of the `onAction` handler in our `ui-base`. So the change you've submitted, whilst working, feels like it's covering another issue we have somewhere

bartbutenaers:
> feels like it's covering another issue we have somewhere

@joepavitt,
Yes indeed you are correct.
I think this is what goes wrong with the example flow from @colinl:
1. He injects a message with only a topic, not a payload.  Althought this is not relevant I think, as you can see below.
2. The msg arrives in ui_base->[onInput](https://github.com/FlowFuse/node-red-dashboard/blob/main/nodes/config/ui_base.js#L1067), where the msg (containing the topic) is stored in the datastore and emmited to the frontend.
3. The msg (containing the topic) arrives in the [onInput](https://github.com/FlowFuse/node-red-dashboard/blob/main/ui/src/widgets/ui-form/UIForm.vue#L156C1-L164C11) of the form vue component, where the msg is ignored because it has no payload.  But even if it would have a payload, only that payload would be processed (and the topic would be ignored anyway):
   ```
           onInput (msg) {
            if (msg.payload) {
                const payload = msg.payload
                for (const key in payload) {
                    this.input[key] = payload[key]
                }
                this.$nextTick(() => { this.validate() })
            }
        },
     ```
4. The user clicks on the form "Submit" button.
5. The [onSubmit](https://github.com/FlowFuse/node-red-dashboard/blob/main/ui/src/widgets/ui-form/UIForm.vue#L94) in Vue is called, which will create a new message from scratch (which only contains a payload) to the server:
   ```
   this.$socket.emit('widget-action', this.id, {
      payload: this.input
   })
   ```
6. The frontend msg (only containing a payload) will arrive in the ui_base-->[onAction](https://github.com/FlowFuse/node-red-dashboard/blob/main/nodes/config/ui_base.js#L576) on the server side, where it will append the topic to that msg:
   ```
   msg = await appendTopic(RED, widgetConfig, wNode, msg)
   ```
   However the frontend msg only contains a payload, and the widgetConfig only contains msg.topic (to tell the function that it should get the topic from the msg).  So ***NO*** topic is appended to the output msg...

You can solve this in various ways, but not sure what is the best way to do it:
+ You could store the topic in the frontend, and send it back on commit.  But seems a bit weird to use an entire roundtrip.
+ You could get the topic in the onAction from the last msg in the datastore (if the msg doesn't contain a topic).  But the appendTopic is implemented in a separate util file, so you cannot do that in the util file.  Which means you need to append the topic to the msg before you call the appendTopic function, which is also very weird.

Any thoughts are welcome!

Steve-Mcl:
@cgjgh thanks for your efforts here but I will be superceeding this with #1662.  It uses your methods but also ensures flow/global are evaluated. It also adds e2e tests.

Regarding the reasoning: I have been through the issue & stepped the code - Barts is correct in his summation.

cgjgh:
Glad this is finally resolved!

PR #1487, opened 2024-11-20
Button - Added additional pointercancel eventlistener

JulianStremel:
## Description

I just wanted to make pointerevents more reliably end for hmi applications.
So i added the pointercancel event to the button component that activates if the "pointerup" function is enabled.
This will fire when closing the window while a button is pressed or other navigation changes happen that would otherwise not fire the "pointerup" event.

## Related Issue(s)

No related issue

## Checklist

- [x] No further modifications and updates needed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16078,9 +16078,9 @@
             }
         },
         "node_modules/vite": {
-            "version": "5.4.17",
-            "resolved": "https://registry.npmjs.org/vite/-/vite-5.4.17.tgz",
-            "integrity": "sha512-5+VqZryDj4wgCs55o9Lp+p8GE78TLVg0lasCH5xFZ4jacZjtqZa6JUw9/p0WeAojaOfncSM6v77InkFPGnvPvg==",
+            "version": "5.4.19",
+            "resolved": "https://registry.npmjs.org/vite/-/vite-5.4.19.tgz",
+            "integrity": "sha512-qO3aKv3HoQC8QKiNSTuUM1l9o/XX3+c+VTgLHbJWHZGeTPVAg2XwazI9UWzoxjIJCGCV2zU60uqMzjeLZuULqA==",
             "dev": true,
             "license": "MIT",
             "dependencies": {


Steve-Mcl (changes requested):
Hi. This looks like a good addition however [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointercancel_event) the pointercancel event can fire for many more reasons and I suspect we would be indiscriminately firing the "up" event.

If we are to include this, i suggest we have a "memory" 

Something like:

```
const activeDownPointers = new Set();
```
 On pointer down
 ```
 activeDownPointers.add(event.pointerId);
 ```
 
 On pointer cancel
 ```
     if (activeDownPointers.has(event.pointerId)) {
        activeDownPointers.delete(event.pointerId);
        // call pointer up code
    }
 ```
 
 Untested

JulianStremel:
> If we are to include this, i suggest we have a "memory"

Sounds reasonable !
I will have a second look at it shortly.

JulianStremel:
@Steve-Mcl after some initial testing i couldnt figure cases out in which a "random" pointercancel would be emitet.
I think due to pointercancel beeing a pointerevent it always cancels the active pointer on the pointerdowned element because it obviously needs a pointer to cancel on ... or is my understanding fundamentally wrong ?

Steve-Mcl:
TBH, I dont know (not played with this event before - just regurgitating what I seen on MDN)

When I get time to pull and test, I will know more.

If however in the meantime you decide to play around some more and find it is not necessary, then thats fine too.

joepavitt:
> I will have a second look at it shortly.

Have you had the chance to add in the `memory` proposed by Steve? I think it would be a good catch just in case there are occasions we've not considered where this could mis-fire

joepavitt:
@JulianStremel just checking you've seen the above

JulianStremel:
> @JulianStremel just checking you've seen the above

Sorry for the long silence i had to shift focus to some other projects for the last couple of weeks.

I tested a "memory" implementation and that works like expected... but the pointercancel event as described by the mdn docs only ever fires for active pointers so also tracking active pointers would be double the work for no logical reason im my mind.

PR #1490, opened 2024-11-20
Configurable Connection Notifications

### LINKED ISSUE #811: Ditch the Connection Re-established popup!
### Description

The Reconnecting message that appears after a connection failure is very annoying, as it appears for several seconds almost every time I bring the dashboard to the foreground on Android.  I don't thinks it conveys any useful information.  I cannot safely click the Close button because if one is a fraction of a second late clicking it then the widget underneath receives the click.

I am not a flowfuse employee

cgjgh:
## Description
Adds configurability to connection notification.

## Related Issue(s)

Resolves #811

              

> Configuration option on the `ui-base` seems an easy middle ground here. 
>               Default should be "on" in my opinion.

_Originally posted by @joepavitt in https://github.com/FlowFuse/node-red-dashboard/issues/811#issuecomment-2084790020_

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/locales/en-US/ui_base.json
+++ b/nodes/config/locales/en-US/ui_base.json
@@ -32,10 +32,10 @@
             "navigationStyleIcon": "Collapse to icons",
             "navigationStyleTemporary": "Appear over content",
             "navigationStyleNone": "Always hide",
-            "notifications": "Notifications",
+            "connectionNotifications": "Connection Notifications",
             "showReconnect": "Show reconnect notification",
             "showDisconnect": "Show disconnect notification",
-            "reconnectNotificationDelay": "Reconnect notification delay (sec)"
+            "notificationDisplayTime": "notification display time (seconds)"
         },
         "layout": {
             "pages": "Pages",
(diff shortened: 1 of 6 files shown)


joepavitt:
The use of `localStorage` is a bit tricky. The `main.mjs` is firing first, before `App.vue` where the LocalStorage is set, but this wouldn't have been noticed as the setting would have been saved/applied to the local storage in a previous refresh.

Instead, we should be utilising the vuex store here.

I'll make the relevant modifications that build on this great foundation

joepavitt:
@cgjgh thanks again for your work on this, checkout https://github.com/FlowFuse/node-red-dashboard/pull/1490/commits/80f3543c36480710a3f3a0cbaf6478308a9afc06 for the changes I've made in transferring this into our state store, rather than the local storage

joepavitt (approved):

cgjgh:
Ok I see.. I initially thought that the Vuex store wasn't updated with the dashboard config at the time the socket events were set up, which is why I resorted to using localStorage for applying notification visibility settings.

> this wouldn't have been noticed as the setting would have been saved/applied to the local storage in a previous refresh

I was aware that the notification settings being applied were from a previous page load, but didn't see the better option at the time.

Anyway, I'm glad your experienced eye saw the Vuex alternative.

Also, if rewriting these PRs is too time-consuming, feel free to leave advice or instructions on what to change, and I can assist with the modifications.

PR #1492, opened 2024-11-22
Slider text field

### LINKED ISSUE #1063: UI Slider - Text Field Input
### Description

https://discourse.nodered.org/t/numeral-entry-in-dashboard-2-slider/89256

Available as a property in Vuetify's slider widget: https://vuetifyjs.com/en/components/sliders/#append-text-field

cgjgh:
## Description

Adds an optional text field for number input next to slider.

https://github.com/user-attachments/assets/ba271903-2645-4501-bfb1-ca904e58589c

## Related Issue(s)

Closes #1063
## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Seeing 2 x E2E failures in the `slider.spec.js` that'll need reviewing please @cgjgh - otherwise, this is an excellent addition

cgjgh:
@joepavitt Yes, I'm aware of it, I added two tests for input max/min validation, they work using `cypress open` but fail with `cypress run` - seems text field blur is not firing consistently... any ideas?

joepavitt:
Thanks as always @cgjgh - another great addition, although, whilst it does bind well with the slider when changing from the slider side, two issues spotted when changing the value directly in the text input:

- It doesn't emit an event back into Node-RED when only changing the text field
- If you have steps in place, e.g. 0-10 with steps of 2, and you put in a number that isn't valid because the steps, but is valid because of the range - the slider will update correctly, but the text-field doesn't correct itself

I think you may need to bind the `text-field` to a different model value, then handle the events accordingly.

joepavitt (changes requested):
Thanks as always @cgjgh - another great addition, although, whilst it does bind well with the slider when changing from the slider side, two issues spotted when changing the value directly in the text input:

- It doesn't emit an event back into Node-RED when only changing the text field
- If you have steps in place, e.g. 0-10 with steps of 2, and you put in a number that isn't valid because the steps, but is valid because of the range - the slider will update correctly, but the text-field doesn't correct itself

I think you may need to bind the text-field to a different model value, then handle the events accordingly.

cgjgh:
@joepavitt I believe this is ready for review now. I hope my contributions, despite my limited experience, are more help than hassle.

joepavitt (approved):

joepavitt:
Looks great - just waiting for one final E2E test pass

PR #1531, opened 2024-12-12
Bump express to pick up path-to-regex

hardillb:
closes https://github.com/FlowFuse/node-red-dashboard/security/dependabot/47
## Description

fixes dependency

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/security/dependabot/47


hardillb:
There may be a follow up fix when NR pushes a new version

Steve-Mcl (approved):

Issue #1534, opened 2024-12-12
UI Group not loading


aitor-develop:
### Current Behavior

After selecting the page using the left-side toolbar group is not loading. 
![image](https://github.com/user-attachments/assets/6d2bc876-5732-4643-967f-a814479ffd0b)

If the group "TAB" is manually selected, then it displays correctly:
![image](https://github.com/user-attachments/assets/4d70a5b0-90e3-42f7-858e-96561edeb088)

The used layout is TABs.

### Expected Behavior

I expected to see the page with all its widgets loaded correctly when I select in the left-side toolbar.

### Steps To Reproduce

1) Select the page
![image](https://github.com/user-attachments/assets/f6c47eb3-caf4-4b5a-a88e-2887f62103d9)

2) Move to another group in the same page that is not the first one
![image](https://github.com/user-attachments/assets/fb492a26-395c-4b9d-be21-dce1467a70b3)

3) Select a new page with the same display "TAB" from a group that is not the first one
2) Wait to open it and see there is nothing.
![image](https://github.com/user-attachments/assets/d4b7fe80-2f83-4c47-8e6c-a228045aaed5)

3) Click on the group name to see content.
![image](https://github.com/user-attachments/assets/7d052b29-82ad-472e-a02f-41dd15585dcf)

### Environment

- Dashboard version: 1.20.0
- Node-RED version:  4.0.2
- Platform/OS: Dockerized NodeRED
- Browser: Brave v1.73.97

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

Steve-Mcl:
Can you please provide a minimal flow that demonstrates this issue?

bartbutenaers:
@Steve-Mcl ,
You can assign this one to me.  Will try to have a look tonight...
Bart

Steve-Mcl:
> @Steve-Mcl , You can assign this one to me. Will try to have a look tonight... Bart

Your wish is my command 😉

bartbutenaers:
@aitor-develop 
Thanks a lot for the (very) extensive description and example flow.
Being able to reproduce the problem fast, really helps to be able to help.
You were lucky that the wife and kids were still asleep, so I could have a look at it 😄 

See [pull request](https://github.com/FlowFuse/node-red-dashboard/pull/1552)

aitor-develop:
Hello, I'd like to re-open this issue, I recently updated the dashboard to **1.26.0** and this same issue is back.
[...]

PR #1552, opened 2024-12-21
Group tabs - select first tab after navigation

bartbutenaers:
## Description

When use group tabs, the first tab is correctly loaded at the start.  However when you select another tab, navigate to another page and back to the original page, then no tab is selected anymore.  And as a result the page only shows the tabs but the tab content area is empty.

This PR makes sure the first tabsheet is always selected, every time you navigate to the page with the tabs:

select-first-tab

## Related Issue(s)

[1534](https://github.com/FlowFuse/node-red-dashboard/issues/1534)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/layouts/Tabs.vue
+++ b/ui/src/layouts/Tabs.vue
@@ -70,6 +70,14 @@ export default {
         WidgetGroup
     },
     mixins: [Responsiveness],
+    beforeRouteEnter (to, from, next) {
+        next(vm => {
+            // Select the first tabsheet every time the user arrives on this page
+            if (vm.orderedGroups && vm.orderedGroups.length > 0) {
+                vm.tab = 0
+            }
+        })
+    },
     data () {
         return {
             tab: 0


joepavitt (approved):

Paul-Reed:
Could this be looked at again please, as it creates an annoying problem, particularly for phone/tablet users.
If I navigate to a tab to perhaps monitor energy usage, my phone screen blanks after 30 seconds, so I touch the screen to wake it up. PWA tells me that it's reconnecting, and then goes back to my first tab (as per this PR).
Then of course I have to navigate back to the energy tab to continue monitoring...
Why can't it load again on the same tab?

bartbutenaers:
@Paul-Reed 
Could you create a separate issue please?
Because otherwise it will be overlooked in the flood of other issues.
And can you there also explain *"as per this PR"*.  Do you mean that this PR caused that problem?

Paul-Reed:
https://github.com/FlowFuse/node-red-dashboard/issues/1604 created.

> Do you mean that this PR caused that problem?

I have been wrong before Bart :smirk:

Issue #1550, opened 2024-12-20
Colon in group name prevents working of group visibility


rucksman:
### Current Behavior

I am injection the following JSON into a ui-control:
```
{
    "groups": {
        "hide": [
            "MyPage:My: Second Group"
        ]
    }
}
```
This throws an error:
```
No item with the name 'My' found
```
Tried different escape methods like quotes and backslash for the colon in my group name, but nothing worked.

### Expected Behavior

Group name with a colon should be processed correctly.

### Steps To Reproduce

Create a group with a colon in its name and try to show/hide it via ui-control.

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

bartbutenaers:
@rucksman,
If you could share a very small example flow to reproduce this, then I will try to find some time to look at this.  I haven't used the ui-control myself yet...

rucksman:
Here you go ... You can basically ignore the text node, I only used it to create the page and group. Just inject hide group and you will see the error in the error 

> No item with the name 'This is' found

```
[
    {
        "id": "a663bb334b6ef130",
        "type": "tab",
        "label": "Bug ui-control",
        "disabled": false,
        "info": "",
        "env": []
    },
    {
        "id": "381db98ea173ca41",
        "type": "inject",
        "z": "a663bb334b6ef130",
        "name": "show group",
        "props": [
            {
                "p": "payload"
            },
            {
                "p": "topic",
                "vt": "str"
            }
        ],
        "repeat": "",
        "crontab": "",
        "once": false,
        "onceDelay": 0.1,
        "topic": "",
        "payload": "{\"groups\":{\"show\":[\"Test Page:This is: My Test Group\"]}}",
        "payloadType": "json",
        "x": 110,
        "y": 80,
        "wires": [
            [
                "0a32d04d05f5bdc5",
                "558baa9d92a21c55"
            ]
        ]
    },
    {
        "id": "0a32d04d05f5bdc5",
        "type": "debug",
        "z": "a663bb334b6ef130",
        "name": "debug show group",
        "active": true,
        "tosidebar": true,
        "console": false,
        "tostatus": false,
        "complete": "true",
        "targetType": "full",
        "statusVal": "",
        "statusType": "auto",
        "x": 290,
        "y": 40,
        "wires": []
    },
    {
        "id": "4f8b8c9d230925dc",
        "type": "inject",
        "z": "a663bb334b6ef130",
        "name": "hide group",
        "props": [
[...]

PR #1554, opened 2024-12-21
ui-control colon in group name

bartbutenaers:
## Description

When a group name contains a colon `":"`:

image

Then you should be able to _**escape**_  (using backslashes) that colon for the ui-control node like this:

image

Otherwise it will trigger an error that group name *"This is"* cannot be found...

This PR splits only by non-escaped colons, and removes the escape characters afterwards.  Then it seems to work:

groupname_colon

## Related Issue(s)

[1550](https://github.com/FlowFuse/node-red-dashboard/issues/1550)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/nodes/widgets/ui-control.md
+++ b/docs/nodes/widgets/ui-control.md
@@ -123,6 +123,7 @@ msg.payload = {
 ```
 
 _Note:_ `pages` can be subbed with `tabs` as per Dashboard 1.0 and `groups` can also be subbed with `group` as per Dashboard 1.0.
+_Note:_ when the `group` name contains a colon `:`, then that should be escaped using a backslash `\`.
 
 ### Enable/Disable
 
--- a/nodes/widgets/locales/en-US/ui_control.html
+++ b/nodes/widgets/locales/en-US/ui_control.html
@@ -33,7 +33,7 @@
     }
 }</pre>
     <h4>Enable/Disable Pages & Groups</h4>
-    <p>Dashboard pages & groups can be disabled & re-enabled by sending a <code>msg.payload</code> object with the format
+    <p>Dashboard pages & groups can be disabled & re-enabled by sending a <code>msg.payload</code> object with the format</p>
     <pre>msg.payload = {
     pages: {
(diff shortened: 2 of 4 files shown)


Steve-Mcl:
Thanks for your support, Bart. ❤️

joepavitt:
Documentation you've added shows to escape it with `\`, but your example escapes with `\\`?

Steve-Mcl:
> Documentation you've added shows to escape it with `\`, but your example escapes with `\\`?

It'll be because it's inside JSON I imagine. 

As for the whole escape using backslash - I have reservations (but cannot put my finger on why). Things like CSV usually just double up the character `::` or html use an alt code. 

Ideally, We would support better structured property for specifying the page (not just a string joined by a random colon) but I understand this is a throwback from dashboard 1.   In fact, I've always been super curious why we don't support the node ID. There is no ambiguity and if I ever change the title of the page, my UI control still works.

bartbutenaers:
Hi guys,
Quick response because familly stuff going on.

1. Yes like Steve says, escape is single `\` but you need extra `\` to get a valid json.  I will need to correct the docs later on.
2. Indeed while creating a fix, I indeed had a bad feeling.  Not about the fix, but about the long string path.  A separate `page` and `group` property is nicer imho. 
3. Yes the node id makes sense.  Perhaps as an alternative way of working?

joepavitt:
> Yes the node id makes sense. Perhaps as an alternative way of working?

Yeah, I think this is the better solution, but we do need to support the existing methodology for backward compatibility. Maybe we `getNode` given the string first, if no node is returned, then assume it's page/group name combo?

Getting a Node's ID in Node-RED is also quite unintuitive though (especially for config nodes), so would need to improve the docs on this.

bartbutenaers:
@joepavitt 
I will try to have a look at this during the weekend.

bartbutenaers:
Closed in favor of successor pull request (see [here](https://github.com/FlowFuse/node-red-dashboard/pull/1583))

Issue #1505, opened 2024-11-28
Can't open WYSIWYG editor


hulzer:
Hi,

I'd like to test the new layout editor feature, but I can't get it working.

When I try to a new page, I can't select the wysiwyg layout from dropdown:
![image](https://github.com/user-attachments/assets/0edc7371-5bc2-4503-9e3a-4dc100487800)

node-red v4.0.2
node-red-dasboard v1.9.1

Thanks.

joepavitt:
As @colinl has mentioned, it is available in the sidebar, but we could add something into that config tray too. Will keep this issue open in order to capture that.

bartbutenaers:
Hi everybody,

I wanted to have a look at this issue, but I don't fully understand how it works.

Suppose a page has layout type xxx (e.g. notebook, ...).  You open the layout in wysiwyg editor, and deploy your changes after chaning the layout.  Then the layout - which arrives in the frontend - still seems to be the original type xxx.

So it looks to me that there is no dedicated *"wysiwyg layout"* type?

Thanks for illuminating me!
Bart

joepavitt:
> So it looks to me that there is no dedicated "wysiwyg layout" type?

That's correct, there was in our original PoC, but byb the time it made it to `main`, @Steve-Mcl had done some great work so that actually, all of the existing layouts are editable through a WYSIWYG editor.

So, a button alongside the layout type editor of "Edit", could then open that WYSIWYG editor for the page in the browser? I had also considered an "Edit Layout" button at the top of the tray, alongside where the existing "Delete" button is, but that could be available for _every_ node - making it easier to jump into layout editing from any point.

PR #1568, opened 2025-01-09
ui-page: layout editor button

bartbutenaers:
## Description

This PR shows a button in the ui-page config screen, to open the layout manager:

image

~~The _**alignment**_ of the button isn't really nice, because it would look better if it was right-aligned with all other elements in the config screen.  But there seems to be quite some CSS going on. Would be nice if anybody could tweak it a bit better.~~

***[EDIT 10/01/2025]*** The alignment problem is now solved, thanks to @GogoVega :

image

BTW I have used a ***global function***, so that the same function can be reused in the ui-page config screen.  I didn't want to duplicate code, because that will run out of sync after a while anyway.  I tried to put it as an extra custom function in the ui-base type object, but Node-RED seems to strip off all functions that it doesn't expect.  I used quite a long name for the global function, in order to avoid conflicts with a future new global function with the same name.  If you have better idea than a global function to accomplish this, please let me know and I will adjust it. 

## Related Issue(s)

[1505](https://github.com/FlowFuse/node-red-dashboard/issues/1505)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


nodes/config/ui_page.html
```diff
@@ -277,6 +291,9 @@
     <div class="form-row">
         <label for="node-config-input-layout"><i class="w-16 fa fa-th"></i> <span data-i18n="ui-page.label.layout"></label>
         <input type="text" id="node-config-input-layout">
+        <button id="node-config-layout-edit" type="button" class="red-ui-button" style="margin-top: 4px; margin-right: 5px;" title="Edit Layout">
+            <i class="fa fa-pencil-square-o"></i>
+        </button>
         <input type="hidden" id="node-config-input-layoutType">
```
GogoVega, line 297: For the alignment:
```suggestion
        <div style="width: 70%; display: inline-flex;">
            <div style="flex-grow: 1;">
                <input type="text" id="node-config-input-layout" style="width: 100%;">
                <input type="hidden" id="node-config-input-layoutType">
            </div>
            <button type="button" id="node-config-layout-edit" class="red-ui-button" style="margin-left: 10px;" title="Edit Layout">
                <i class="fa fa-pencil-square-o"></i>
            </button>
        </div>
```

Don't forget to translate the title (hardcoded here) 😉

bartbutenaers: @GogoVega 
That looks indeed very nice that way:

image

Really appreciated!!

> Don't forget to translate the title (hardcoded here)

Yes indeed I had also considered to translate the title, but there were other untranslated titles.  So I was thinking to myself: enough higher priority stuff to day, and Rome wasn't build in one day neither :innocent:

Issue #1588, opened 2025-01-23
Error Accessing Dashboard with FlowFuse Authentication Enabled


sumitshinde-84:
### Current Behavior

When I try to access the dashboard page while flowfuse authentication is enabled getting following error page, for more information read this slack thread : https://flowforgeworkspace.slack.com/archives/C03FAUBPPJ7/p1737623419750459

![Image](https://github.com/user-attachments/assets/afcea826-9fdc-4077-9a90-4e7148930acf)

### Expected Behavior

It should open the dashboard page with a form to enter credentials, or, if already logged in, display the expected dashboard UI.

### Steps To Reproduce

1. Enabled FlowFuse User Authentication and save settings
2. Restart instance 
3. Try to access the dashboard

### Environment

- Dashboard version: 1.22.0
- Node-RED version: 4.0.8
- Node.js version: 20.18.1
- npm version: 
- Platform/OS: FlowFuse Cloud (Linux 5.10.217-205.860.amzn2.aarch64 arm64 LE)
- Browser: Safari  

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

robmarcer:
This is stopping a customer (https://app-eu1.hubspot.com/contacts/26586079/record/0-2/10868925937) from using one of our Blueprints. They were planning to roll this out to their UK factory on Monday. It would be very good for their confidence in FF if we can maintain that timeline.

joepavitt:
Investigating now

joepavitt:
So, we know the cause - it's due to the way that Safari caches the auth session. 

The session token expires (either due to timeout, or due to an underlying instance restart). When a user navigates to the page, Safari goes to validate the session variables against FF Auth, but due to an issue with Safari and the `allow access origin: *`, the CORS check fails, which prevents Dashboard from getting any new content or session variables.

More details here: https://medium.com/prodigy-engineering/is-safari-getting-stricter-with-cors-8767c809c726#:~:text=If%20you%20have%20an%20issue,control%2Dallow%2Dorigin%20domains

knolleary:
I have installed MS Edge on MacOS and can reproduce the error (as the customer has reported this on Edge rather than safari).

To reproduce, given a NR instance serving dashboard with FF Auth enabled, start with a clean browser session having cleared the cache/cookies for the NR instance.

1. Login to FF App
2. Open the Dashboard in Edge/Safari - dashboard should load okay
[...]

PR #1616, opened 2025-02-11
Dashboard: Allow control over whether it's possible to install Dashboard as a PWA

joepavitt:
## Description

- Adds a new "Allow Installation" option to a Dashboard
- Will toggle on/off the loading of the `manifest.yml` from the Dashboard, which in turn permits installation

## Related Issue(s)

Closes https://github.com/FlowFuse/node-red-dashboard/issues/1588

### DIFF
--- a/docs/nodes/config/ui-base.md
+++ b/docs/nodes/config/ui-base.md
@@ -38,7 +38,13 @@ _Example of how the 'App Icon' looks on a browser tab_
 ![Example of how the 'App Icon' looks](/images/node-examples/ui-base-app-icon-launcher.png "Example of how the 'App Icon' looks"){data-zoomable style="max-width: 200px; margin: auto"}
 _Example of how the 'App Icon' appears when installed onto an iPhone_
 
-## Title Bar Style Options <AddedIn version="1.10.0" />
+## Allow Install <AddedIn version="1.23.0" />
+
+It is possible to install any FlowFuse Dashboard applications as a Progressive Web App onto a desktop or mobile device, giving the appearance and user-experience of a native mobile application. You can read more about that functionality [here](../../user/pwa.md).
+
+The "Allow Install" option provides the capability to enable/disable the option to "Install" your Dashboard as an application. If you are using FlowFuse authentication, this is recommended, as the PWA workers cache content separate from authentication sessions.
+
+## Header Style Options <AddedIn version="1.10.0" />
 
 ### Title Bar - Default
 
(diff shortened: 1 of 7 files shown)


nodes/config/locales/en-US/ui_base.html
```diff
@@ -1,27 +1,66 @@
 <script type="text/html" data-help-name="ui-base">
-    <p>UI Base</p>
+    <a href="https:/ddashboard.flowfuse.com/nodes/config/ui-base.html" target="_blank" style="display: block; margin-bottom: 6px;">Online Documentation</a>
```
knolleary, line 2:
```suggestion
    <a href="https://dashboard.flowfuse.com/nodes/config/ui-base.html" target="_blank" style="display: block; margin-bottom: 6px;">Online Documentation</a>
```

knolleary:
I still see the `<link rel="manifest"...>` is present in `dist/index.html` when I build this branch. 

Once I comment that out, then I no longer get the auth errors from #1588 - I think.

docs/nodes/config/ui-base.md
```diff
@@ -38,7 +38,13 @@ _Example of how the 'App Icon' looks on a browser tab_
 ![Example of how the 'App Icon' looks](/images/node-examples/ui-base-app-icon-launcher.png "Example of how the 'App Icon' looks"){data-zoomable style="max-width: 200px; margin: auto"}
 _Example of how the 'App Icon' appears when installed onto an iPhone_
 
-## Title Bar Style Options <AddedIn version="1.10.0" />
+## Allow Install <AddedIn version="1.23.0" />
+
+It is possible to install any FlowFuse Dashboard applications as a Progressive Web App onto a desktop or mobile device, giving the appearance and user-experience of a native mobile application. You can read more about that functionality [here](../../user/pwa.md).
+
+The "Allow Install" option provides the capability to enable/disable the option to "Install" your Dashboard as an application. If you are using FlowFuse authentication, this is recommended, as the PWA workers cache content separate from authentication sessions.
```
knolleary, line 45:
```suggestion
The "Allow Install" option provides the capability to enable/disable the option to "Install" your Dashboard as an application. If you are using FlowFuse authentication, this is not recommended, as the PWA workers cache content separate from authentication sessions.
```

nodes/config/ui_base.js
```diff
@@ -80,6 +80,10 @@ module.exports = function (RED) {
             config.notificationDisplayTime = 5 // Show for 5 seconds
         }
 
+        if (!('allowInstall' in config)) {
+            config.allowInstall = true // Show for 5 seconds
```
knolleary, line 84:
```suggestion
            config.allowInstall = true
```

knolleary (commented):
Appears to work in my local testing, although can be tricky to be 100% sure when working with localhost and http endpoints.

I've proposed a change to clarify we recommend *not* enabling this feature if you are using FF Auth. It does raise the question on whether we want it enabled by default or not. As it stands, this PR enables it by default. We could take the approach of disabling by default for *new* instances and only default for true for existing instances .

Steve-Mcl (commented):
Tested locally by firing up existing dashboard with new code.  Unfortunately the check is not enabled by default and the app is still installable.

image

Steve-Mcl:
> Tested locally by firing up existing dashboard with new code. Unfortunately the check is not enabled by default and the app is still installable.
> 
> image

Fixed in [d666bc9](https://github.com/FlowFuse/node-red-dashboard/pull/1616/commits/d666bc9feb24a1b29f8fdd59952aaf11b2a4e844)

Additionally, raised PR #1665 to cover other NG compatibility checks

Steve-Mcl (approved):
Locally tested - working as expected

PR #1651, opened 2025-04-02
Docs: Ensure we have an appropriate example timestamp in the Chart docs

### LINKED ISSUE #1640: Issue with Chart documentation - Short timestamp, seconds, prevents data from rendering.
### Current Behavior

When the X-Axis Type is set to Timescale, putting a 10 digit, seconds, Epoch time in the data objects prevents the series from plotting, although the timestamps and series labels show up!

Image
This works... dropping the last 3 0's breaks it.

From the documentation...

Image

### Expected Behavior

The fact that the timestamps show up, just makes it all the more confusing when following the example code in the docs. Given that the one example of Epoch time use was clearly designed to stipulate 10 digits, I expected that to work!

Could the node not just pad the timestamp if it is too short? 
Its not like many folk are going to want mS precision on a chart, at least not in typical applications. Also Influx, a common data source I expect, seems to work better at 1S precision, at least when using numeric timestamps, which are easier to handle when building dynamic queries!

### Steps To Reproduce

Set a line/scatter chart to X Timescale, as per docs and feed it this or something similar...
[Data.txt](https://github.com/user-attachments/files/19438102/Data.txt)

(E1cid pointed out the issue and fix... Hero!)

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:
RE. All above... Sorry, Couldn't say, and don't know how to find out.
However all components updated on the day of posting THIS, in an attempt to fix the issue!

joepavitt:
## Description

In my lazy attempt to write a timestamp using 1 -> 0, I accidentally typed out a legitimate timestamp in seconds from Epoch, but this led to confusion around scale/units.

Instead, I've put in a legitimate timestamp to ensure this confusion doesn't occur going forward.

## Related Issue(s)

Closes #1640

### DIFF
--- a/docs/nodes/widgets/ui-chart.md
+++ b/docs/nodes/widgets/ui-chart.md
@@ -136,6 +136,8 @@ Then, the last piece of the puzzle would be to set the `y` property would be one
 - If your data is a simple numerical value, you can leave this blank, and the chart will automatically use the value of `msg.payload`.
 - If your data is an object, you can provide the key of the value in your data, e.g. `{"myTime": 1743608192522, "myValue": 123}` would set the "Y" property to the type `key` and the value `myValue`.
 
+Note: Timestamps must be in milliseconds (ms). This follows the JavaScript standard for representing time as the number of milliseconds elapsed since the Unix epoch (January 1, 1970, 00:00:00 UTC).
+
 #### Interpolation Methods for Line Charts
 
 Interpolation defines how the line is drawn between data points on a line chart. In Dashboard, you can choose between several interpolation methods to suit your data visualization needs:
--- a/docs/nodes/widgets/ui-chart.md
+++ b/docs/nodes/widgets/ui-chart.md
@@ -129,12 +129,12 @@ A very common use case of Node-RED is to process timeseries data, such as sensor
 The value for the `x` property would then be one of two things:
 
 - If your data is a simple numerical value, you can leave this blank, and the chart will automatically use the current date/time.
-- If your data is an object, you can provide the key of the timestamp in your data, e.g. `{"myTime": 1234567890}` would set the "X" property to the type `key` and the value `myTime`.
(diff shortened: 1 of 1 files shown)


docs/nodes/widgets/ui-chart.md
```diff
@@ -129,12 +129,12 @@ A very common use case of Node-RED is to process timeseries data, such as sensor
...
 
 - If your data is a simple numerical value, you can leave this blank, and the chart will automatically use the current date/time.
-- If your data is an object, you can provide the key of the timestamp in your data, e.g. `{"myTime": 1234567890}` would set the "X" property to the type `key` and the value `myTime`.
+- If your data is an object, you can provide the key of the timestamp in your data, e.g. `{"myTime": 1743608192522}` would set the "X" property to the type `key` and the value `myTime`.
 
 Then, the last piece of the puzzle would be to set the `y` property would be one of two options:
 
 - If your data is a simple numerical value, you can leave this blank, and the chart will automatically use the value of `msg.payload`.
-- If your data is an object, you can provide the key of the value in your data, e.g. `{"myTime": 1234567890, "myValue": 123}` would set the "Y" property to the type `key` and the value `myValue`.
+- If your data is an object, you can provide the key of the value in your data, e.g. `{"myTime": 1743608192522, "myValue": 123}` would set the "Y" property to the type `key` and the value `myValue`.
```
Steve-Mcl, line 137:
```suggestion
- If your data is an object, you can provide the key of the value in your data, e.g. `{"myTime": 1743608192522, "myValue": 123}` would set the "Y" property to the type `key` and the value `myValue`.

Note: Timestamps must be in milliseconds (ms). This follows the JavaScript standard for representing time as the number of milliseconds elapsed since the Unix epoch (January 1, 1970, 00:00:00 UTC).
```

Suggestion: clarify what kind of timestamp

Steve-Mcl (approved):
I have pre-approved however I feel it warrants a note.

I have made a suggestion that we add a note explaining what the timestamp is expected to be

PR #1663, opened 2025-04-08
Handle trailing slashes in the URL

### LINKED ISSUE #1476: URL with trailing "/"
### Current Behavior

"www.example.com/dashboard/mypage"  works but "www.example.com/dashboard/mypage/" does not work.

When nginx is used as a reverse proxy, it adds a trailing "/" , I imagine that the trailing "/" could be an issue with other configurations.

### Expected Behavior

Ignore trailing "/".

### Steps To Reproduce

_No response_

### Environment

- Dashboard version:  1.19.1
- Node-RED version: 4.0.5: Maintenance Release Editor
- Node.js version:
- npm version:
- Platform/OS: Linux
- Browser: FireFox 132.0.2 (64-bit)

I can not provide an initial effort estimate

Steve-Mcl:
closes #1476

## Description

After some consideration, we will stick with slash-less URLs like /page1, and enforce that style with a 301 redirect middleware. 
It’s cleaner, SEO-friendly, backward-compatible, and avoids unnecessary rewrites.

## Related Issue(s)

#1476

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -126,6 +126,23 @@ module.exports = function (RED) {
 
             uiShared.app.use(config.path, uiShared.httpMiddleware, express.static(path.join(__dirname, '../../dist')))
 
+            // Redirect any requests with a trailing slash to the same path without the trailing slash
+            uiShared.app.use((req, res, next) => {
+                // Skip if not a GET or HEAD request
+                if (!(req.method === 'GET' || req.method === 'HEAD')) {
+                    next()
+                    return
+                }
+
+                if (req.path.slice(-1) === '/' && req.path.length > 1) {
+                    const query = req.url.slice(req.path.length)
+                    const safePath = req.path.slice(0, -1).replace(/\/+/g, '/')
+                    res.redirect(301, safePath + query)
+                } else {
+                    next()
+                }
+            })
+
             uiShared.app.get(config.path + '/_setup', uiShared.httpMiddleware, (req, res) => {
                 let socketPath = join(RED.settings.httpNodeRoot, config.path, 'socket.io')
                 // if no leading /, add one (happens sometimes depending on httpNodeRoot in settings.js)
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -126,18 +126,30 @@ module.exports = function (RED) {
 
             uiShared.app.use(config.path, uiShared.httpMiddleware, express.static(path.join(__dirname, '../../dist')))
 
-            // Redirect any requests with a trailing slash to the same path without the trailing slash
+            // Middleware to enforce canonical URLs without trailing slashes.
+            //
+            // In some configurations (e.g., with NGINX), requests might include
+            // a trailing slash. For consistency, SEO benefits, and to reduce duplicate
+            // content issues, we redirect such URLs to their non-trailing version.
+            //
+            // This middleware:
+            // - Applies only to GET and HEAD requests.
+            // - Ignores the root path ("/").
+            // - Redirects (301) URLs ending with a trailing slash to their non-trailing equivalent.
+            // - Preserves the query string.
+            // - Collapses any redundant slashes (e.g., "//" becomes "/").
             uiShared.app.use((req, res, next) => {
                 // Skip if not a GET or HEAD request
                 if (!(req.method === 'GET' || req.method === 'HEAD')) {
                     next()
                     return
                 }
-
+                // Check if someone is trying to open a page with with a / at the end
+                // If yes, redirect (with a 301 status) to the correct page without the trailing slash.
                 if (req.path.slice(-1) === '/' && req.path.length > 1) {
-                    const query = req.url.slice(req.path.length)
-                    const safePath = req.path.slice(0, -1).replace(/\/+/g, '/')
-                    res.redirect(301, safePath + query)
+                    const query = req.url.slice(req.path.length) // get the query string (if any)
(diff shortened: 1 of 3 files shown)


Steve-Mcl:
@robertsLando @Bond246 @rsp92028 
Would the changes in this PR satisfy your issue?


The important part of the change is here: 
https://github.com/FlowFuse/node-red-dashboard/pull/1663/files#diff-76241199d5a97f1b00433a720ea273842ca6515d31015b1b24aee147e2a975bdR137-R143


In effect, what it does is:

1. It checks if someone is trying to open a page with with a `/` at the end
2. It then redirects (with a 301 status) to the correct page without the trailing slash.

nodes/config/ui_base.js
```diff
@@ -126,6 +126,23 @@ module.exports = function (RED) {
...
+                    return
+                }
+
+                if (req.path.slice(-1) === '/' && req.path.length > 1) {
+                    const query = req.url.slice(req.path.length)
+                    const safePath = req.path.slice(0, -1).replace(/\/+/g, '/')
+                    res.redirect(301, safePath + query)
+                } else {
+                    next()
+                }
```
robertsLando, line 143: I suggest to add a comment here mentioning the underlying issue that this fixes as it's not easy to remember the logic that requires this :)

robertsLando (approved):
Didn't tested this on my end but code side looks good

nodes/config/ui_base.js
joepavitt: Agree with adding a comment here, otherwise, this looks good to go

joepavitt: @Steve-Mcl can you action this please? Then we can get a new Dashboard release out.

Steve-Mcl: Done Joe.

joepavitt (approved):

dceejay:
@Steve-Mcl @joepavitt - please see this thread - https://discourse.nodered.org/t/flowfuse-dashboard-2-0-issue-err-too-many-redirects/97107/6

I also saw this too_many_redirects loop when using my worldmap node (which sets up it's own endpoint also) but never figured out why...

dceejay:
Hi - I was using it within an iframe so should was just set to /worldmap - but on one instance it went into that loop and wouldn't come out.. See also https://discourse.nodered.org/t/worldmap-dashboard-2-0-iframe-example-error-too-many-redirects/97123 for another user with same issue

Bond246:
For me the problem still exists.
I use traefik reverse  proxy with an "addprefix" middleware that redirects the base-path "/" to "/dashboard/some-page". 
Before the 1.24er dashboard-version there was just a problem with trailing slashes and traefik said it can't reach `/dashboard/some-page/`. 

Now with the new version the browser-path is updated to `/dashboard/some-page` and traefik try to reach `/dashboard/some-page/dashboard/some-page`... obviously not working...

Steve-Mcl:
@Bond246 is this not an issue with your setup? 

You state that **"addprefix" middleware that redirects the base-path "/" to "/dashboard/some-page".** so why would your configuration be appending "dashboard/some-page" to "dashboard/some-page"?

All the fix in this PR does is to redirect (with a 302 status code) from `/dashboard/xxx/` to `/dashboard/xxx` (the redirect from `dashboard/` to `dashboard/first-page` was an existing feature)

Steve-Mcl:
@Bond246 could you share your traefik config (sanitise where necessary).

Steve-Mcl:
@Bond246 I am not up to speed on traefik but 

You currently have two middlewares:
* nodered-dashboard-page1: adds prefix /dashboard/page1
* nodered-dashboard-page2: adds prefix /dashboard/page2

BUT:
* They just add a prefix → e.g., `/` becomes `/dashboard/page1`,
* they don’t check the incoming path — they will blindly apply the prefix even if the request is already under `/dashboard/page1`.


This is why you’re seeing things like:

```
/dashboard/page1 → /dashboard/page1/dashboard/page1 → /dashboard/page1/dashboard/page1/dashboard/page1 ...
```

You could try Redirect only:
* `/`
* `/dashboard`
* `/dashboard/`

But leave:
* `/dashboard/page1`
* `/dashboard/page2`
* anything under `/dashboard/*`
untouched.


Something like:
```
apiVersion: traefik.io/v1alpha1
kind: Middleware
metadata:
  name: nodered-dashboard-redirect
  namespace: nodered
spec:
  redirectRegex:
    regex: ^(/dashboard)?/?$
    replacement: /dashboard/page1
    permanent: true
```


---

Why I suspect `addPrefix` is not right here: `addPrefix` always prepends the string, no matter what the incoming path is. AFAIK, it doesn’t check if the path already has it so can lead to repeated prefixes. I _think_ by switching to `redirectRegex`, you can conditionally match and rewrite only when necessary?

Bond246:
@Steve-Mcl you are correct, addPrefix always would add the path even if it already exists.
But i also could got to google.com/foo/bar and would get 404 becaus it doesn't exist. So in my case the user doesn't know and doesn't have to know that there is something like /dashboard/xyz because he stays in his single specific space under /dashboard/page1 and thats it. Everything he needs is there.

But i will try to test redirectRegex which in this case would make the path that is added visible. Not so nice in my opinion like addPrefix but would fix the issue.

Bond246:
`redirectRegex` worked well like it should.
Not so nice like `addprefix` where the path was not provided to the browser url-bar but ok for what i want to do with it.

PR #1685, opened 2025-04-25
Ensure we overflow by default if child widget extends beyond widget's defined height

joepavitt:
## Description

Screenshot 2025-04-25 at 21 00 33

If we set a height on a `ui-template` that's ignored if the child content is taller. This change ensures that the content instead scrolls to ensure consistency in the rendering of row heights in a Dashboard.


Steve-Mcl (approved):

knolleary:
This change appears to have had some unexpected side-effects:

User reported:

image

Also #1704 appears relevant as a change in overflow behaviour in 1.24.

PR #1700, opened 2025-05-09
Reconnect all clients when flows are redeployed

### LINKED ISSUE #870: Ui-Control behaviour upon Depoly
### Current Behavior

Hi

This is more a doubt regarding how it is suppose to work.

The node `ui-control` after we make a Node-Red `Deploy` does not output anything until we force a refresh on the windows where the dashboard is open.

https://github.com/FlowFuse/node-red-dashboard/assets/17281826/903ee9e2-c5b0-4a68-8b67-049dcb4ea232

Is this behaviour expected?
 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- Dashboard version: v1.9.1
- Node-RED version: 3.1.7
- Node.js version: 20.11.0
- npm version: 10.4.0
- Platform/OS: Server Linux; Client: Windows
- Browser: Edge

luisthieme:
## Description

Not a perfekt solution but it works.

## Related Issue(s)
 
Closes #870 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -835,10 +835,10 @@ module.exports = function (RED) {
                 cleanupEventHandlers(conn)
             }
 
-            for (const [id, socket] of uiShared.ioServer.sockets.sockets) {
-                socket.disconnect();
-            }            
-            
+            for (const socket of uiShared.ioServer.sockets.sockets.values()) {
+                socket.disconnect()
+            }
+
             close(node, function (err) {
                 if (err) {
                     node.error(`Error closing socket.io server for ${node.id}`, err)
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -834,6 +834,11 @@ module.exports = function (RED) {
             for (const conn of Object.values(uiShared.connections)) {
                 cleanupEventHandlers(conn)
             }
+
+            for (const [id, socket] of uiShared.ioServer.sockets.sockets) {
+                socket.disconnect();
+            }            
+            
             close(node, function (err) {
                 if (err) {
                     node.error(`Error closing socket.io server for ${node.id}`, err)
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -331,6 +331,15 @@ module.exports = function (RED) {
             return
         }
 
+        // disconnect all existing sockets (without force, and before removing all event handlers)
+        for (const socket of uiShared.ioServer.sockets?.sockets?.values()) {
+            try {
+                socket.disconnect()
+            } catch (_e) {
+                // ignore error  
+            }
+        }
+
         // determine if any ui-pages are left, if so, don't close the server
         const baseNodes = []
         const pageNodes = []
@@ -833,10 +842,6 @@ module.exports = function (RED) {
             uiShared.ioServer?.off('connection', onConnection)
             for (const conn of Object.values(uiShared.connections)) {
                 cleanupEventHandlers(conn)
-            }
-
-            for (const socket of uiShared.ioServer.sockets.sockets.values()) {
-                socket.disconnect()
             }
 
             close(node, function (err) {
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -332,11 +332,14 @@ module.exports = function (RED) {
         }
 
         // disconnect all existing sockets (without force, and before removing all event handlers)
-        for (const socket of uiShared.ioServer.sockets?.sockets?.values()) {
-            try {
-                socket.disconnect()
-            } catch (_e) {
-                // ignore error  
+        const sockets = uiShared.ioServer.sockets?.sockets
+        if (sockets && typeof sockets.values === 'function') {
+            for (const socket of sockets.values()) {
+                try {
+                    socket.disconnect()
+                } catch {
+                    // ignore error
+                }
             }
         }
 


joepavitt:
Thanks @luisthieme - we'll get this reviewed and hopefully merged this week

nodes/config/ui_base.js
```diff
@@ -834,6 +834,11 @@ module.exports = function (RED) {
             for (const conn of Object.values(uiShared.connections)) {
                 cleanupEventHandlers(conn)
             }
+
+            for (const socket of uiShared.ioServer.sockets.sockets.values()) {
+                socket.disconnect()
+            }
```
Steve-Mcl, line 840: This code should be in the `close` handler.

Also note, the close handler already has `uiShared.ioServer.disconnectSockets(true)` which does what you are doing here (but without any internal optimisations or considerations.)

That all said, what you wrote obviously works - so something is a-miss!

Potentially, because the `close` handler does other clean up, it is an "order of events" issue?

My suggestion is to move this loop into the `close` handler near the top (below the `if (!uiShared.ioServer) {` guard) around [line 333](https://github.com/FlowFuse/node-red-dashboard/pull/1700/files#diff-76241199d5a97f1b00433a720ea273842ca6515d31015b1b24aee147e2a975bdR333)

Also, please be defensive to permit following code to occur should there be an issue.  e.g, use conditional chaining `uiShared.ioServer.sockets?.sockets?.values()` and  wrap it in a `try..catch` to permit code below to run should there be an issue calling `disconnect` on any of the elements.

Steve-Mcl: Also, it may simply be a case of swapping these [2 lines](https://github.com/FlowFuse/node-red-dashboard/pull/1700/files#diff-76241199d5a97f1b00433a720ea273842ca6515d31015b1b24aee147e2a975bdR365-R366):

```
        uiShared.ioServer.disconnectSockets(true)
        uiShared.ioServer.removeAllListeners()
```

i.e. disconnect before removing listeners.

It might also be that we include the parameter `true` (force disconnection) and your loop does not. IDK.

Would you be able to re-test with this suggest swap as an alternative on your current setup to see if that handles it instead of looping all the sockets?

Steve-Mcl: Yes, I figured the forceful disconnection would be why feedback is not emitted to connected clients.

We could remove the `true` and that would likely work (when swapped of course) but lets stick with what you know works. Lets move your loop (a defensive version of it) to [line 333](https://github.com/FlowFuse/node-red-dashboard/pull/1700/files#diff-76241199d5a97f1b00433a720ea273842ca6515d31015b1b24aee147e2a975bdR333)

Something like:

```
function close (node, done) {
        if (!uiShared.ioServer) {
            done()
            return
        }
        // disconnect all existing sockets (without force, and before removing all event handlers)
        for (const socket of uiShared.ioServer.sockets?.sockets?.values()) {
            try {
                socket.disconnect()
            catch (_e) {
                // ignore error  
            }
        }
        ...
```

(untested)

luisthieme: I tested it, and it works as expected. Thank you for the fast and constructive feedback on this issue!

joepavitt: Just checking here, not seen any new commits into the branch/PR since the above was agreed upon. Which of you will be taking that action as it'd be nice to get this into a release today if we can.

luisthieme: I have already committed the agreed-upon solution. [commit](https://github.com/FlowFuse/node-red-dashboard/pull/1700/commits/4b9a4818287bf05040cc03e9ea3264a9ec919057)

Steve-Mcl:
Hmm, some odd, probably unrelated test fails...

```
  7 passing (17s)
  1 failing

  1) Node-RED Dashboard 2.0 - Buttons
       Button 1 (bool) outputs a bool payload & topic from msg.topic:

      Timed out retrying after 4000ms
      + expected - actual

      +'msg.topic from inject-1'
      
      at Context.eval (webpack://@flowfuse/node-red-dashboard/./cypress/support/commands.js:123:75)
```

```
  9 passing (29s)
  2 failing

  1) Node-RED Dashboard 2.0 - Control - Show/Hide
       can hide and show a particular page from the navigation options:

      AssertionError: Timed out retrying after 4000ms: Too many elements found. Found '3', expected '2'.
      + expected - actual

      -3
      +2
      
      at Context.eval (webpack://@flowfuse/node-red-dashboard/./cypress/tests/widgets/control.spec.js:91:48)

  2) Node-RED Dashboard 2.0 - Control - Enable/Disable
       can enable/disable a particular page from the navigation options:

      AssertionError: Timed out retrying after 4000ms: Not enough elements found. Found '2', expected '3'.
      + expected - actual

      -2
      +3
```

Steve-Mcl (approved):

cgjgh:
Has anyone else had an issue with this doing the opposite, (clients not reconnecting) or is it just me? Some of my clients do not reconnect unless I reload the page, whereas when removing this change, clients are reconnecting flawlessly.

joepavitt:
@cgjgh under what circumstances are your clients disconnecting in the first place?

joepavitt:
> All clients disconnect as before, but with new code some do not reconnect. I have not done any extensive testing, but just noticed that removing code made everything work flawlessly.

The initial bug report was, on a deploy, clients do not reconnect. Are you saying that has never been an issue for you?

So that I can push a release out, I will revert these changes. We'll do a more thorough review and then do a follow up release (hopefully) with this included.

cgjgh:
@joepavitt The referenced issue #870 specifically mentions ui-control behavior. I may have had this issue as well, and this PR _may_ have fixed this, but the behavior I was describing with this change seems to be a broader issue, as after deploys socket does not reconnect (correctly at least, as browser console outputs “SIO connected” but no config is being received as per usual). Reloading page reconnects socket correctly.

To sum it up, another issue seems to have been introduced. Will need further testing to fully understand it.

PR #1706, opened 2025-05-13
Contain overflow handling to template widgets

### LINKED ISSUE #1704: Slider thumb values not visible
### Current Behavior

There are various visual issues with the slider, [see this thread](https://discourse.nodered.org/t/slider-thumb-values-and-dashboard-1-24-0/97036)

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- Dashboard version: 1.24.0
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS: Rpi
- Browser: Firefox, Edge and Chrome

Steve-Mcl:
closes #1704

## Description

Contain overflow handling to template widgets (semi-reverts https://github.com/FlowFuse/node-red-dashboard/pull/1685)

## Related Issue(s)

#1704

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/layouts/Grid.vue
+++ b/ui/src/layouts/Grid.vue
@@ -115,7 +115,6 @@ export default {
         }
     },
     mounted () {
-        console.log('grid layout mounted')
         if (this.editMode) { // mixin property
             this.updateEditStateObjects()
             this.initializeEditTracking() // Mixin method
--- a/ui/src/layouts/Group.vue
+++ b/ui/src/layouts/Group.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="nrdb-layout-group--grid" :style="`grid-template-columns: repeat(min(${ columns }, var(--layout-columns)), 1fr); grid-template-rows: repeat(${group.height}, var(--widget-row-height)); `">
+    <div class="nrdb-layout-group--grid" :style="`grid-template-columns: repeat(min(${ columns }, var(--layout-columns)), 1fr); `">
         <div
             v-if="resizable" ref="group-resize-view" class="nrdb-resizable" :class="{'resizing': groupResizing.active}"
             :style="{ 'width': groupResizing.current.width ? `${groupResizing.current.width}px` : null, 'z-index': 99 }"
--- a/ui/src/layouts/Group.vue
+++ b/ui/src/layouts/Group.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="nrdb-layout-group--grid" :style="`grid-template-columns: repeat(min(${ columns }, var(--layout-columns)), 1fr); grid-template-rows: repeat(${group.height}, minmax(var(--widget-row-height), auto)); `">
+    <div class="nrdb-layout-group--grid" :style="`grid-template-columns: repeat(min(${ columns }, var(--layout-columns)), 1fr); grid-template-rows: repeat(${group.height}, var(--widget-row-height)); `">
         <div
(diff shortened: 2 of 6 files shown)


joepavitt:
Making some changes locally, turns out the "fix" in #1685 wasn't enough for templates anyway, so whilst the `overflow-y` was part of the fix, I'm having to modify how we set the CSS grid when rendering in a group.

I've also found that the `markdown` node also needs the `overflow-y`

joepavitt:
There are also issues with `textarea` option when using `ui-input`, but that seems to be a new beast as it's falling over with Vuetify's hardcoded `rows` field. We can't remove that option, it has to be set to _something_, working out the best option to best align with our layouts.

joepavitt:
Fixing one thing, finding another issue soon after - currently fixing `ui-form`

joepavitt:
This got very messy, but I have tested extensively, across all widgets, and I believe we now have the behaviour we want across them all. Most of the work was handling cases where `size` has been set to `auto`, at which point, content _should_ adapt it's height

Button groups and forms proved a challenge as the former can increase in height when wrapped, and the height of a form is defined by the number of fields too. In each case, where `auto` size is chosen, the heights adapt, but where an explicit height is defined, they'll resort to a scroll bar.

I've then made changes to `Group.vue` to provide cleaner CSS grid styling. A key missing piece was the definition of `gap` at the `nrdb-ui-widget` level, this meant a widget of "height" 3 was shorter than their parent group of height 3 as the parent considered the gap between the rows, the widget did not.

Steve-Mcl:
Joe, I pulled this to give it a once over with the flows of the dashboard demo site and I think, from my understanding of your changes, the ui-markdown node in auto sizing should expand to its content - is that right?

This is what I see:

chrome_ueKixzIWFa

joepavitt:
Hmmm... I did think I tested this on my machine, it's the classic case of _"it's working for me"_:

Screenshot 2025-05-14 at 09 17 56
Screenshot 2025-05-14 at 09 18 01

joepavitt:
I have also tested in Safari and it's working there too 🤔

joepavitt:
Interesting, when using the demo flows, it does break. Investigating the difference.

joepavitt:
Okay, the issue is coming from the group sizing. Somehow my local groups have `height` as `0`, which works nicely. For the demo dashboards, they have `height` of `1`. Given that we want sizing driven by the widgets, I'm going to remove the use of `height` from a group in the layout code.

joepavitt:
@Steve-Mcl fi pushed

joepavitt:
@Steve-Mcl FYI merging this now too, then we can get this in fix point release

joepavitt (approved):

Hugobox:
Hi @joepavitt ! I noticed my template stopped working after updating Dashboard 1.23.0 to 1.24.1
Not sure if its related to this pull request or not. Looking for your insight.
I basically have this flow where I feed an array of images in base64 to the following template but the images no longer appear.

```
<template>
    <div>
        <v-row>
        <span v-if="msg.payload" v-for="image in msg.payload.images" style="width:150px; display: block; padding: 5px;">
              <v-img  class="thumbnail" :width="150" :src="image.image"></v-img>
        </span>
        </v-row>
    </div>
</template>
```
Here's a simple flow to reproduce if it helps:
[ImagesTemplate.json](https://github.com/user-attachments/files/20510679/ImagesTemplate.json)



Do you think it is related? Thanks!

PR #1718, opened 2025-05-26
Revert "Reconnect all clients when flows are redeployed"

joepavitt:
Reverts FlowFuse/node-red-dashboard#1700 whilst we investigate [reported inconsistencies](https://github.com/FlowFuse/node-red-dashboard/pull/1700#issuecomment-2907534559) with the reconnecting behaviour

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -331,18 +331,6 @@ module.exports = function (RED) {
             return
         }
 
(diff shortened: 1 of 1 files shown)


ZJvandeWeg (approved):
Mostly approving considering the author. Didn't review the changes in depth but would like progress.

PR #1736, opened 2025-06-09
widget: Introduce the Progress Bar

ZJvandeWeg:
Through Vuetify a progress bar was already available, but it wasn't great to use and a bit mystical. With this change I'm trying to implement the widget in the _smallest_ way possible, so it can be extended later without backwards compat issues.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/widgets/locales/en-US/ui_progress.json
+++ b/nodes/widgets/locales/en-US/ui_progress.json
@@ -1,0 +1,12 @@
+{
+    "ui-progress": {
+        "label": {
+            "progress": "Progress",
+            "group": "Group",
+            "name": "Name",
+            "label": "Label",
+            "color": "Color",
+            "className": "Class"
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


joepavitt:
There could be merit in us adding this as a "Type" under the Gauge, rather than being it's own standalone widget?

ZJvandeWeg:
> There could be merit in us adding this as a "Type" under the Gauge, rather than being it's own standalone widget?

I expect a gauge to have fast moving data, and a progress bar to be more static. Also, the 'loading' part of the widget hasn't been added yet to keep the delta small, but is not something that the Gauge would support nicely.

joepavitt (changes requested):
### Required Changes

- No In-Editor Documentation
Screenshot 2025-06-23 at 16 27 37

- No Online Docs

joepavitt:
Screenshot 2025-06-23 at 16 33 52

For spacing/sizing, also need to consider the row height for the widget, and the empty space that this widget introduces. Should be, by default, will the progress bar to the full height (I think it would look odd), or at least vertically center it.

joepavitt:
We'll need to add E2E tests at some point, but happy to move forward for now

joepavitt (approved):

PR #1755, opened 2025-07-02
v1.25.0

joepavitt:
## Description

Version Bump

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@flowfuse/node-red-dashboard",
-    "version": "1.24.1",
+    "version": "1.25.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@flowfuse/node-red-dashboard",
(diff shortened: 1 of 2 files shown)


joepavitt:
@Steve-Mcl FYI - good to go, and have merged in https://github.com/FlowFuse/node-red-dashboard/pull/1754

Steve-Mcl (approved):
Approved but with a nod to the strange test fail:

```
  1) Node-RED Dashboard 2.0 - Forms
       Delivers topic from msg.topic:
     CypressError: `cy.type()` failed because it targeted a disabled element.

The element typed into was:

  > <input size="1" type="text" id="input-v-15" aria-describedby="input-v-15-messages" class="v-field__input">

Ensure the element does not have an attribute named `disabled` before typing into it.

      at Context.eval (webpack://@flowfuse/node-red-dashboard/./cypress/tests/widgets/form.spec.js:77:81)
 ```
 
 I have set them away again in the hope it is a transient error.

PR #1759, opened 2025-07-12
Allow basic widget ordering of widgets in a subflow

Steve-Mcl:
## NOTE
This PR merges into an earlier fix branch so to minimise review confusion, the simple fix in PR https://github.com/FlowFuse/node-red-dashboard/pull/1758 whould be reviewed and merged first.

## Description

Allow user to manually enter sorting order when a widget is in a subflow (similar to how width and height are handled)

#### Number
image

#### Chart

image

#### Other Nodes
They all follow same pattern

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/widgets/ui_audio.html
+++ b/nodes/widgets/ui_audio.html
@@ -38,12 +38,12 @@
                     // change inputs from hidden to text & display them
                     $('#node-input-width').attr('type', 'text')
                     $('#node-input-height').attr('type', 'text')
-                    $('div.form-row.nr-db-ui-element-sizer-row').hide()
-                    $('div.form-row.nr-db-ui-manual-size-row').show()
+                    $('div.form-row.nr-db-ui-element-hide-in-subflow').hide()
+                    $('div.form-row.nr-db-ui-element-show-in-subflow').show()
                 } else {
                     // not in a subflow, use the elementSizer
-                    $('div.form-row.nr-db-ui-element-sizer-row').show()
-                    $('div.form-row.nr-db-ui-manual-size-row').hide()
+                    $('div.form-row.nr-db-ui-element-hide-in-subflow').show()
+                    $('div.form-row.nr-db-ui-element-show-in-subflow').hide()
                     $('#node-input-size').elementSizer({
                         width: '#node-input-width',
                         height: '#node-input-height',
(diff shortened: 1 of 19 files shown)


colinl:
Is this to allow ordering of widgets within the subflow or within the group?

joepavitt:
It will allow any nodes, that are used in a subflow, to have their order set, which defines their position in their assigned group, even after being added to a subflow

colinl:
So does that mean that if a subflow contains multiple widgets that the widgets will not have to be adjacent in the group?  Currently the subflow itself appears as a single entity in the dashboard panel.  Is that going to disappear?

Also does this mean that if a group contains a number of subflows and a widget is added at the start, that all the widgets in following subflows would have to be manually incremented or will node-red do that automatically?

joepavitt:
> So does that mean that if a subflow contains multiple widgets that the widgets will not have to be adjacent in the group?

Not sure yet, I haven't dove into the PR to experiment.

> Also does this mean that if a group contains a number of subflows and a widget is added at the start, that all the widgets in following subflows would have to be manually incremented or will node-red do that automatically?

Manually required.

joepavitt:
Working excellently

colinl:
I am trying to work out how one would use this feature.  If I have a group with a number of widgets in it, and a subflow with some widgets, and I and I want to position the subflow widgets between to existing widgets, how would I know what numbers to use for the order, and would the code automatically move all the later widgets down?

Also does this have an impact on third party ui nodes?

joepavitt:
> and a subflow with some widgets, and I and I want to position the subflow widgets between to existing widgets, how would I know what numbers to use for the order, and would the code automatically move all the later widgets down?

This only impacts the ordering _within_ the subflow. There is an issue currently, once a node moves inside a subflow, the order of those widgets (within the subflow) cannot be changed. This PR resolves that.

joepavitt:
> Also does this have an impact on third party ui nodes?

I believe the same logic applied to the core nodes would also need to be applied to third-party nodes, within their own code bases. Although @Steve-Mcl can confirm.

colinl:
> This only impacts the ordering _within_ the subflow

I asked that earlier and the answer was

>It will allow any nodes, that are used in a subflow, to have their order set, which defines their position in their assigned group, even after being added to a subflow

That appears to say that the order is the order in the group, not in the subflow.  Perhaps the question I asked then was unclear.

WRT third party nodes have you checked for unforeseen consequences if a third party node is included in a subflow?

joepavitt:
> That appears to say that the order is the order in the group, not in the subflow. Perhaps the question I asked then was unclear.

That was my mistake. Upon review, it just controls the order _within_ the subflow, which I think makes more sense. The ordering of that subflow is still then possible via the right-side ordering tooling.

> WRT third party nodes have you checked for unforeseen consequences if a third party node is included in a subflow?

There is no change for third-party nodes vs. how they operate now.

colinl:
OK, I will try to test it out, if I can find a few minutes.

PR #1768, opened 2025-07-15
Improve sizing and appearance of the progress widget

joepavitt:
## Description

The progress widget feels out of place to the other widgets, in particular because it doesn't fill the space of a row. As such, I've increased the size/height to support that. I've also added a little border and padding to make it blender with the UI more, as a solid block looked a little too amateur.

### Before

Screenshot 2025-07-15 at 16 12 35

### After

Screenshot 2025-07-15 at 16 32 34


Steve-Mcl (approved):
Approved & acknowledging test fail is unrelated

PR #1823, opened 2025-09-12
Chart: Ensure bar/histogram charts without a series color by data index

joepavitt:
## Description

- Default in ChartJS was to color by the index/data point, such that each bar had a unique colour when the bar chart was specified to have "None" for it's series.
- In eCharts, the default behaviour is to colour by series, as such, the single series of data plotted was all rendered as the same colour.
- This PR overrides the default, as a function of the `categoryType` property (series) set on the chart widget.

### Before

Screenshot 2025-09-12 at 11 14 13

### After

Screenshot 2025-09-12 at 11 13 41

## Related Issue(s)

https://discourse.nodered.org/t/first-impressions-of-dashboard-2-charts-v1-27-0/99012/

### DIFF
--- a/ui/src/widgets/ui-chart/UIChart.vue
+++ b/ui/src/widgets/ui-chart/UIChart.vue
@@ -631,6 +631,7 @@ export default {
                         name: label,
                         type: chartJStoECharts.seriesType(this.chartType),
                         stack: this.props.stackSeries ? 'total' : null,
+                        colorBy: this.props.categoryType === 'none' ? 'data' : 'series',
                         data: []
                     }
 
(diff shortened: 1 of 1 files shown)


jdotbdot:
You can colour the bars differently by choosing a series, eg id.
I think this is a more rational choice if one wants each bar coloured differently (for example to facilitate visual correlation between two charts.)
However it is impractical because the more different values a chart shows, the thinner the bars get.
image

Perhaps a better approach is to address that oddity (which is not new to v1.27)?

Also somewhere to define extra colours would be welcome!

joepavitt:
> However it is impractical because the more different values a chart shows, the thinner the bars get.

The reason for this is because it creates a grouped bar-chart, whereby a bar exists for each x-value _and_ the series-value. If series is set to "None", then it just renders them all against the single x-value.

I have modified the default here to align with the past versions of Dashboard

hardillb (approved):

PR #1825, opened 2025-09-12
Chart: Set top/bottom in accordance to presence of legend and title

joepavitt:
### Problem

- Default eCharts hardcodes bottom and top such that there is extra padding either side of the chart. In Dashboard, this is particularly noticeable when smaller charts are rendered.
- Additionally, when a chart does not use a title, the chart doesn't automatically adjust it's y-position to fill the now available space.

### Solution

- Introduce manual calculations that better size the chart vertically based on the presence of a title and/or legend.

### Before

Screenshot 2025-09-12 at 13 29 26

### After

Screenshot 2025-09-12 at 13 27 56

## Related Issue(s)

https://discourse.nodered.org/t/first-impressions-of-dashboard-2-charts-v1-27-0/99012/18

### DIFF
--- a/ui/src/widgets/ui-chart/UIChart.vue
+++ b/ui/src/widgets/ui-chart/UIChart.vue
@@ -37,6 +37,9 @@ export default {
     },
     computed: {
         ...mapState('data', ['messages']),
+        hasTitle () {
+            return this.props.label !== ''
+        },
         chartType () {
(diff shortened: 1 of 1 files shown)


hardillb (approved):

hardillb:
@joepavitt approved not merged as there are lot of test failures not just the usual 1

I have kicked off a test re-run

PR #1836, opened 2025-09-14
Chart: Be smarter about when we commit the update to the eChart options

### LINKED ISSUE #1819: v1.27.0: ui-chart is no more usable
### Current Behavior

with the new implementation, Firefox and Chrome are very laggy and even complains about JS blocking it.
Diagrams are also blinking when they are updated; watch this video: (in the middle I press F5 for reload)

https://github.com/user-attachments/assets/3f40c2c5-4193-4dbf-b35b-4f27d5952a39

Also the diagrams look quite different:
Image

### Expected Behavior

faster and more responsive as the old chart implementation.
this are the screenshots for comparison:

https://github.com/user-attachments/assets/32c8b30e-c922-4fe7-b65e-b81ba387ae23

Image

### Steps To Reproduce

have a complex diagram that is often completely updated; I gather all required data in a node and then export every minute a complete diagram with up to 3000 data points

### Environment

- Dashboard version: 1.27.0
- Node-RED version: 4.1.0
- Node.js version: 22.15.0-1nodesource1
- npm version:
- Platform/OS: RaspberryOS
- Browser: Firefox / Chrome on Ubuntu 24.04.3 LTS 

joepavitt:
## Description

Down to ~70ms to render 1,000 data points (previously 8 seconds) with this change.

### Changes

- Use a temporary `options` variable to track required changes, and _then_ update the eCharts with `setOption` once we've processed all data received in a message. Previously, we were using the chart itself as the source of truth, and regularly reading/writing to it, which caused the performance bottleneck.

## Related Issue(s)

Closes #1819

### DIFF
--- a/ui/src/widgets/ui-chart/UIChart.vue
+++ b/ui/src/widgets/ui-chart/UIChart.vue
@@ -646,7 +646,7 @@ export default {
         },
         /**
          * Update the chart given a set of eCharts Options
-         * @param options 
+         * @param options
          */
         updateChart (options) {
             this.updateYAxisLimits(options)


cstns (commented):
The chart works as expected with large data sets. 

One question: why are options passed around instead of stored as a data property on the component? Centralizing them in the component would make them accessible to all methods and also visible in Vue DevTools, while also improving debugging.

joepavitt:
> why are options passed around instead of stored as a data property on the component? Centralizing them in the component would make them accessible to all methods and also visible in Vue DevTools, while also improving debugging.

Yes, it's something I played with, but made it slower so I stuck with this approach.

PR #1858, opened 2025-09-19
Sidebar: Ensure "Fixed" layout adapts to mobile view

### LINKED ISSUE #1674: Sidebar "Fixed" mode does not fall back to "Appear over content" on mobile
### Current Behavior

According to the official documentation, when the Sidebar Style is set to Fixed, it should remain visible on desktop but automatically switch to "Appear over content" on mobile (below 768px width).
This behavior does not happen — the sidebar stays permanently visible, even on small screens.

### Expected Behavior

The sidebar should automatically switch to "Appear over content" behavior on screens smaller than 768px when Fixed is selected.

### Steps To Reproduce

Install a fresh instance of Node-RED with @flowfuse/node-red-dashboard
Set the Sidebar Style in ui-base to Fixed
Deploy and open the dashboard on a mobile device or using Chrome DevTools responsive view (e.g. iPhone 8)
Observe that the sidebar does not collapse or float over content, even if window.innerWidth is < 768

### Environment

Node-RED v3.1.9 with latest @flowfuse/node-red-dashboard
Tested on Chrome, Firefox
Tested with clean configuration, no custom CSS or JS

joepavitt:
## Description

- Changes the layout style for mobile if "Fixed" is set. At 768 pixels (standard "mobile" breakpoint), we now remove the fixed sidebar and revert to the "Appears over Content" as per our documentation

## Related Issue(s)

Closes #1674

### DIFF
--- a/ui/src/layouts/Baseline.vue
+++ b/ui/src/layouts/Baseline.vue
@@ -1,7 +1,7 @@
 <template>
     <v-app class="nrdb-app nrdb-app--baseline" :class="`nrdb-view-density--${density}`" :style="customThemeDefinitions">
         <v-app-bar v-if="appBarStyle !== 'hidden'" :style="{'position': navPosition}" :elevation="1">
-            <template v-if="!['none', 'fixed', 'hidden'].includes(navigationStyle)" #prepend>
+            <template v-if="!['none', 'fixed', 'hidden'].includes(effectiveNavigationStyle)" #prepend>
                 <v-app-bar-nav-icon @click="handleNavigationClick" />
             </template>
             <v-app-bar-title>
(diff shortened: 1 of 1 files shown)


ui/src/layouts/Baseline.vue
```diff
@@ -269,8 +277,16 @@ export default {
...
         this.updateTheme()
+        this.checkMobile()
+        window.addEventListener('resize', this.checkMobile)
+    },
+    beforeUnmount () {
+        window.removeEventListener('resize', this.checkMobile)
     },
     methods: {
+        checkMobile () {
+            this.isMobile = window.innerWidth < 768
```
Steve-Mcl, line 288: Magic number is somewhat arbitrary and not conducive to being "mobile"

What is the logic here? Is it just that the screen is sufficiently small (regardless of being mobile) to trigger this logic?

joepavitt: > What is the logic here? Is it just that the screen is sufficiently small (regardless of being mobile) to trigger this logic?

Yes, "is mobile" isn't really the right term, more "is small", but its a breakpoint that's synonymous with mobile.

PR #1865, opened 2025-09-23
Chart: Remote animation when updating data

### LINKED ISSUE #1853: Echart - disable animation
### Description

Add the ability to disable chart animations like in previous versions of ui-chart. This is because if you have an update with the same data, you'll still see an animation, but the data won't change.

I am no FlowFuse team member

joepavitt:
## Description

- Sets `animationDuration` on the `series` of the `pie` charts, as for some reason eCharts doesn't pass the top-level property down for pie charts
- Set the value of the `animationDurationUpdate` to `0` so that when data is appended to the chart (after the first batch) no animation takes place.

## Related Issue(s)

Closes #1853

### DIFF
--- a/ui/src/widgets/ui-chart/UIChart.vue
+++ b/ui/src/widgets/ui-chart/UIChart.vue
@@ -221,9 +221,7 @@ export default {
                         }
                     },
                     color: this.props.colors,
-                    series: [],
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
Merging and acknowledging unrelated `✖  widgets/slider.spec.js ` e2e error

PR #1867, opened 2025-09-24
v1.28.0

joepavitt:
Version Bump


Steve-Mcl (approved):

colinl:
I see that 1.28.0 is available via npm but is not available via the palette manager yet. Does a refresh need to be triggered?

Steve-Mcl:
Thanks @colinl - i've pulled the trigger. Give it ~30m to appear in Node-RED.

Issue #1855, opened 2025-09-18
eCharts: allow "minData" / "maxData" for yAxis.min / .max


m-schaeffler:
### Description

allow "minData" / "maxData" for yAxis.min / .max, so that the diagramm starts / ends with min and max value to maximize the use of the area of the chart

at the moment this is blocked by the checks of the editor:
<img width="399" height="54" alt="Image" src="https://github.com/user-attachments/assets/e4e505cf-c463-40f2-b204-55a02c17dfa3" />

### Have you provided an initial effort estimate for this issue?

I am no FlowFuse team member

colinl:
Out of interest, can you provide some example data where the default auto scaling in v1.27.1 is not good enough?

m-schaeffler:
mainly this is when all values are in a very small range:

<img width="645" height="362" alt="Image" src="https://github.com/user-attachments/assets/c4042cc2-b21e-4151-8dd6-c26536e0ba07" />
If one value would be below 20°C it is even worse

colinl:
Yes, that is not ideal.

joepavitt:
> mainly this is when all values are in a very small range:

That's interesting - the logic I built in doesn't consider the scale of the numbers at all - only the range/difference between the two numbers, you can see it here:

https://github.com/FlowFuse/node-red-dashboard/blob/fc1caafef6aa6e4c921010d4f945429feafcc0a3/ui/src/widgets/ui-chart/helpers/axis.helper.js#L1-L46

I'm also open to improving this logic too. As well as supporting the proposal here. I think a good proposal here could be a `roundingUnit` of `1` maybe for a `range` of `<5`?

colinl:
I have submitted a [PR](https://github.com/FlowFuse/node-red-dashboard/pull/1874) with a suggested alternative algorithm.
For the data above, with values between about 20.2 and 22.8 it provides a range of 20 to 23

Snotmann:
Got the same ... the resolution is much too low:

<img width="623" height="232" alt="Image" src="https://github.com/user-attachments/assets/9c666a03-9e63-4daa-9cac-ea0b4f5072bb" />

Maybe a option for resolution steps are possible like 3 or 5 over and under the highest, lowest value.
Regards,
Snot

joepavitt:
> Maybe a option for resolution steps are possible like 3 or 5 over and under the highest, lowest value.

[...]

PR #1874, opened 2025-09-27
Proposed yaxis scaling algorithm for ui-chart

colinl:
## Description

A suggestion for an improved y axis scaling algorithm for when no axis limits are supplied.
This was generated with some help from ChatGPT and the algorithm is apparently a well-known numerical technique sometimes called "nice numbers for axis scaling" and is used, for example, in the D3 library.

## Related Issue(s)

It is relevant to #1855 but does not close that.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-chart/helpers/axis.helper.js
+++ b/ui/src/widgets/ui-chart/helpers/axis.helper.js
@@ -1,6 +1,8 @@
 const getAxisMinMax = (value) => {
     const min = typeof value?.min === 'number' ? value?.min : 0
-    const max = typeof value?.max === 'number' ? value?.max : 1
+    let max = typeof value?.max === 'number' ? value?.max : 1
+    // protect against min == max which is the case when only one point has been added
+    max = (max == min) ? min+1 : max
 
     const range = max - min
 
@@ -8,9 +10,11 @@ const getAxisMinMax = (value) => {
     const roughStep = range / (targetTicks - 1)
 
     const step = niceNumber(roughStep)
+    // number of decimal places necessary to represent step cleanly
+    const decimals = Math.max(0, -Math.floor(Math.log10(step)))
     // round min down and max up using multiples of step
-    const axisMin = Math.floor(min / step) * step
-    const axisMax = Math.ceil(max / step) * step
+    const axisMin = Number((Math.floor(min / step) * step).toFixed(decimals))
+    const axisMax = Number((Math.ceil(max / step) * step).toFixed(decimals))
 
     return {
         min: axisMin,
--- a/ui/src/widgets/ui-chart/helpers/axis.helper.js
+++ b/ui/src/widgets/ui-chart/helpers/axis.helper.js
@@ -1,3 +1,24 @@
+// Round x to a nice number, using factors of 1, 2, 5 or 10
+// and determine how many decimal places are needed to show it
+const niceNumber = (x) => {
+    // find x as a number between 1 and 9.9999 with power of 10 multiplier
+    // eg, 68.7 is 6.87 * 10^1 so exp is 1 and f is 6.8
+    const exp = Math.floor(Math.log10(x))
+    const f = x / Math.pow(10, exp)
+
+    // round f to 1, 2 5 or 10
+    let niceFraction
+    if (f < 1.5) niceFraction = 1
+    else if (f < 3) niceFraction = 2
+    else if (f < 7) niceFraction = 5
+    else niceFraction = 10
+    // and scale it up to the range of the input value so if x were 68.7 this returns 50
+    const niceX = niceFraction * Math.pow(10, exp)
+    // determine the number of decimal places necessary to represent this
+    const decimals = Math.max(0, -exp)
+    return [niceX, decimals]
+}
+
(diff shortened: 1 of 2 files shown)


colinl:
I will look at fixing the unit tests if it is decided that this is a worthwhile improvement.

colinl:
Also, I think it would be worth putting some caching in this code.  It is called twice (once for min and once for max) for every point added to the chart.  It is only necessary to recalculate if the data min/max object passed in has changed. For most situations, once a reasonable amount of points have been added, the range will not change.

joepavitt:
> Also, I think it would be worth putting some caching in this code. It is called twice (once for min and once for max) for every point added to the chart. It is only necessary to recalculate if the data min/max object passed in has changed. For most situations, once a reasonable amount of points have been added, the range will not change.

This is because of how eCharts needs these functions passed in. They require a `function` be passed to each of the `min`/`max` fields, which is then passed the `value.max` and `value.min`, it's annoying, but I couldn't see another way of doing this without also having to calculate the `min` and `max` values myself (which is also duplicating effort that eCharts already does for us)

For the algorithm fix, this seems to be a very similar approach. I think our fix is simpler - just need to add more granularity for the ranges between 1 and 100 which we don't currently have.

colinl:
I am not sure about the current approach being simpler, it is more lines of code.  If more granularity is added then that will complicate it still further.
Also it has some fundamental issues I think.  If, for example, it is fed values between 0 and 0.001 then it scales 0 to 0.1 which is no good. Extending the lower range only pushes the problem down to smaller numbers. The algorithm here correctly scales it 0 to 0.001.  Some engineers do work with very small numbers.

On the problem with recalculating every time, I can see how to improve that, I think, but I will leave that for a separate PR once the algorithm is finalised.

joepavitt:
Struggling to follow the intricacies of the maths here, however, if you can update the unit tests, then that'll give us a clear "this range gives this output" list to ensure we have good coverage

joepavitt:
Looking at the unit test results, it seems the new algorithm is still struggling with rounding problems, e.g.:

```
axis.helper
       getAxisMinMax
         should handle negative small values (min: -0.3, max: 0):

      AssertionError: expected -0.30000000000000004 to be -1
      + expected - actual

      --0.30000000000000004
      +-1
```

Where is has deemed the `min` to be a `0.30000000000000004` is the data ranges from `-0.3` to `0`

colinl:
I think that is the rounding problem fixed. I will try all the unit test ranges and make sure they look sensible and update the tests.

colinl:
Is it possible to run the unit tests locally before pushing?

joepavitt:
> Is it possible to run the unit tests locally before pushing?

Yep, `npm run test` will do the trick

colinl:
Great, thanks

colinl:
I don't think that the E2E test failure is as a result of my changes.

colinl:
I have realised that there is a better way of fixing the rounding issue, so have pushed a commit for that. Also I have improved the comments to make the maths more understandable, hopefully.

Steve-Mcl:
A quick look at your test cases suggests this may (mostly) satisfy #1889 - do you agree?

colinl:
Yes it should satisfy #1889.  I was just about to post that.

ui/src/widgets/ui-chart/helpers/axis.helper.js
```diff
@@ -1,49 +1,49 @@
...
+    if (f < 1.5) niceFraction = 1
+    else if (f < 3) niceFraction = 2
+    else if (f < 7) niceFraction = 5
+    else niceFraction = 10
+    // and scale it up to the range of the input value so if x were 68.7 this returns 50
+    const niceX = niceFraction * Math.pow(10, exp)
+    // determine the number of decimal places necessary to represent this
+    const decimals = Math.max(0, -exp)
+    return [niceX, decimals]
+}
```
Steve-Mcl, line 45: I know that due to code arrangement, this is not a failure, but its making me itchy the fact a constant function (`niceNumber`) is declared on line 28 but is called on line 14.

If `niceNumber` were a regular function it would be hoisted and I would not have baulked, but as this is a constant expression, i would prefer it to be moved up to before it is called.

i.e. Can we please move all of `const niceNumber = (x) => { ... }` (lines 26 ~ 45) to above the declaration of `const getAxisMinMax = (value) => {`

Steve-Mcl (commented):
Functionally, all good.

However, its making me twitch a little due to order of declaration.

Please move the `niceNumber` constant expression to above its first usage.

ui/src/widgets/ui-chart/helpers/axis.helper.js
colinl: Yes, you are quite right.
Do you think it should be  moved to the top, or should it be inside getAxisMinMax()?  I can see the argument for putting it inside, but I think keeping it outside makes the code easier to follow.

Steve-Mcl: I prefer outside (that way we can expose it later for other things and add tests if required)

PR #1879, opened 2025-09-29
Chart: Fix stacking in lines & add "Area" chart type

### LINKED ISSUE #1847: eChart: stackSeries option is used in line chart
### Current Behavior

when the option stackSeries is set to true from a former bar chart, it is also affecting line charts.

### Expected Behavior

stackSeries has no effect on line charts

### Steps To Reproduce

configure a stacked bar chart and then change it into a line chart

### Environment

- Dashboard version: 1.27.1
- Node-RED version: 4.1.0
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

joepavitt:
## Description

- Fixes the (unwanted) stacking in a line chart
- Adds a new "Area" chart type which does support stacking should users actually want this feature.

### After

Screenshot 2025-09-29 at 14 01 07

## Related Issue(s)

Fixes #1847

### DIFF
--- a/nodes/widgets/locales/en-US/ui_chart.json
+++ b/nodes/widgets/locales/en-US/ui_chart.json
@@ -2,6 +2,7 @@
     "ui-chart": {
         "label": {
             "line": "Line",
+            "area": "Area",
             "bar": "Bar",
             "scatter": "Scatter",
             "pie": "Pie",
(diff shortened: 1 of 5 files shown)


Steve-Mcl:
oh - not good on the e2e front: `✖  8 of 29 failed (28%)` I've re-run them

Issue #1876, opened 2025-09-28
v1.28.0: ui-chart X-axis grid line spacing


Subnum12:
### Current Behavior

after upgrade from 1.26 to 1.28.0 grid lines in ui-chart does get created in a strange behavior. The X-axis grid lines do not have the same spacing and are now much more frequent than the X-axis labels. Everything is fine with the Y-axis.

<img width="709" height="260" alt="Image" src="https://github.com/user-attachments/assets/da11e20c-a0a5-4a77-9b24-200b5fabd309" />

<img width="740" height="268" alt="Image" src="https://github.com/user-attachments/assets/73602772-6249-437d-8ee1-c91257e0729c" />

### Expected Behavior

Same behavior for X-axis grid lines as for Y-axis grid lines (same spacing over the chart, appear only if a corresponding axis value label esists)

### Environment

- Dashboard version: 1.28.0
- Node-RED version: 4.1.10
- Node.js version: 20.19.5

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

colinl:
This is mostly a duplicate of #1816 . If you think that does not cover it then please post additional comments there.

Subnum12:
That's quite possible, but it's not just that the grid lines are compressed; there are also generally more grid lines than grid line labels. The count should match.

colinl:
Can you try v1.29.0 please and see if that is good enough.

Subnum12:
Yes, it's significantly improved now.

Issue #1816, opened 2025-09-09
Investigate why compressed grid at either end of timescale charts


joepavitt:
### Current Behavior

Flagged here: https://github.com/FlowFuse/node-red-dashboard/issues/782#issuecomment-3270612844

Can be seen in the demo video here: https://github.com/FlowFuse/node-red-dashboard/issues/782#issuecomment-3258660432

### Expected Behavior

Gridlines should be consistent in spacing

### Steps To Reproduce

Running on `main` of Dashboard currently (after the migration to eCharts).

### Environment

- Dashboard version:
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

colinl:
@Steve-Mcl this can be closed now. I don't have authority to do that.

PR #1886, opened 2025-10-02
ui-chart: Remove additional vertical gridlines

colinl:
## Description

Remove extra and unevenly spaced vertical grid lines in ui-chart line chart.

For me, setting minorSplitLine to false removes the extraneous grid lines.  However there may be a reason it was set true, so there may be side effects that I have not noticed.

#1816 
#1876

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-chart/UIChart.vue
+++ b/ui/src/widgets/ui-chart/UIChart.vue
@@ -269,9 +269,6 @@ export default {
                         axisLabel: {
                             hideOverlap: true, // hide labels that would overlap adjacent label
                             color: textColor
-                        },
-                        minorSplitLine: {
-                            show: false
                         }
                     },
                     yAxis: {


colinl:
Before
image


After
image

colinl:
@Steve-Mcl, if you are looking for PRs to push through then please consider this one, which I think is a significant improvement to the aesthetics of the chart.

Steve-Mcl:
I looked at this one yesterday but I don't have enough information to make a informed decision right now.

Something that might help convince me is a comparison to the old chart.js with the same data. I guess what I'm saying is I have a million things to do but if someone were to present charts from before and after as a comparison it would save me time pulling this PR and testing it myself (the pictures would help me avoid pulling and testing and just be a simple code review) so if you've got the time it would be greatly appreciated.

colinl:
Here is basically the same chart on 1.26.0
image

I will create some more examples and post them here.

colinl:
Here are the two chart examples included in the dashboard examples imports
In 1.26.0
image
And in 1.28.0 with the PR
image
I had to hack the line chart example as it would not run in 1.26, which is why the timestamps are different. I just did a quick hack in order to see what it looks like.

I have checked that Categorical x axis line charts are not affected by the PR.

Steve-Mcl (changes requested):
After some play time with this the issue i am happier with removing this option all together.

Issue #1905, opened 2025-10-14
Chart with Timescale X axis - old data points not being removed as new ones are added


colinl:
### Current Behavior

For a chart with Timescale X axis and an X axis time limit set (e.g. last 1 hour), running in Append mode with new data points regularly added, then, as new data points are added, old data from before the specified time limit should be removed.  This does not in fact happen.  If the chart is left on screen then old data points are never removed.
If a different page is selected and then a return is made to the page then the refreshed chart is correctly limited to recent data.

### Steps To Reproduce

Import and deploy the flow below.  The Chart is configured with a time limit of 1 minute.  Leave the flow running for more than a minute and it will be seen that old data are not removed from the chart.

Navigate to a different dashboard page and then back to this one and the old data will be removed.

```
[...]

PR #1906, opened 2025-10-15
Fix pruning of old data from chart in Append mode #1905

colinl:
## Description

Correct code that prunes old data from the chart.
In addition, only prune data if the action is Append.  If the action is Replace then I think the users intention will be that all the data added (presumably as a batch of data) should be shown even if it is years old.

## Related Issue(s)

#1905

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


colinl:
Don't review this yet please, there is another issue https://discourse.nodered.org/t/dashboard-2-0-chart-not-showing-timestamped-data-from-mysql/99412. I think the server side is pruning data on page refresh even in Replace mode.

colinl:
@Steve-Mcl This is ready for review now.  I have added the ability to disable time based pruning by setting the value to 0.  This is so that the user can show historical data without it being immediately pruned away.

Does the fact that the docs and help text have been updated need to be flagged to the translators or is there an automatic system for this?

colinl:
I don't think that the test failures are mine.

Steve-Mcl:
> I don't think that the test failures are mine.

No, they are unrelated (its a odd timing issue we have never got to the bottom of)

colinl:
It seems that at least one of the chart examples relies on the fact that if historical data are added to the chart then no pruning occurs, unless the page is refreshed. I will fix the example. This does bring up the possibility that users may also be relying on this. I don't know what to do about that, other than note the change in the release notes. The adjustment required is to set the limit time to 0 in the chart.

colinl:
I have fixed the examples - two of them didn't work even without the latest mods.
So this is again ready for review.

colinl:
The data points come in as an object with properties x and y, but when they are added into the series they go in as and array of two elements, [0] is the x value and [1] is the y value.  See this code for example at line 670
```
                // Add data point
                if (this.props.xAxisType === 'category') {
                    // for categories, we need to update the existing data point for this x-value
                    const xIndex = options.series[sIndex].data.findIndex(d => d[0] === datapoint.x)
                    if (xIndex !== -1) {
                        options.series[sIndex].data[xIndex] = [datapoint.x, datapoint.y]
                    } else {
                        options.series[sIndex].data.push([datapoint.x, datapoint.y])
                    }
                } else {
                    // for other axes, we can just add the data point
                    options.series[sIndex].data.push([datapoint.x, datapoint.y])
                }
```

Steve-Mcl (approved):

Steve-Mcl:
Merging while acknowledging unrelated e2e test fail.

Issue #1927, opened 2025-11-07
The "Layout Editor" button for Dashboard Pages (when clicked) redirects to the wrong URL if altered `RED.settings.httpNodeRoot`


f0lg0:
### Current Behavior

The onclick event when opening the layout editor for a given UI page in `nodes/config/ui_base.html`

```js
async function showDashboardWysiwygEditor (pageId, baseId) {
    // call to httpadmin endpoint requesting layout editor mode for this group
    const windowUrl = new URL(window.location.href)
    const base = RED.nodes.node(baseId)
    const dashboardBasePath = (base.path || 'dashboard').replace(/\/$/, '').replace(/^\/+/, '')
    const authTokens = RED.settings.get('auth-tokens') || {}
    const headers = {}
    if (authTokens?.access_token) {
        headers.Authorization = `${authTokens.token_type || 'Bearer'} ${authTokens.access_token}`
    }

    // promisify the ajax call so we can await it & return false after opening the popup
    const ajax = () => {
        return new Promise((resolve, reject) => {
            $.ajax({
                url: `${dashboardBasePath}/api/v1/${baseId}/edit/${pageId}`, // e.g. /dashboard/api/v1/123/edit/456
                type: 'PATCH',
                headers,
                data: {
                    mode: 'edit',
                    dashboard: baseId,
                    page: pageId
                },
                success: function (data) {
                    // construct the url to open the layout editor
                    const _url = new URL(`${dashboardBasePath}/${data.path}`.replace(/\/\//g, '/'), windowUrl.origin)
                    _url.searchParams.set('edit-key', data.editKey)
                    const editorPath = windowUrl.pathname?.replace(/\/$/, '').replace(/^\/+/, '') || ''
                    if (editorPath) {
                        _url.searchParams.set('editor-path', editorPath)
                    }
                    resolve(_url)
                },
                error: function (err) {
                    reject(err)
                }
            })
        })
    }
    try {
        const url = await ajax()
        const target = `ff-dashboard-${baseId}` // try to reuse the same window per base
        const newWindow = window.open(url, target)
        // if we have an access_token then send it to the new window once opened
        // This becomes necessary if the user has gotten logged out since opening the editor.
        // Also, this squares up any token mismatch due to iframes/Same-Origin Policy etc.
        if (authTokens?.access_token && newWindow) {
            // Wait 1s then send the local access_token
            // (it is unlikely user will have done any editing and be wanting to deploy within 1s!)
[...]

PR #1929, opened 2025-11-13
fix: ensure dashboard editor URL is correctly constructed when Node-RED is running with a custom httpNodeRoot setting

f0lg0:
More info [here](https://github.com/FlowFuse/node-red-dashboard/issues/1927#issuecomment-3522127843)

### DIFF
--- a/nodes/config/ui_base.html
+++ b/nodes/config/ui_base.html
@@ -315,7 +315,7 @@ async function showDashboardWysiwygEditor (pageId, baseId) {
                     // construct the url to open the layout editor
                     const pathParts = []
                     const trimSlashes = (s) => s.replace(/^\/|\/$/g, '')
-                    const httpRoot = RED.settings.httpNodeRoot;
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
Just a heads up. I set copilot away on some dashboard PRs to give me a head start when i get around to reviewing and merging.

I appreciate the fact you guys are responding and debating the suggestions (even if they are nonsense) (should speed up my reviews - hopefully merged before start of next week)

f0lg0:
> Just a heads up. I set copilot away on some dashboard PRs to give me a head start when i get around to reviewing and merging.
> 
> I appreciate the fact you guys are responding and debating the suggestions (even if they are nonsense) (should speed up my reviews - hopefully merged before start of next week)

actually I think that suggestion was correct and useful so I've accepted it, but I may be wrong because I don't remember the default value for `RED.settings.httpNodeRoot` if it is not set. Anyway, I think having that extra check suggested by Copilot doesn't harm anyone.

Issue #1933, opened 2025-11-17
Gauge and Text nodes configured with static value do not show the value until message is sent


colinl:
### Current Behavior

If a gauge or text node is configured with the value field set to a string or number type the value is not shown in the dashboard until a message is sent

### Expected Behavior

The value should be shown immediately the widget appears in the dashboard

### Steps To Reproduce

Configure a gauge or text node with a string or numeric value field, deploy and view it in the dashboard. The value does not appear.  Then send it a message and the value will appear.

### Environment

- Dashboard version: 1.29.0
- Node-RED version: 4.1.1
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

colinl:
See forum thread https://discourse.nodered.org/t/ui-text-value-delayed-initialisation-2-bugs/99435

I am working on a PR for this.

PR #1934, opened 2025-11-17
Fix 1933, gauge and text with static value not shown till messages passed in

colinl:
## Description

If a gauge or ui-text is configured with the value field of type string or numeric then use that value immediately the widget is shown rather than waiting for a message to arrive.

## Related Issue(s)

#1933 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-gauge/UIGauge.vue
+++ b/ui/src/widgets/ui-gauge/UIGauge.vue
@@ -30,8 +30,9 @@ export default {
         value: function () {
             // if the value field is of type number or string then fetch the value from the properties
             // rather than the payload, so that the value is displayed immediately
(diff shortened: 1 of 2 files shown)


colinl:
The changes suggested by copilot are to fix a different issue - that the tile, tank and battery widgets do not currently cope with string values at all. Currently they display NaN when passed strings. Whether we want to get into that or not I don't know.  The gauge dial node does correctly show a string such as `7 litres` and coerces it to a number for the dial.
Since the other types do not display the number directly, but convert them to % values I am not sure that sending them strings makes any sense at all.

Issue #1892, opened 2025-10-05
msg.ui_update clears values in ´text´ node


Switchblade2:
### Current Behavior

If I send msg.payload and msg.ui_update in one msg to the text node then all is ok.
But if I only send msg.ui_update, it first looks ok, but if the web-page is reloaded, then value in the text field is erased. 

Test flow:
Click on A to set a value and colour. After Reloading webpage the value will remain (as expected)
Click on B to change the colour. Reloading the webpage will cause the value to disappear.
Click on C to set a new value. After Reloading webpage the value will remain (as expected)
So all is ok if you press A or C, but pressing C and then reloading webpage will cause the issue.
In my normal use case C is sent hourly followed by B if certain conditions are met. So I will often find the text field without value when I open the dashboard.

`[
    {
        "id": "9b512d81403d76be",
        "type": "inject",
        "z": "27e5b3e5cc7a2679",
        "name": "B",
        "props": [
            {
                "p": "payload"
            },
            {
                "p": "topic",
                "vt": "str"
            }
        ],
        "repeat": "",
        "crontab": "",
        "once": false,
        "onceDelay": 0.1,
        "topic": "",
        "payload": "",
        "payloadType": "date",
        "x": 170,
        "y": 840,
        "wires": [
            [
                "0bf7bac0ab238b06"
            ]
        ]
    },
    {
        "id": "429e4c30238f509c",
        "type": "function",
        "z": "27e5b3e5cc7a2679",
        "name": "payload & ui_update",
        "func": "msg = {}\n\nmsg.payload = (6.669).toFixed(1);\nmsg.ui_update = { color: \"#66cd00\" }\n\nreturn msg;",
        "outputs": 1,
        "timeout": 0,
        "noerr": 0,
        "initialize": "",
        "finalize": "",
        "libs": [],
        "x": 360,
        "y": 800,
        "wires": [
            [
                "c0221554f3a9397b"
            ]
        ]
    },
    {
        "id": "c0221554f3a9397b",
        "type": "ui-text",
        "z": "27e5b3e5cc7a2679",
        "group": "e1c6e6cb771c2108",
        "order": 3,
        "width": "2",
        "height": "1",
        "name": "kWh price",
        "label": "¢/kWh prices",
        "format": "{{msg.payload}}",
        "layout": "col-center",
        "style": true,
        "font": "Arial,Arial,Helvetica,sans-serif",
        "fontSize": "14",
        "color": "#717171",
[...]

PR #1937, opened 2025-11-18
Don't save message in datastore unnecessarily

colinl:
## Description

In ui_base.js onInput() each message is stored in the datastore before being sent to the client. This causes a problem if the message contains only msg.ui_update, msg.enabled, or msg.visible, so does not contain a msg.payload value for display.  In those cases, if the message is stored and then the browser is refreshed, the empty payload is sent to the client, causing the display value to be cleared.  Note that ui_update, enabled and visible are handled separately via the state store so do not need to be replayed.

This PR simply prevents the message with no payload from being stored in the data store.

I have looked through all the node types and cannot see any situations in which an empty payload should be replayed, apart from the Chart node, which supplies its own onInput() method which is run first, so the payload will no be empty by the time this code is run.

However, I cannot be certain that there will be no side effects of this change.

## Related Issue(s)

#1892

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/nodes/config/ui_base.js
+++ b/nodes/config/ui_base.js
@@ -32,7 +32,7 @@ function hasProperty (obj, prop) {
  * @param {*} msg
  * @returns true if other properties found
  */
-function hasExtraProps(message) {
+function hasExtraProps (message) {
     const allowed = ['_msgid', 'ui_update', 'class', 'visible', 'enabled']
     const keys = Object.keys(message).filter(key => message[key] !== undefined)
 


colinl:
It is certainly possible that widgets with custom onInput handlers could benefit from this change. This would require further analysis. I do know that the change here is required in order for ui_update to be used with the chart node. Perhaps a new issue should be raised to consider whether those with custom onInput handlers need the change, in order not to hold up the fix for the chart node.

I don't know how to implement a test of the type described.

Steve-Mcl:
Hey Colin, just FYI, I ran copilot on some of the PRs to get a head start on when I get around to reviewing (hopefully before next week).

Dont worry if some of the copilot suggestions are nonsense - but I appreciate you reading them and considering and responding (should speed things up)

Steve-Mcl:
I gotta admit I am a little nervous with this one since it will affect most of the widgets.

My sticking point is if a user has sent a ui_update (or something else) that should be stored and have deliberately left payload undefined so as to not affect the widget, it wont be stored. e.g. if i send `msg.ui_update.label` ONLY, I want that to be persisted. I also dont want to affect payload.

The solution may be to have a kinda `datastore.merge` that retains existing payload? 

What do you think @colinl ?

colinl:
The ui_update updates are stored separately in the state store  (the message is in the data store). eg at the start of ui_gauge.js
```
            beforeSend: async function (msg) {
                const updates = msg.ui_update
                if (updates) {
                    const hasLabelKey = Object.keys(updates).includes('label')
                    const hasTitleKey = Object.keys(updates).includes('title')

                    if (!hasLabelKey && hasTitleKey) {
                        updates.label = updates.title
                    }

                        // dynamically set "label" property
                        statestore.set(group.getBase(), node, msg, 'label', updates.label)
                    }
                    if (typeof updates.gtype !== 'undefined') {
                        // dynamically set "gauge type" property
                        statestore.set(group.getBase(), node, msg, 'gtype', updates.gtype)
                    }
...
```
Similarly msg.enabled and msg.class are handled separately as they must persist longer than individual messages.  You have got me wondering about nodes that allow the outgoing msg topic to be configured as an input msg property. I will have a look at that tomorrow.  There are several issues to do with that feature not working properly, I wouldn't want to make it worse.  Possibly I will see what those problems are when I look at how that is supposed to work.

Also what about those with msg passthrough?  Is that ever valid with no msg.payload.  I will look at that too.

Steve-Mcl:
Thanks Colin. Appreciated.

NOTE: if deemed necessary, you could make this change conditional - only for the chart node, that may be enough for getting it over the bar?

_untested - for ideas - code_
```js
const thisIsNotAChartNode = node.type === "chart" // or whatever we can to idnetify it

// always update datastore BUT if it is a chart, ensure payload is not undefined before commiting
if (thisIsNotAChartNode || typeof msg.payload !== 'undefined') {
    datastore.save(n, widgetNode, msg)
}
```

colinl:
Aargh, you are right.  With the advent of typed inputs on some nodes the value of interest could be in any property.  The safer way to handle this is to test for the message only having properties ui_update, visible, class or enabled (which are all saved in the state store).  If it has no other properties then don't store the message as it would effectively be an empty message.

colinl:
@Steve-Mcl is the E2E failure real? It is in the slider.

Steve-Mcl:
> @Steve-Mcl is the E2E failure real? It is in the slider.

It does appear to be. I re-ran tests and the same test " emits a value, and updates, when clicked" failed again.

may need to fix the reason for this error before realise.

colinl:
Are there any docs for doing that on the E2E test?  I don't know anything about them.

colinl:
It's OK, I have worked out what it is all about I think.  I can't see how that test is failing, it just clicks the slider and expects a message with the right value.  It is particularly odd as the next test in the file does the same thing and then does more, and it is passing.

Steve-Mcl:
> Are there any docs for doing that on the E2E test? I don't know anything about them.

it is a long and painful journey unfortunatly.



> I can't see how that test is failing, it just clicks the slider and expects a message with the right value. It is particularly odd as the next test in the file does the same thing and then does more, and it is passing.

ok, in that case, dont worry about this here.

colinl:
OK, it is ready for review again then.  Since the situatin now is that it does not store the message if it is effectively empty then I am as confident as I can be that there should be no side effects.  One can never be 100% of course.

Issue #1495, opened 2024-11-22
Water Tank Gauge Units


Donner176:
### Description

The current functionality converts the numeric payload value to a percentage based on the specified max range. Having the ability to specify units of measure (gallons, liters, ounces, etc.) and display the actual payload value would greatly enhance the usability of the Water Tank gauge.

### Have you provided an initial effort estimate for this issue?

I am no FlowFuse team member

sam-k-1:
I am also wanting this functionality

colinl:
I don't know why the tank displays % rather than allowing the user to specify the value and units in the same way as the other gauge styles.  The problem would be improving that without breaking existing dashboards.

gemini86:
This issue has been raised over a year ago and put in the backlog. Also currently the 'value' config item breaks the gauge if you set it to a JSONATA expression to prepend a suffix, resulting in the value displayed to be "NaN". I also cannot figure out what the segment config column on the right is supposed to do, as the documentation doesn't actually tell you about half the options on the config page.

colinl:
For the segments config it says
"Collection of objects that define the relevant color for a given value, and the overlay text source for Tile gauges."

So the right hand column is the overlay text for Tile gauges and is not relevant to the other types.

gemini86:
okay thanks- I think it's worth removing those when set to tank type... I'm working on a pull request so I'll try to clean this up as well- had to educate myself on vite and the build process to do so.

We'll see if it's worth anything, maybe just a starting point for somebody more experienced to jump off from

colinl:
Make the changes on a branch on your fork. Don't merge the branch into main.  Then when you commit the branch and push, github will ask if you want to create a PR on the flowfuse dashboard repository.  Do that and then we will see what you are suggesting and can comment on it.  Any further commits you push will automatically get added to the PR.

If you are going to remove the right hand column I would keep that for a separate PR.  Keep this for the option of showing values either as % or formatted values, which can then apply to other gauge types too.

gemini86:
Thanks for the coaching, colin. I'll keep it simple.

Issue #1275, opened 2024-09-07
Gauge Type: Tank Level: unit configuration


burner-:
### Description

Can we get configurable unit and presentation for tank level gauge. I like to show value in mm so probably some kind value template will give flexibility to fill wishlist of users. Example something like this: {msg.payload}mm. I am not 100% sure that is it good to do math in node (currenty it convers result to percentage). If it is wanted then template can handle it that way {msg.payload/100}% but I think it is more "standard" way to just let another node do math and let ui node just show raw results and use internal math only for visualisation.  

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Hi @burner- thanks for the suggestion. Our approach with this now in Dashboard 2.0 is to utilise the in-built "template" node.

I'll keep this issue open as we can update the documentation to reflect this recommendation, which is currently missing.

joepavitt:
This would now be supported with the TypedInput `value` field: https://dashboard.flowfuse.com/nodes/widgets/ui-gauge#formatting-value

gemini86:
> This would now be supported with the TypedInput `value` field: https://dashboard.flowfuse.com/nodes/widgets/ui-gauge#formatting-value

This does not actually do anything. The readout is still percent and we cannot add a suffix without it displaying "NaN".

PR #1966, opened 2025-12-14
Enable number display on UI Gauge Tank widget

gemini86:
## Description

The tank gauge looks so great, yet offers the least flexibility in value rendering. I suggest fixing this.

- Removed automatic conversion of the incoming `msg.payload` to a percentage of min/max. The gauge background wave animation already indicates this. 
- Restored options for prefix and suffix display. Now the water tank and display units like 'liters' or 'gallons', which is very useful in an industrial HMI.

Notes:
- Documentation on water type gauge is basically not there, so this change actually makes the gauge perform in a way more in line with the current documentation.
- I have tested within node-red locally to verify functionality.

## Related Issue(s)

https://github.com/FlowFuse/node-red-dashboard/issues/1495
https://github.com/FlowFuse/node-red-dashboard/issues/1275

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
## Labels

### DIFF
--- a/nodes/widgets/locales/en-US/ui_gauge.json
+++ b/nodes/widgets/locales/en-US/ui_gauge.json
@@ -30,7 +30,8 @@
             "units": "Units",
             "icon": "Icon",
             "defaults": "Defaults",
-            "value": "Value"
+            "value": "Value",
+            "valueAsPercent": "Display Value as Percent"
         },
         "errors": {
             "unique": "All 'from' values must be unique."
--- a/nodes/widgets/ui_gauge.html
+++ b/nodes/widgets/ui_gauge.html
@@ -142,6 +142,7 @@
                 order: { value: 0 },
                 value: { value: 'payload', required: false },
                 valueType: { value: 'msg' },
+                valueAsPercent: { value: Boolean(this.gtype === 'gauge-tank') }, // default to true for tank gauges
                 width: {
                     value: 3,
                     validate: function (v) {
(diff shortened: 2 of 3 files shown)


colinl:
This needs to be conditional, so the user can select % or value, otherwise it will break the dashboard for existing users who expect, and want, a % value.
Ideally the option should be present on all gauge types.

gemini86:
> This needs to be conditional, so the user can select % or value, otherwise it will break the dashboard for existing users who expect, and want, a % value. Ideally the option should be present on all gauge types.

I originally did intend to do this but thought it would be worth discussion. 

Currently, the user is configuring a value type with the typedInput dropdown. If they have configured, EG, `msg.payload.abc`, but we now go and introduce a "percentage" dropdown option which becomes the default, their configuration will be dropped because we cannot know what property they're passing in as a value. For that reason, I think the percentage option should be an additional checkbox which is checked by default on tank and battery gauges.

colinl:
> I think the percentage option should be an additional checkbox which is checked by default on tank and battery gauges.

Absolutely.  That is what I assumed would be the case.  Then it will display the value (wherever it comes from) as a number or as a percentage of range.

gemini86:
pushed some changes which add a checkbox for displaying value as percent- but this is only checked by default for the 'tank' type gauge, others uncheck by default.

It would really be my personal preference not to have checkboxes checking and unchecking themselves but if we really see displaying the actual value by default a "breaking change" (subjective I think) then I see no other way around this.

colinl:
I think it is only when a user has a tank on his dashboard and upgrades to the latest version that the current default must be used, so the display does not change when he upgrades.  If a new tank is added then probably it would be ok for it to default to value rather than %. @Steve-Mcl might have thoughts on this though.

Issue #711, opened 2024-03-27
Show / hide widgets (finer granularity than on group level)


NorbNorb:
### Description

Using UI Control, it is possible to dynamically show and hide groups or pages. If you only want to show or hide a text field dynamically, you have to wrap it in its own group, which not only makes it look bloated, but also forces it to be torn out of the logical order of the intended arrangement.
Suggestion: can a widget such as a text, text input, dropdown, etc. also be shown and hidden with UI Control?

### Have you provided an initial effort estimate for this issue?

I am no FlowFuse team member

liulysk:
I second this. Discussion: https://discourse.nodered.org/t/single-widgets-visibility/89689

IShouldTryAgain:
I too would like to support this. I am working on using the dashboards productively and this would elevate the possibilities whilst keeping the design. Discussion: https://discourse.nodered.org/t/single-widgets-visibility/89689

GuentherMair:
@NorbNorb: a short note to everyone who stumbles over this, the discussion on https://discourse.nodered.org/t/single-widgets-visibility/89689about and the introductory comment **"wrap it in its own group, which not only makes it look bloated"**.

You may use a "template" node with the following content:
```CSS
.no-vcard-title .v-card-item {
  display: none;
}

.no-vcard-title .v-card {
  padding: 16px 0px 0px 0px;
}
```

This should be configured as "Type: CSS (All Pages)" and linked to the UI where you want to use it. Then open the Dashboard sidebar, edit the group's where you want to apply it and add "no-vcard-title" to the "Class" field. Now save it and you'll be set without things being too bloated.

Nevertheless, I'd love to be able to dynamically hide UI elements out of the box.

joepavitt:
Another request for this: https://discourse.nodered.org/t/how-to-programmatically-hide-a-specific-button-in-a-group-widget-in-node-red-dashboard/92758/2

joepavitt:
Have increased priority given the attention this has had

okimselim:
You can set up using a ui-template node set to type "CSS (All Pages) with this:

`.hide-widget {
    display:none !important;
}

.show-widget {
    display: grid !important;
}`

Then you can dynamically send a message to a widget node with msg.ui_template.css set a to string of "hide-widget" to hide the widget on the dashboard or "show-widget" to bring it back. 

[...]

PR #1973, opened 2025-12-15
Fix widget enable/disable via ui_control

colinl:
## Description

Fix widget enable/disable via ui_control node

## Related Issue(s)

Partial fix for #711.  Widget enable/disable fixed, but does not address widget show/hide

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/layouts/Group.vue
+++ b/ui/src/layouts/Group.vue
@@ -144,6 +144,10 @@ export default {
             if (widget.state.class) {
                 classes.push(widget.state.class)
             }
+            // hide the widget if required
+            if (this.resizable === false && widget.state.visible === false) {
+                classes.push('d-none')
+            }
             return classes.join(' ')
         },
         getWidgetEditingClass (widget) {


Steve-Mcl:
> Partial fix for #711. Widget enable/disable fixed, but does not address widget show/hide

To include visibility fix, try adding:

```js
            if (this.resizable === false && widget.state.visible === false) {
                classes.push('d-none')
            }
```

to `ui/src/layouts/Group.vue` at line 147 (inside `getWidgetClass`)

It is essentially stating: "If not wysiwyg editing (resizable false) and  widget.state.visible is false, apply class `d-hide`"
d-hide remove the widget from webpage layout by setting `display: none !important`

colinl:
Excellent, that works if I pass `msg.visible` true or false direct to the widget.
It still doesn't work if I do it via the ui_control node.  I will have a look at that.

Steve-Mcl:
> Excellent, that works if I pass `msg.visible` true or false direct to the widget. It still doesn't work if I do it via the ui_control node. I will have a look at that.

Weird, it worked for me. Double check your msg is formatted to use 
```
{ widgets: { show: ["id-of-node"], hide: [] } }
```

colinl:
It does work after I spell the node name correctly :)
I have done some more testing, with fully qualified node name, node not on screen when hiding and so on.
Ready for review.

Steve-Mcl:
Another candidate for E2E tests here me thinks. Prevent regression or breakage of the features.
If you wanna gave a go pls. Also Happy to ship without E2E tests if we raise a follow up.

PS, in case it is not obvious, thanks very much for your contributions. 
I wish I had more time to do more than test and review.
Hopefully we (FF) will be hiring more staff soon and this will ease up 🤞

colinl:
I will raise an issue for the E2E tests. I would like to get the remaining problems with chart options sorted first.

Issue #1978, opened 2025-12-19
Various problems with Chart Option settings on Series


colinl:
### Current Behavior

There are a number of issues using msg.ui_update.chartOptions where the options specified act on Series. 
1. The settings are not persistent across page refresh.
2. If the options message is sent before there is data on the chart for the series being referenced then the option may not be applied when data are added.
3. The docs do not specify that any settings on series must include the series name in order to identify it.
4. After fixing 1 and 2 a further issue is apparent.  If the chartOptions for a chart that is not on the currently displayed page are updated, and then the page is changed to show the chart, the modified options are not used.  If the page is then refreshed then they do appear.

Related forum post https://discourse.nodered.org/t/echarts-options-via-msg-ui-update-not-persistant/99968

### Environment

- Dashboard version: 1.30.0
- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

colinl:
Item 4 added to description.

colinl:
@Steve-Mcl I need some help with item 4.  I cannot work out what is going on.

If I make updates to a chart that is not currently displayed and then switch to that page then I can see in data-tracker.mjs onWidgetLoad (msg, state) that `msg` contains the data to be shown on the chart and `state` contains
```
{"enabled":true,"visible":true,"class":"","chartOptions":{"series":[{"name":"a","lineStyle":{"color":"rgb(0, 0, 0)"}},{"name":"b","lineStyle":{"color":"rgb(0, 255, 0)"}}]}}
```
so the chartOptions have successfully got to that code.  That then runs
```
            store.commit('ui/widgetState', {
                widgetId,
                config: state
            })
```
which I thought should magically update `this.props`.  Then onWidgetLoad calls onLoad() in UIChart.vue with the chart data.

In onLoad() in UIChart.vue the data for the chart is passed in as expected, but `this.props` just has an empty chartOptions entry.  If I refresh the page then the modified chartOptions are correctly available in onLoad().

I have spent most of the day trying to work out what is going on, and failed.  Can you throw any light on the problem?

Steve-Mcl:
push your work to a branch and raise a "draft" PR Colin & I'll take a look.

Issue #1977, opened 2025-12-19
Excessive CPU usage with a dynamic, stacked bar chart


TMRh20:
### Current Behavior

I recently migrated to this dashboard and have been seeing excessive CPU usage in my main flow. After struggling for a little while, I cleared a single charts history, and the CPU usage dropped immediately.

I have it set up as follows:
1. Incoming, short MQTT message ("Node 1", "Node 2", etc.)
2. Function:
```
  msg.topic = msg.payload;

  let counterNodes=context.get('counterNodes') || 0;

  counterNodes++;
  if(counterNodes > 100){
    counterNodes = 0;
  }
  context.set('counterNodes',counterNodes);
  msg.payload=counterNodes;
  return msg;
```
<br>
3. This feeds directly into a chart with the following configuration:

<img width="631" height="753" alt="Image" src="https://github.com/user-attachments/assets/fd01eb78-aeec-4bf9-9c0d-138849100e5e" />

<br><br>
Essentially, this is an activity monitor for an IoT network. Each node gets a unique position in the chart, and the values range from 0 to 100. So you can see if a node is inactive if its bar stops moving.

<img width="852" height="480" alt="Image" src="https://github.com/user-attachments/assets/d9a267dc-df16-4ef2-a2fd-ea9d7dd0ada3" />

<br><br>

My interim solution is to add an injection node that clears the data every hour.

### Expected Behavior

I expected this chart to behave similarly to the original dashboard chart. The fact that I can resolve the issue by clearing the data every hour, tells me the chart is not getting rid of old data fast enough.

### Steps To Reproduce

Create a bar chart with dynamic data that fluctuates over time.

### Environment

- Dashboard version: 1.30.0
- Node-RED version: 4.1.2
- Node.js version: 20.19.6
- npm version: 10.8.2
- Platform/OS: Raspberry Pi 5 w/RPi OS 64 bit
- Browser: Firefox

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

colinl:
You have the Mode set to Append, which means you are continually adding new values. What happens if you set it to Replace?

TMRh20:
If I set it to replace, then there is only a single bar that moves, whether I choose side-by-side or stacks.

colinl:
OK, possibly it is not getting rid of old samples.  I will take a look.

colinl:
Do you know whether it is the node red server that is clogging up or the browser viewing the dashboard?

[...]

PR #1980, opened 2025-12-21
Fix chart options issues

colinl:
## Description

Partial fix for #1978.
Points 1 and 2 in the issue are believed to be fixed (though further testing is still required.
Points ~~3 (docs)  and~~ 4 still to be addressed

## Related Issue(s)

#1978

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/widgets/ui-chart/UIChart.vue
+++ b/ui/src/widgets/ui-chart/UIChart.vue
@@ -425,7 +425,9 @@ export default {
                     const seriesCount = this.chart.getOption().series.length
                     this.add(msg)
                     // if any series have been added, re-apply any chartOptions passed in
-                    if (this.chart.getOption().series.length > seriesCount) {
+                    // Also re-apply if this is a radial (pie or doughnut) chart as the radius for these are re-calculated
+                    // when adding data, which may remove any radius applied via msg.ui_update.
+                    if (this.chart.getOption().series.length > seriesCount || this.props.xAxisType === 'radial') {
                         // update the chart first from options applied in previous sessions
                         const chartOptions = this.chartOptions
                         if (chartOptions) {


colinl:
Point 4 is proving problematic.
Below is a flow that can be used to test it.

1. Ensure that a page other than 'Issue 1978 page' is open
2. Click the upper Inject node which will add data to the chart, then the lower Inject which should change the line colours to green and black
3. Open the 1978 page
4. It will be seen that the line colours are still blue/red
5. Refresh the page and the new line colours will appear

Debugging has shown that chartOptions with the line colours is getting into data-tracker.mjs onWidgetLoad() in the `state` parameter, but it does not appear to get from there  to `this.props`  when `onLoad()` in UIChart.vue is run. 

```
[{"id":"ed5e8b9dd7a75ea4","type":"group","z":"997da33a0beedade","name":"Issue #1978 test","style":{"label":true},"nodes":["697fd0b3c0f93657","e1d93299fa687a9c","350de7435ab7e7fb","e66ebdef3e42281c"],"x":54,"y":6979,"w":572,"h":182},{"id":"697fd0b3c0f93657","type":"ui-chart","z":"997da33a0beedade","g":"ed5e8b9dd7a75ea4","group":"f2036242466ec497","name":"Chart #1978","label":"Chart #1978","order":1,"chartType":"line","category":"[\"a\",\"b\"]","categoryType":"json","xAxisLabel":"","xAxisProperty":"ts","xAxisPropertyType":"property","xAxisType":"time","xAxisFormat":"","xAxisFormatType":"auto","xmin":"","xmax":"","yAxisLabel":"","yAxisProperty":"payload","yAxisPropertyType":"msg","ymin":"","ymax":"","bins":10,"action":"replace","stackSeries":false,"pointShape":"circle","pointRadius":4,"showLegend":true,"removeOlder":"0","removeOlderUnit":"3600","removeOlderPoints":"","colors":["#0095ff","#ff0000","#ff7f0e","#2ca02c","#a347e1","#d62728","#ff9896","#9467bd","#c5b0d5"],"textColor":["#666666"],"textColorDefault":true,"gridColor":["#e5e5e5"],"gridColorDefault":true,"width":"3","height":"4","className":"","interpolation":"linear","x":530,"y":7020,"wires":[[]]},{"id":"e1d93299fa687a9c","type":"inject","z":"997da33a0beedade","g":"ed5e8b9dd7a75ea4","name":"Historical data series a and b","props":[{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"[{\"ts\":1764798161000,\"a\":7,\"b\":5},{\"ts\":1764798221000,\"a\":8,\"b\":4},{\"ts\":1764798281000,\"a\":9,\"b\":5}]","payloadType":"json","x":220,"y":7020,"wires":[["697fd0b3c0f93657"]]},{"id":"350de7435ab7e7fb","type":"inject","z":"997da33a0beedade","g":"ed5e8b9dd7a75ea4","name":"Black/green","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"","payloadType":"date","x":170,"y":7120,"wires":[["e66ebdef3e42281c"]]},{"id":"e66ebdef3e42281c","type":"function","z":"997da33a0beedade","g":"ed5e8b9dd7a75ea4","name":"black/green","func":"msg = {\n  ui_update: {\n    \"chartOptions\": {\n      \"series\": [\n        {\n          \"name\": \"a\",\n          \"lineStyle\": {\n            \"color\": \"rgb(0, 0, 0)\"\n          }\n        },\n        {\n          \"name\": \"b\",\n          \"lineStyle\": {\n            \"color\": \"rgb(0, 255, 0)\"\n          }\n        },\n      ]\n    }\n  }\n}\n\nreturn msg;","outputs":1,"timeout":0,"noerr":0,"initialize":"","finalize":"","libs":[],"x":330,"y":7120,"wires":[["697fd0b3c0f93657"]]},{"id":"f2036242466ec497","type":"ui-group","name":"Issue 1978 group","page":"3aab758ffe5762d6","width":6,"height":1,"order":-1,"showTitle":false,"className":"","visible":"true","disabled":"false","groupType":"default"},{"id":"3aab758ffe5762d6","type":"ui-page","name":"Issue 1978 page","ui":"ID-BASE-1","path":"/Issue1978","icon":"home","layout":"grid","theme":"8dea5d50a632c4c6","breakpoints":[{"name":"Default","px":"0","cols":"3"},{"name":"Tablet","px":"576","cols":"6"},{"name":"Small Desktop","px":"768","cols":"9"},{"name":"Desktop","px":"1024","cols":"12"}],"order":-1,"className":"","visible":"true","disabled":"false"},{"id":"ID-BASE-1","type":"ui-base","name":"Dashboard","path":"/dashboard","appIcon":"","includeClientData":false,"acceptsClientConfig":["ui-control","ui-notification","ui-gauge-classic"],"showPathInSidebar":false,"headerContent":"page","titleBarStyle":"default","showReconnectNotification":false,"notificationDisplayTime":"1","showDisconnectNotification":true,"allowInstall":true},{"id":"8dea5d50a632c4c6","type":"ui-theme","name":"CDL test","colors":{"surface":"#ffffff","primary":"#0094ce","bgPage":"#eeeeee","groupBg":"#ffffff","groupOutline":"#cccccc"},"sizes":{"density":"default","pagePadding":"12px","groupGap":"12px","groupBorderRadius":"4px","widgetGap":"12px"}},{"id":"593dbf95e8d644c2","type":"global-config","env":[],"modules":{"@flowfuse/node-red-dashboard":"1.30.0"}}]
```

colinl:
Is the linter error significant?
Error:   191:17  error    Unexpected mutation of "props" prop    vue/no-mutating-props
at the line where I save the new chartOptions?

The linter is also complaining about 
` config.chartOptions ||= {}`
saying that the code has to be nodjs 14 compatible.  That can't be right can it?

Steve-Mcl:
> The linter is also complaining about
> ` config.chartOptions ||= {}`
> saying that the code has to be nodjs 14 compatible. That can't be right can it?

That'll be a linter limitation or rule setting. We need to update to next version of eslint but it's a rather large task (lots of breaking changes) so for now just go old school `thing = thing || default`



> Is the linter error significant?
> Error: 191:17 error Unexpected mutation of "props" prop vue/no-mutating-props
> at the line where I save the new chartOptions?

Yes, kinda. It's about the Vue adage: props down, events up. I.e. parent sets props, child emits changes up to the parent. There are a few ways to fix it. Gimmie 5mins...

Steve-Mcl:
Something like:

1. Add a local var in data()
Line ~38:

```js
data () {
    return {
        // ... (existing variables)
        resizeObserver: null,
        
        // add line
        dynamicChartOptions: null
    }
},
```


2. Update the Event Handler (onDynamicProperties)
use new local variable instead of the prop.
Line ~191:
```js
onDynamicProperties (msg) {
    // ...
    if (updates.chartOptions) {
        if (this.chart) {
            this.chart.setOption(updates.chartOptions)
        }
        
        // this.props.chartOptions = updates.chartOptions
        this.dynamicChartOptions = updates.chartOptions
    }
},
```

3. Update mounted() to check local variable
When the chart loads, check we have a local override first; otherwise, use the default prop.
Line ~160:

```js
// const chartOptions = this.props.chartOptions
const chartOptions = this.dynamicChartOptions || this.props.chartOptions

if (chartOptions) {
    chart.setOption(chartOptions)
}
```

4. Update onMsgInput() to check local variable
When new data comes in, use updated options (from local variable) (if they exist).
Line ~323:

```js
if (this.chart.getOption().series.length > seriesCount) {
    
    // const chartOptions = this.props.chartOptions
    const chartOptions = this.dynamicChartOptions || this.props.chartOptions
    
    if (chartOptions) {
        this.chart.setOption(chartOptions)
    }
}
```

100% untested and may not work.  If not, we might need to use props down, emit up pattern.

colinl:
OK, though I don't think there is any need to check it in mounted(), as the local variable will always be empty at that point, won't it?
There isn't any need to get the new props back down to the server as it already knows about the latest updates.

colinl:
Ready for you to have a look at the problem with props.chartOptions not making it into onLoad() when you have time.
I wonder whether it is to do with the fact that chartOptions does not exist in the configuration in the flows file.

colinl:
Docs updated, just point 4 outstanding.

colinl:
I've got it.  In UIChart.vue the export defaults props declaraion was missing  ` state: { type: Object, default: () => ({}) }`

colinl:
OK, all sorted. Ready for review.

colinl:
On the undefined issue, I have always previously used typeof, but I read that with modern JS, undefined is immutable so there is no need to do this any more. Happy to revert to this method though.

Steve-Mcl:
> On the undefined issue, I have always previously used typeof, but I read that with modern JS, undefined is immutable so there is no need to do this any more. Happy to revert to this method though.

You are correct however, when a variable is checked, but is not defined, an error will be thrown instead of a logic test true false evaluation. So for best practice I tend to stick to type check even though in this particular instance both were good.

Steve-Mcl (approved):

Steve-Mcl:
I'll likely get this merged on Monday & if time, I'll pull main & do some testing and try to get a release published.

colinl:
It would be good to also get in the PR I have just done for #1977, which is a serious issue for those affected.

PR #2069, opened 2026-03-18
fix: let browser handle navigation requests to support auth proxies

### LINKED ISSUE #2068: Service worker prevents auth redirect when dashboard is behind an authentication proxy
### Current Behavior

When the dashboard is running as a PWA, and is deployed behind an authentication proxy (e.g., Cloudflare Access), expired auth sessions result in "There was an error loading the Dashboard" instead of redirecting the user to the login page.

The service worker registers a NavigationRoute with createHandlerBoundToURL('index.html'), which intercepts all navigation requests (request.mode === 'navigate') and serves a cached index.html. This prevents the browser from ever reaching the authentication proxy's HTTP redirect (typically a 302) to the login page.

### Steps to Reproduce
1. Deploy Node-RED with @flowfuse/node-red-dashboard behind an authentication proxy (e.g., Cloudflare Access)
2. Install the dashboard as a PWA on an iPhone
3. Open the app and let the service worker install
4. Wait for the auth session to expire
5. Refresh the page, or open it again if it was previously closed
6. Expected: Browser follows the proxy's redirect to the login page
7. Actual: Service worker serves cached index.html, which fails to load because API requests are unauthorized, and user sees "There was an error loading the Dashboard"
This effectively breaks the dashboard for any deployment behind an authentication proxy once the session expires. The only workaround is uninstall the PWA and install it again.

See also [discourse conversation](https://discourse.nodered.org/t/dashboard-2-0-pwa-service-worker-prevents-authentication-redirect-when-session-expires/100574).

### Environment
- @flowfuse/node-red-dashboard latest (main branch)
- Tested with Cloudflare Access (Zero Trust) with Google IdP
- Reproducible across Safari (macOS/iOS)

### Expected Behavior

When the auth session expires, refreshing the dashboard should redirect the user to the authentication provider's login page. After re-authenticating, the user should be returned to the dashboard without needing to manually clear browser state or unregister the service worker.

No, but see #2069 for a suggested solution.

pakerfeldt:
## Description

When the dashboard is deployed behind an authentication proxy (Cloudflare Access, OAuth2 Proxy, Authelia, etc.), the service worker intercepts all navigation requests and serves a cached `index.html`. This prevents the browser from following the proxy's HTTP redirect to the login page when the auth session expires, resulting in "There was an error loading the Dashboard."

Remove the `NavigationRoute` registration so the service worker never handles navigation requests. The browser performs normal network navigation and can follow auth redirects naturally.
To prevent `precacheAndRoute` from implicitly serving `index.html` for navigation requests (via its `directoryIndex` and `cleanURLs` defaults), both options are explicitly disabled.
Static asset requests (JS, CSS, fonts, images) are unaffected, they are not navigation requests and continue to be served from the precache via exact URL match.

## Testing
Tested behind Cloudflare Access (Zero Trust) with Google IdP:
- Auth session expiry → browser redirects to login page → re-authentication returns to dashboard
- Static assets still served from precache

## Related Issue(s)

Fixes #2068

## Checklist

 Unsure how we would write tests for the service worker.
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/ui/src/sw.js
+++ b/ui/src/sw.js
@@ -1,37 +1,26 @@
 /// <reference lib="webworker" />
 import { clientsClaim } from 'workbox-core'
-import { cleanupOutdatedCaches, createHandlerBoundToURL, precacheAndRoute } from 'workbox-precaching'
-import { NavigationRoute, registerRoute } from 'workbox-routing'
+import { cleanupOutdatedCaches, precacheAndRoute } from 'workbox-precaching'
 
-// self.__WB_MANIFEST is the default injection point
-precacheAndRoute(self.__WB_MANIFEST)
+// Navigation requests (request.mode === 'navigate') are intentionally NOT
+// handled by the Service Worker. This lets the browser perform normal network
+// navigation, which is critical when the dashboard sits behind an
+// authentication proxy (Cloudflare Access, OAuth2 Proxy, Authelia, etc.).
+// Without this, the SW would serve a cached index.html, preventing the browser
+// from following HTTP redirects to the login page when the auth session expires.
+//
+// directoryIndex and cleanURLs are disabled so that precacheAndRoute does not
+// implicitly serve index.html for navigation requests to '/dashboard/'.
+// Static assets (JS, CSS, fonts, images) are still served from the precache
+// via exact URL match, so page loads remain fast.
+precacheAndRoute(self.__WB_MANIFEST, {
+    directoryIndex: null,
+    cleanURLs: false
+})
 
 // clean old assets
 cleanupOutdatedCaches()
 
-/** @type {RegExp[] | undefined} */
-const denylist = []
-
-// in dev mode, do not precache anything
-if (import.meta.env.DEV) {
-    // don't precache anything
-    console.log('Development mode, not pre-caching anything')
-    denylist.push(/.*/)
-} else {
-    // don't precache anything where the urls pathname ends with a slash (including times when the url has a query string)
-    // this permits the request to be handled by the server which will do a redirect as required
-    const configPath = self.location.pathname.split('/')[1]
-    denylist.push(new RegExp(`/${configPath}/[^?]*/(\\?.*)*$`))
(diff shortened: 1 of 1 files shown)


pakerfeldt:
With issue #2068  and this PR, I hope we can have a conversation on how to resolve the problem. This PR is a suggested solution that has been proving to work for me, but maybe people with more experience of service workers would come up with a better approach.

colinl:
As I said in the linked issue I cannot replicate the issue on Android.  I can, however, say that I have been running with this PR applied for over a week with a short expiry time on CFZT and have not seen any problems with the PWA install on my Android phone or a PC.

colinl:
Just a note to say that if anyone is testing anything in this area there can be problems when switching between builds.  I have had sometimes to not only uninstall and re-install the app, but also to clear the Chrome cache in order to get it to work reliably.  Without clearing the Chrome cache after rebuilding the dashboard source I have sometimes had the PWA hang showing the dashboard icon on startup.

colinl:
Sadly I have to revoke my comment saying that this PR is benign on Android.  With the PR, running as a PWA on Android I occasionally see a hang on startup, usually (perhaps solely) when the phone has been restarted.  On opening the app it just shows the node red icon, which stays there indefinitely.  I have not yet worked out exactly how to recover from the situation, but I think swiping it away and retrying does it, but not every time.  Similarly I have not determined exactly how to repeat the problem, it can run for days with regular phone restarts without exhibiting the problem before it shows itself.  I am convinced though that it only happens with the PR in place.  Without it I do not see the problem.

If I connect to a PC and inspect the window on the phone using Chrome it does not show any activity ongoing.

pakerfeldt:
I'm still successfully running on this PR myself.

colinl:
Is there any possibility of yourself borrowing an Android device to see if you can replicate the problem that I have seen with this PR in place?

PR #2078, opened 2026-03-25
Handle redirection to login page if session has expired when websocket connection is attempted

### LINKED ISSUE #2074: Hang with Connection Lost message when using authentication proxy and session times out whilst page is inactive
### Current Behavior

When using an authentication proxy such as cloudflare zero trust and the authentication session times out whilst the page is inactive (for example the PC is suspended or when running on a mobile device the OS puts the browser to sleep) then a Connection Lost is shown when the page is activated and remains shown until 5 minutes have elapsed or the page is refreshed. 
The same effect can be seen if the machine is left active, then after the session has expired disconnect the network connection between the browser and the server, wait (about 30 secs) for errors to appear in the developer console and then reconnect the network. The developer console will continue to show errors for 5 minutes before showing the login request.
If the browser developer console is opened it can be seen that the software is repeatedly trying to open the websocket connection but this fails because it gets redirected to the login session. After 5 minutes this times out and the s/w reloads the page, which brings up the login screen.

### Expected Behavior

The user should be taken to the login screen.

### Steps To Reproduce

1. Open a dashboard page in a browser on a PC
2. Suspend the PC
3. Wait for the authentication session to expire, or revoke the session
4. Resume the PC

OR

1. Open a dashboard page in a browser on a PC
3. Wait for the authentication session to expire, or revoke the session
4. Disconnect the network between browser and node red 
5. Wait for transport errors in the developer console
6. Reconnect the network

### Environment

- Dashboard version: 1.30.2
- Node-RED version: 4.1.7
- Node.js version: 20.20.0
- npm version:
- Platform/OS: Ubuntu on Laptop
- Browser: Edge

colinl:
Perform an http(s) pre-fetch before websocket connection in order to detect that login is required

## Description

Before an attempt is made to open the websocket connection, an attempt is made to http(s) fetch '_setup'. That is a small json file.  
If there is a network error then flow drops through to the existing retry strategy.
If the fetch successfully fetches the file then it is ok to make the connection.
If the content is not of type json then the domain is accessible, but there is an issue of some sort, such as the server wants to redirect to a login page, or has served the login page directly.  In that case the code forces a page reload, which will take the user to the login page.

## Related Issue(s)

Closes #2074 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


colinl:
This is working well for me on a PC and on Android, using Cloudflare Zero Trust and google as the authenticator.  I can't test it on anything else though.  I don't know what we can do to check it on other systems.

n-lark (approved):
Thanks @colinl! Tested and all looks good on my end. On testing other systems: I hear you, but since this is an improvement over what we have now, I'd rather take the step forward and watch for any reports.

Issue #2090, opened 2026-04-09
Option to remove inactive series from legend after configurable timeout


omidteimoori:
### Description

The `ui-chart` correctly creates and displays new series/topics when data arrives.

However, if one topic stops sending data, the chart keeps the topic/series visible in the legend and internal chart state indefinitely. The old topic is not removed automatically, even after it has clearly become inactive.

At the moment, the only reliable way I have found to fully remove the topic is to clear/reset the chart data manually (for example by sending an empty array), which is too destructive because it also removes the rest of the chart history.

There should be an optional property on `ui-chart` to automatically remove inactive topics/series after a configurable timeout.

For example:

- `Remove inactive series: true/false`
- `Inactive timeout: <seconds/ms>`

This could be optional because some users may want inactive series to remain visible, while others need them removed once the source stops publishing.

This is especially important for OT / IIoT dashboards where devices, stations, or MQTT topics may stop publishing temporarily or permanently.
If the topic remains in the chart legend after it is no longer active, the chart becomes misleading because it suggests the series is still part of the live visualization.

Suggested behavior could be as bellow: 
When enabled, if a series/topic has not received a new datapoint for the configured timeout, remove that series/topic from the chart legend and chart series list.
Ideally this should be done without requiring a full chart reset.

Happy to assist developing this feature.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

colinl:
Does it not seem more sensible that the legend should only appear if the chart contains on display at least one data point for that series?

colinl:
Also it should already be possible to selectively hide legends using msg.ui_update.chartOptions

omidteimoori:
I agree that in many cases it is sensible for the legend to only show series that currently have at least one displayed data point. However, in some applications, especially live industrial, telemetry, MQTT, and OT/IIoT dashboards, a series may temporarily stop publishing, and users may want one of two different behaviours depending on the purpose of the chart:

1. Keep the series visible so operators know that source existed recently, even if it is currently inactive.
2. Remove the series automatically after a configurable inactivity period so the chart reflects only currently active/live sources.

Regarding `msg.ui_update.chartOptions`, that is useful for manually controlling legend visibility, but it does not appear to solve the series inactivity lifecycle by itself in a built-in and automatic way. The main point here is having a native option for handling inactive topics/series without requiring flow-side custom logic for each implementation.

colinl:
[...]

PR #2094, opened 2026-04-10
Add opt-in inactivity expiry for ui-chart series/topics

omidteimoori:
## Summary
- add an opt-in ui-chart setting to expire inactive series/topics after a configurable timeout
- remove only the inactive series from chart state so active series remain untouched
- preserve existing ui-chart behavior when the option is disabled

Refs issue #2090

### DIFF
--- a/nodes/widgets/ui_chart.html
+++ b/nodes/widgets/ui_chart.html
@@ -123,6 +123,8 @@
                 removeOlder: { value: 1, validate: RED.validators.number(), required: true },
                 removeOlderUnit: { value: '3600', required: true },
                 removeOlderPoints: { value: '', validate: function (value) { return numberValidator(value, true) } },
+                removeInactive: { value: false },
+                removeInactiveAfter: { value: 60, validate: RED.validators.number(), required: true },
                 colors: { value: ['#0095FF', '#FF0000', '#FF7F0E', '#2CA02C', '#A347E1', '#D62728', '#FF9896', '#9467BD', '#C5B0D5'] },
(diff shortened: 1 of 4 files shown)


colinl:
Have you considered how this will work if the data points are added as a complete set of historical data from, for example, a database?  Will all the legends suddenly disappear after the specified time?

omidteimoori:
Yes. In the current implementation, inactivity is measured from the last time a series was received by the dashboard, not from the datapoint’s own x-value. So if a chart is populated from a one-off historical dataset and no further updates arrive, those series will expire after the configured timeout. That is intentional for the live-data use case this option targets.
For historical/static datasets, users can simply leave the option disabled, in which case ui-chart behaves exactly as it does today.

colinl:
Would it not have been much simpler to limit it to timestamp x axes and use the timestamp of the latest sample?

The PR you have submitted is remarkably complex and will take a lot of time to review, I suspect you will find it difficult to find a developer to do that.

omidteimoori:
That is absolutely possible, and I agree it would reduce the complexity of the implementation quite a bit.
The only real limitation with that approach is that chart types or modes that do not use a timestamp x-axis would not be able to support the feature. In practice, that would exclude things like pie and doughnut charts, because inactivity cannot really be determined there in the same clear and reliable way.

I am happy to narrow the implementation to that timestamp-based scope if that is the preferred direction.

colinl:
@Steve-Mcl do you have a view on this?

PR #2129, opened 2026-05-24
Improve passing of ui_update.chartOptions to clients to prevent memory growing unrestricted

### LINKED ISSUE #2012: Dynamic Chart options grow unrestricted

colinl:
## Description

Currently if a sequence of msg.ui_update.chartOptions is sent to a chart these are pushed into an array so they can be replayed when required.  This means that the memory usage can increase without bound if (for example) the user sends the chartOptions property with every sample.  The ideal solution would be to import lodash.merge in the client code so that the client can keep a merged set of options as does the client, but the overheads of this are best avoided if possible.

This PR solves the problem by always sending the full merged set of chartOptions to the client whenever the message includes ui_update.chartOptions, which does slightly increase the comms overhead but this is unlikely ever to be significant. It also simplifies the code in the client.

## Related Issue(s)

Closes #2012 

See also PR #2013 to which this PR provides an alternative.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


colinl:
@n-lark would it be possible for you to review this?

n-lark (changes requested):
Hey thank you for working on this! I think chart options aren't persisted client-side here. If you send a `chartOptions` update mid-stream, the next data message overwrites it back to default. The live update gets applied but isn't stored, so the replay re-applies `this.chartOptions` which is stale. 

You can reproduce with [this flow](https://gist.github.com/n-lark/f0ea9ad5d6cd34d1c4994d437ed01bc9) - click 1, 2, 3. Ring shrinks on 2, snaps back on 3.

If you just store the latest one it should fix it and stay bounded, something like: 

```
// data() 
mergedChartOptions: null,

// onDynamicProperties, after setOption line 195–196ish
this.mergedChartOptions = updates.chartOptions

// re-apply block line line 435–437ish
const chartOptions = this.mergedChartOptions || this.chartOptions
if (chartOptions) this.chart.setOption(chartOptions)
```

Lmk what you think!

colinl:
Yes you are right, normally the options are remembered by the chart itself, but it is necessary to save and re-apply if new series are added, and for the special case of radial charts. I will get that sorted.

colinl:
I have simplified your suggestion a little by saving the merged options in `mounted`.
It now runs your flow correctly and I have retested with a number of other test flows.

n-lark (approved):
Nice ty!

PR #2141, opened 2026-06-09
[1998] UI-TABLE only renders data when layout configuration is set to auto

### LINKED ISSUE #1998: UI-TABLE only renders data when layout configuration is set to auto
### Description

The UI-TABLE component only displays data when the layout configuration is set to auto.
If the layout mode is changed to any other option (e.g. manual), the table renders correctly at a structural level but no data is displayed, even though the data source is valid and unchanged.

### Actual Behavior

Data is only rendered when the layout configuration is set to auto. Any other layout mode causes the table to appear empty.

### Impact

This behavior limits layout customization and makes it impossible to use UI-TABLE in dashboards that require manual or fixed layout control.

### Expected Behavior

The UI-TABLE component should display data regardless of the selected layout configuration, as long as the data source is valid.

### Steps To Reproduce

	1.	Add a UI-TABLE component.
	2.	Bind it to a valid data source.
	3.	Set the layout configuration to auto → data is displayed correctly.
	4.	Change the layout configuration to a different mode → data is no longer shown.

#### Manual Config
Image

Image

#### Auto Config
Image

Image

### Environment

- Dashboard version: 1.30.0
- Node-RED version: 3.0.2 or 4.0.9
- Node.js version: 14.20.0 or 22.4.1 
- npm version: 6.14.17 or 10.8.1
- Platform/OS: Ubuntu Linux or macOS
- Browser: Chromium / Chrome

n-lark:
## Description

When a `ui-table` was given a fixed size, the data table got clamped to a single grid row, hiding the header and all but one row. This was because the component had multiple root elements, so the `grid-row-end` style that makes a widget fill its configured height never applied to it.

Wrapped the table in a single root element and used a flex column so it now fills its configured height and scrolls internally, while auto sizing still grows to fit. Added a fixed-size table to the test fixture with a regression test.

## Related Issue(s)

Resolves https://github.com/FlowFuse/node-red-dashboard/issues/1998

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/cypress/fixtures/flows/dashboard-tables.json
+++ b/cypress/fixtures/flows/dashboard-tables.json
@@ -42,6 +42,30 @@
         ]
     },
     {
+        "id": "dashboard-ui-table-fixed-height",
+        "type": "ui-table",
+        "z": "node-red-tab-tables",
+        "group": "dashboard-ui-group",
+        "name": "Fixed Height",
+        "label": "fixed height",
+        "action": "replace",
+        "order": 7,
+        "width": 6,
+        "height": 8,
+        "maxrows": 0,
+        "passthru": false,
+        "autocols": true,
+        "selectionType": "none",
+        "columns": [],
+        "x": 320,
+        "y": 320,
+        "wires": [
+            [
+                "test-helper"
+            ]
+        ]
+    },
+    {
         "id": "1ad285ca06fd711f",
         "type": "inject",
         "z": "node-red-tab-tables",
@@ -67,7 +91,8 @@
                 "dashboard-ui-table-single-row-click",
                 "dashboard-ui-table-multi-select",
                 "dashboard-ui-table-table-buttons-string-value",
-                "dashboard-ui-table-buttons-text-from-payload"
+                "dashboard-ui-table-buttons-text-from-payload",
+                "dashboard-ui-table-fixed-height"
             ]
         ]
     },
--- a/cypress/tests/widgets/table.spec.js
+++ b/cypress/tests/widgets/table.spec.js
@@ -11,8 +11,13 @@ describe('Node-RED Dashboard 2.0 - Tables', () => {
     })
 
     it('renders the provided data when a fixed widget size (not auto) is configured', () => {
-        cy.get('#nrdb-ui-widget-dashboard-ui-table-fixed-height').find('tbody tr').should('have.length', 5)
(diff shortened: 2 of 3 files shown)


Steve-Mcl (approved):
My only pause is any custom styles added by users may now break with an additional div wrapper around the div->.v-table & wonder if new CSS proposed could be applied to the existing (multiple) parent wrapper divs?

Image


If not, then fine, lets get this merged (will let you decide / merge)

n-lark:
Hey @Steve-Mcl so I looked into alternate solutions for this and what I came up with was adding a `tableGridStyle` computed that works out the table's start row:

```
   tableGridStyle () {
            if (!this.props.height) {
                return {}
            }
            const startRow = 1 + (this.props.label ? 1 : 0) + (this.props.showSearch ? 1 : 0)
            return {
                'grid-row': `${startRow} / -1`,
                'min-height': 0
            }
        },
```

So what this does is count the optional header rows (label/search) and set an explicit `grid-row` on the table so it spans the widget's full fixed height instead of collapsing to one row. This doesn't change the HTML structure, but it does assume label/search are one row each which could break in future. I added Cypress tests to catch that.

I'm not sold that this is a better solution since it feels a little fragile to me, but if keeping the DOM structure identical is important then this is the way to go. What do you think?

Steve-Mcl:
> I'm not sold that this is a better solution since it feels a little fragile to me, but if keeping the DOM structure identical is important then this is the way to go. What do you think?

It is probably a very small percentage of people who apply custom styling then an even smaller subset who target the table with an ultra specific selector - so I would say the blast radius is small enough to go with the original solution. Thanks for taking the extra time to look into this and lets go back to your v1 solution.

colinl:
@n-lark It would be worth checking this with the row height set to Compact in the theme if you have not already done that.

n-lark:
Hi @Steve-Mcl thanks for the consult, I moved back to the original solution. 

@colinl I did check all 3 row heights - thanks for the heads up! 

#### Compact 

Screenshot 2026-06-11 at 6 57 36 AM

#### Comfortable 
Screenshot 2026-06-11 at 6 58 11 AM

#### Default 

Screenshot 2026-06-11 at 6 58 28 AM

Steve-Mcl (approved):

Issue #2132, opened 2026-06-03
The "Layout Editor" Save button for Dashboard Pages (when clicked) redirects to the wrong URL if altered RED.settings.httpNodeRoot


ZmRZjFuDsDiR4qi5801F:
### Current Behavior

Related to issue #1927
Button "Layout Editor" works fine, but the Save button has similar issue and does not work if RED.settings.httpNodeRoot is other than default (e.g. "node-red"):  There is a 2nd /node-red in the built URL

Example (settings.json): httpNodeRoot = "node-red"

https://..../**node-red/node-red**/dashboard/api/v1/eb98e9e4f268f7cc/flows 404 (Not Found)

### Steps To Reproduce

Node-RED settings.json: httpNodeRoot = "node-red"
Start Node-RED
Create any Dashboard  with a single element
Press Edit Layout button
Change Layout
Press Save Layout
->404

### Environment

- Dashboard version:  1.30.2
- Node-RED version: 4.1.10
- Node.js version: 24.16.0
- npm version: Shipped with 24.16.0
- Platform/OS: Linux
- Browser: Edge

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

ZmRZjFuDsDiR4qi5801F:
Please could you give a estimation if this will be scheduled for a fix, I think it's a low hanging fruit. Thank you.

PR #2152, opened 2026-06-27
fix: reload dashboard under http node root

### LINKED ISSUE #2149: forcePageReload ignores httpNodeRoot
### Current Behavior

https://github.com/FlowFuse/node-red-dashboard/blob/666dd4fa933b9373e62530a74be862905d6f157f/ui/src/main.mjs#L88

When an error occurs, `forcePageReload` is called and redirects user to `/dashboard`.

### Expected Behavior

However the base path can be changed using the `httpNodeRoot` setting, in which case the user is presented with a 404 on error.

Instead; `forcePageReload` should redirect to `join(httpNodeRoot || '/', 'dashboard')`

### Steps To Reproduce

_No response_

### Environment

- Dashboard version: ~1.30.2
- Node-RED version: 5.0.0
- Node.js version: 24
- npm version:
- Platform/OS:
- Browser:

Saithej2k:
Summary

- Build forced reload URLs from the dashboard base path instead of always using /dashboard.
- Use the _setup response URL when unauthenticated responses arrive before setup is stored.

Testing

- git diff --check
- node --check ui/src/main.mjs

Closes #2149


colinl:
Is the change to use the setup url in the case of authentication error anything to do with issue #2149 or is that to address a different issue?

Saithej2k:
Good point. I narrowed that part out so the auth path no longer depends on the `_setup` response URL. The reload URL now comes from the stored setup base path when available, or from the current dashboard path, which keeps the change focused on #2149.

colinl:
While you have your brain in gear for the dashboard url, you might like to have a look at #2132.

Does the Open Dashboard button in the editor work if the root is altered?

Saithej2k:
I checked both paths. The Open Dashboard button already prefixes the configured dashboard path with RED.settings.httpNodeRoot, so it produces the correct URL when the root is changed.

The #2132 failure is separate: the layout editor save request derived its API path from the dashboard URL and then prepended the editor path, which could duplicate the custom root. I fixed that path and added coverage for default, shared, and different admin/node roots in #2165. I kept this PR scoped to the forced-reload behavior from #2149.

n-lark (approved):
Approved - thank you for the contribution @Saithej2k! 

Tested under a custom `httpNodeRoot` (/nr): a forced reload now targets `/nr/dashboard` instead of `/dashboard` → 404, and the default root still reloads to `/dashboard`. The auth-before-`_setup` case is covered by the current-URL fallback, so narrowing out the `_setup`-URL handling doesn't regress it. Open Dashboard button / #2132 is separate (tracked in #2165), out of scope here.

PR #2205, opened 2026-09-01
[1344] Undesirable reconnection behaviour 

### LINKED ISSUE #1344: Undesirable reconnection behaviour
### Current Behavior

While working on https://github.com/FlowFuse/node-red-dashboard/pull/1343 I noted we manually attempt to recover dashboard connection using fixed times (no jitter) and if that is unsuccessful it performs a full refresh.

In many settings (industrial/professional/signage/etc) this is undesirable.  consider the following:
1. 50 displays showing dashboards
2. Server maintenance is required - server is taken offline for 10 minutes
3. All 50 screens attempt a refresh and end up showing an ERR_NAME_NOT_RESOLVED
   *  "This site can’t be reached. xxxxxxxxx.com’s server IP address could not be found. Try: Checking the connection, Checking the proxy, firewall and DNS configuration. ERR_NAME_NOT_RESOLVED"
   * These now require manual intervention

### Expected Behavior

To use the WS built in capabilities for auto reconnect (e.g. making use of `randomizationFactor` for the exponential backoff jitter)

### Steps To Reproduce

disconnect or power off the server and wait until the dashboard gives up reconnection attempts.

### Environment

- Dashboard version: 1.17.x (current)
- Node-RED version: all
- Node.js version: all
- npm version: all
- Platform/OS: all
- Browser: all

n-lark:
## Description

Screens recover on their own, with no reload and no manual intervention. For #1344's 50-screens/10-min-outage case:

- **Never give up**: it used to quit after ~5 min and reload the page, stranding screens on a dead `ERR_NAME_NOT_RESOLVED` error when the server was still down. Now it just keeps retrying and reconnects when the server's back. (Safe to drop the reload: the login-redirect case it covered is now caught by #2078.)
- **Jitter**: each screen waits a slightly random amount before retrying, so 50 of them don't hit the server at the same instant.
- **Count-based**: retry pace tracks the number of attempts, not the clock, so a phone asleep for hours retries right away on wake instead of stalling.
- **Fast-wake**: a tab returning to the foreground reconnects immediately instead of waiting out the retry timer (and never fires two reconnects at once).

<details><summary> 🧪 Test plan</summary>

For testing I tried to emulate the real-life conditions this runs under: Android suspend/wake, auth-proxy session expiry, a server outage across many screens as closely as the environment allows. Lmk if we think this is enough.

- [x] **Normal reconnect**: killed and restarted the server (*a normal restart/deploy*); reconnected on its own, no reload
- [x] **Never gives up**: left the server down 10+ min; kept retrying, never reloaded to an error page
- [x] **Auto-recovery**: brought the server back after the outage; reconnected by itself, no refresh
- [x] **Fast-wake**: froze then woke the tab via `chrome://discards` (*a phone suspending and resuming a backgrounded tab*); reconnected the instant it woke
- [x] **Logged-out path still works**: faked an expired login, so the server returns a login page instead of data (via a DevTools override); it reloaded to the login screen instead of retrying forever, so dropping the give-up didn't strand a logged-out user
- [x] **Jitter**: unit test (can't run 50 real screens); two clients get different retry delays, within ±50%
- [x] **No needless reconnect when healthy**: switched tabs while connected; nothing fired, no reconnect
- [x] **No duplicate reconnect chains**: a tab waking mid-retry doesn't kick off a second reconnect.

</details>

## Related Issue(s)

Resolves https://github.com/FlowFuse/node-red-dashboard/issues/1344

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/test/ui/reconnect-interval.spec.js
+++ b/test/ui/reconnect-interval.spec.js
@@ -1,0 +1,40 @@
+const should = require('should') // eslint-disable-line no-unused-vars
+
+const { nextReconnectInterval } = require('../../ui/src/util/reconnect-interval.js')
+
+const noJitter = () => 0.5
+
+describe('nextReconnectInterval', function () {
+    describe('count-based escalation', function () {
+        it('uses 2.5s for the first few attempts', function () {
+            nextReconnectInterval(0, noJitter).should.equal(2500)
+            nextReconnectInterval(3, noJitter).should.equal(2500)
+        })
+
+        it('escalates to 5s after 4 attempts', function () {
+            nextReconnectInterval(4, noJitter).should.equal(5000)
+            nextReconnectInterval(13, noJitter).should.equal(5000)
+        })
+
+        it('caps at 30s from 14 attempts on, and never gives up', function () {
+            nextReconnectInterval(14, noJitter).should.equal(30000)
+            nextReconnectInterval(1000, noJitter).should.equal(30000)
+        })
+    })
+
+    describe('jitter', function () {
+        it('stays within ±50% of the base interval', function () {
+            for (const rc of [0, 4, 14]) {
+                const base = nextReconnectInterval(rc, noJitter)
+                nextReconnectInterval(rc, () => 0).should.equal(base * 0.5)
+                nextReconnectInterval(rc, () => 1).should.equal(base * 1.5)
+            }
+        })
+
+        it('varies the delay across clients (desync)', function () {
+            const a = nextReconnectInterval(0, () => 0.1)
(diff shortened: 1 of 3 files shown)


colinl:
I will do some testing with android.

Steve-Mcl (approved):
Good stuff. The fast recovery path works well in desktop chrome. Also tested on mobile (Android / chrome)

colinl:
Working well for me too on Android and Brave browser

PR #2227, opened 2026-09-15
[2226] Dashboard hangs on the loading screen after resuming a PWA  

### LINKED ISSUE #2226: Dashboard hangs on the loading screen after resuming a PWA
### Current Behavior

Occasionally, when returning to an installed PWA, the dashboard hangs on the loading screen forever. The Node-RED icon shows for a couple of seconds, then it switches to the loading screen with the cycling dots and just sits there indefinitely. It does not recover on its own; a swipe-down force-refresh fixes it.

It is intermittent (roughly zero to a couple of times a day) and more likely to surface under heavy reconnect churn, e.g. a short auth-token expiry. It appears to be pre-existing, reported by @colinl. 

### Expected Behavior

On returning to the PWA the dashboard should finish loading, or recover on its own, rather than hang on the loading screen until a manual refresh.

### Steps To Reproduce

Intermittent, so not reliably reproducible by hand:

1. Install the dashboard as a PWA.
2. Use it normally, ideally behind an auth proxy with a short session/token expiry to increase reconnect churn.
3. Background the app, then return to it.
4. Occasionally it reloads and hangs on the loading screen with the cycling dots.

### Environment

- Dashboard version: 1.31.0
- Node-RED version: 5.0.0
- Node.js version: 24
- npm version: 1.31.0
- Platform/OS: Android (installed PWA)
- Browser: Chrome

n-lark:
## Description

I suspect the intermittent PWA "stuck on the loading screen forever" is a failed initial socket connect. It fires `connect_error` rather than `disconnect`, and the `reconnect` loop was only wired to `disconnect`, so nothing retried. This starts the `reconnect` loop on that first failure too, so it should recover on its own instead of hanging.

## Related Issue(s)

Resolves https://github.com/FlowFuse/node-red-dashboard/issues/2226

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


colinl:
For testing I have applied this over branch 2068-sw-auth-proxy, I presume that is ok.

colinl:
I think you might have cracked it. I am seeing initially the same issue, but after about 10 seconds on the Loading screen there is a flicker as the Loading screen itself refreshes, and then the dashboard loads correctly.
Great work. I will continue to test it through tomorrow but it is looking good.

cstns (approved):

ui/src/main.mjs
colinl, line 1: I think maybe, as a matter of principle, this should include `retryCount = 0` before `reconnect()`.

colinl:
I am happy that this has solved the problem, at least for me.

ui/src/main.mjs
n-lark: Added thank you!

PR #2229, opened 2026-09-15
Dashboard data store - Adding the store 

### LINKED ISSUE #2236: Adding a data store
### Story 2 — Adding a data store 
 
> As a flow builder, I want Dashboard to keep the current value of every widget, so that state exists somewhere rather than only in messages that have already been sent.
 
Dashboard injects its proxied object into global context under its own namespace at startup. Setters defend the shape so a flow can't replace or corrupt it. Keys are derived from each widget's configuration — nobody names anything yet.

n-lark:
## Description

> [!IMPORTANT]
> This is part of a multi stage epic and should NOT be merged to main as is. 

Plumbing only: the store exists in global context but nothing reads from it yet, so dashboards behave exactly as before.

### The store

`nodes/store/reactive.js` is a proxy-backed store where writing a value is the change event, so there is no separate "store it, then emit it" step. Each key holds an entry with the value plus room for quality, timestamp and history. Reads return the value by default, and a `$` prefix returns the full record:

```js
global.get('dashboardStore.myKey')    // the value
global.get('dashboardStore.$myKey')   // { value, quality, timestamp, history }
```

Nested objects and arrays are watched too, so a deep change still fires the event. The backing object has a null prototype with prototype-pollution guards, clones on write so a flow cannot corrupt stored state, and flattens to plain `{ key: value }` for serialisation.

### Injection

`ui-base` injects the store into global context at `global.dashboardStore` on startup. It is idempotent across multiple base nodes and redeploys, so the same store is reused, and it rehydrates existing data after a restart.

## Related Issue(s)

Closes https://github.com/FlowFuse/node-red-dashboard/issues/2236

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


colinl:
Is there information publicly available on what this is about? The Flowfuse/engineering issue is not accessible.

n-lark:
Hey @colinl I'll get the issues moved to this repo before I move this out of draft. Will report back.

n-lark:
Hey @colinl I moved the parent epic Epic: Dashboard data store #2234 and the subtasks into this repo.

colinl:
So I see, thanks. This will keep you busy for a while :)

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
...
+
+    records[STORE] = true
+
+    const handler = {
+        get (target, prop, receiver) {
+            if (typeof prop === 'symbol') return Reflect.get(target, prop, receiver)
+            if (prop === 'toJSON') {
+                return () => Object.fromEntries(Object.entries(target).map(([k, r]) => [k, r.value]))
+            }
+            if (prop.startsWith('$')) return target[prop.slice(1)]
```
cstns: The `$` view hands back the live `Entry`, so it's read-only by convention only:

```js
store.k = 1
store.$k.value = 'TAMPERED'   // store.k is now 'TAMPERED', no onChange, no clone
```

Same with `store.$k.history.push(...)`, which grows straight past `maxHistory`.

Doesn't bite today since nothing reads the store, but once #2239 wires `onChange` to the socket this is a second write path that can disagree with the first, which is the thing the epic is trying to get rid of. Might be worth handing back a frozen projection rather than the record itself.

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+    return new Proxy(records, handler)
+}
+
+// Inject the store into global context once. On redeploy the existing store is reused;
+// after a restart with a persistent context store, existing plain data is rehydrated.
+function attachToContext (globalContext, opts = {}) {
+    const namespace = opts.namespace || NAMESPACE
+    const existing = globalContext.get(namespace)
+    if (existing && existing[STORE]) return existing
```
cstns, line 123: Tried this on a live instance: one `global.set('dashboardStore', {...})` from a function node and the store is gone. `$robot` comes back undefined and nothing observes writes after that. A full deploy does re-attach and rehydrate the plain value, so it recovers, but you lose metadata and history, and until then it's silently dead.

#2236 has "detect and recover from a flow replacing it" ticked, and the test at `reactive.spec.js:232` documents it as a known risk instead. Fine while nothing reads the store, but from #2239 onwards that's every dashboard quietly not updating, with no warning. Maybe a re-check on write, or a periodic guard?

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
...
+                target[prop] = rec
+            } else {
+                rec.history.push({ value: cloneValue(rec.value), timestamp: rec.timestamp })
+                if (rec.history.length > maxHistory) rec.history.shift()
+            }
+            const notify = (path) => { rec.timestamp = now(); onChange?.(prop, rec, path) }
+            // clone on write so a flow reusing its own object can't mutate stored state
+            rec.value = deepReactive(cloneValue(value), notify, prop, cloneValue)
+            rec.timestamp = now()
+            onChange?.(prop, rec, prop)
```
cstns: If `onChange` throws, the exception escapes into whoever did the write, which is going to be a user's function or change node. The value is already stored by then, so you get a successful write that looks like a failure.

"The write is the event" means the event's failure becomes the write's failure. Probably cheap to wrap now, before `onChange` starts doing socket IO in #2239.

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
...
+    if (Buffer.isBuffer(v)) return Buffer.from(v)
+    if (v instanceof Date) return new Date(v)
+    if (Array.isArray(v)) return v.map(deepClone)
+    const out = {}
+    for (const k of Object.keys(v)) out[k] = deepClone(v[k])
+    return out
+}
+
+function deepReactive (value, notify, path, clone) {
+    if (!isReactable(value)) return value
```
cstns, line 34: No cycle guard, so a self-referencing object blows the stack with a RangeError. Happens with both cloners.

Not reachable today since nothing writes, but it opens up as soon as widget messages land in the store. `msg.req`/`msg.res` happen to be safe because `isReactable` rejects class instances, so it's only plain circular objects.

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
...
+        deleteProperty (t, p) {
+            const existed = p in t
+            const ok = delete t[p]
+            if (existed) notify(`${path}.${String(p)}`)
+            return ok
+        }
+    })
+}
+
+function createDataStore ({ maxHistory = 5, onChange, clone = deepClone, now = Date.now } = {}) {
```
cstns, line 60: This keeps 5 full clones of every value and nothing reads history yet. I measured roughly 15MB retained for six writes of a 3,600 point series, which is the #2238 cost times six, before #2238 is even started.

Since it's unused, defaulting to 0 and turning it on when something actually consumes it might be worth considering.

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
...
+
+            let rec = target[prop]
+            if (!rec) {
+                rec = new Entry()
+                target[prop] = rec
+            } else {
+                rec.history.push({ value: cloneValue(rec.value), timestamp: rec.timestamp })
+                if (rec.history.length > maxHistory) rec.history.shift()
+            }
+            const notify = (path) => { rec.timestamp = now(); onChange?.(prop, rec, path) }
```
cstns: `notify` closes over `rec`, which outlives the value it was built for. If anything holds the old value object across a reassignment:

```js
const v = store.k   // { n: 1 }
store.k = { n: 2 }
v.n = 999           // fires onChange(..., 'k.n') and bumps $k.timestamp
```

The store itself stays correct, but you get an event for a value that isn't in it. Any widget caching its value object would emit phantom changes.

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
+const STORE = Symbol('dashboardDataStore')
```
cstns: This is module-local, so if two copies of the dashboard end up loaded in one Node-RED (happens more than you'd think with local plus project installs), the second `attachToContext` won't recognise the first's store. It rehydrates a copy and replaces it, leaving the first base node writing into an orphan. `Symbol.for('...')` would sidestep it.

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
...
+function createDataStore ({ maxHistory = 5, onChange, clone = deepClone, now = Date.now } = {}) {
+    const cloneValue = (v) => (v && typeof v === 'object') ? clone(v) : v
+    const records = Object.create(null)
+
+    records[STORE] = true
+
+    const handler = {
+        get (target, prop, receiver) {
+            if (typeof prop === 'symbol') return Reflect.get(target, prop, receiver)
+            if (prop === 'toJSON') {
```
cstns, line 70: Small one: a key literally called `toJSON` returns the serialiser instead of the stored value. Node ids are safe so it doesn't matter yet, but #2236 has readable keys as an open question, so probably worth settling before that lands.

nodes/store/reactive.js
```diff
@@ -0,0 +1,118 @@
...
+
+// Inject the store into global context once. On redeploy the existing store is reused;
+// after a restart with a persistent context store, existing plain data is rehydrated.
+function attachToContext (globalContext, opts = {}) {
+    const namespace = opts.namespace || NAMESPACE
+    const existing = globalContext.get(namespace)
+    if (existing && existing[STORE]) return existing
+
+    const store = createDataStore(opts)
+    if (existing && typeof existing === 'object') {
```
cstns: If someone already had an array sitting at `global.dashboardStore`, this quietly turns it into keys `"0"`, `"1"`, `"2"`. Edge case, but it's silent.

cstns (changes requested):
Nice piece of work, the core is solid and the test coverage is genuinely good. I checked the branch out, ran the suite (224 pass, lint clean), poked the store directly, and drove a live Node-RED 4.1.8 with the dashboard in a browser.

Confirmed working:
- No user-visible change. Text and gauge update on inject, replay-on-connect still fine after a reload, store shows up in the context sidebar and stays empty. The diff doesn't touch the frontend at all, so that one is structural rather than just observed.
- Restart persistence and rehydration with localfilesystem + `cache: true`, end to end. Values come back, reactivity re-arms.
- `cache: false` is handled. The get throws, `ui_base` warns, store is cleanly absent. It's incidental rather than an explicit check, and the warning leaks a Node-RED internal string, but it does the job.
- Untrusted JSON carrying `__proto__` doesn't reach `Object.prototype` through either cloner. Tried the top level and the nested set trap.

Two I'd want closed before #2239 builds on this, since both turn into silent correctness bugs the moment the store drives rendering:
- the `$` view is writable, so there's a second write path that skips `onChange` and clone-on-write
- a flow replacing the namespace kills the store until the next full deploy, with no detection

The rest are smaller and left inline.

One process thing: #2236 has all its ACs ticked, but "write each widget's current value into it, and clear the key when a widget is removed" and the corruption-recovery one aren't implemented (nothing calls the store, `ui_base` passes no `onChange`). Worth un-ticking so the epic reads honestly. Fully get that this is intentionally inert plumbing, just want the tracking to match.

nodes/store/reactive.js
n-lark: Good catch - fixed! `$` returns a frozen projection now, and the history entries are frozen too since `$k.history[0].value` was the same thing. Added specs too.

nodes/store/reactive.js
n-lark: added!

nodes/store/reactive.js
n-lark: Added guard good catch

nodes/store/reactive.js
n-lark: added

nodes/store/reactive.js
n-lark: fixed - used `isReactable`

nodes/store/reactive.js
n-lark: fixed made it a reserved key

nodes/store/reactive.js
n-lark: Added check to ignore old values

nodes/store/reactive.js
n-lark: Hey so the array case is already handled in [#2233 here](https://github.com/FlowFuse/node-red-dashboard/pull/2233/changes#diff-93c52b6d19196d5c4a80748d35eadc2b67f81e0295b21ab57751d5054d861decR81), where array values skip history entirely, so charts and tables both drop out of this. Can cut to 0 if we want but that should help the issue found here.

nodes/store/reactive.js
n-lark: So the acceptance criteria is for the overall stack - I tired to split things up to keep the PRs more focused. Could have been more clear on that my bad. 

The fixes you are describing are in https://github.com/FlowFuse/node-red-dashboard/pull/2230 then extended in https://github.com/FlowFuse/node-red-dashboard/pull/2233. https://github.com/FlowFuse/node-red-dashboard/pull/2230 tracks what we injected and warns through `node.warn` when it sees the namespace has been replaced. https://github.com/FlowFuse/node-red-dashboard/pull/2233 is where it stops needing a deploy, since `attachToContext` runs on every write there, so a clobbered store gets picked up on the next message, re-created, and the plain value rehydrated.

n-lark:
Hey @cstns ty for the review all points have been addressed. Around the AC, this is a stack that spans 3 PRs(https://github.com/FlowFuse/node-red-dashboard/pull/2230+ https://github.com/FlowFuse/node-red-dashboard/pull/2233) to break things up, so they are addressed just later on in the stack. Could have been more clear about that my bad.

cstns (approved):

nodes/store/reactive.js
cstns: Ah that makes sense, and no worries, splitting the stack is reasonable. Only suggestion would be noting on #2236 which PR covers which AC, so it's obvious nothing's been dropped.

I did check #2230's behaviour on a live instance: the warning fires, but only at deploy time, so the store stays dead and silent between the clobber and the next deploy. If #2233 moves `attachToContext` onto the write path then that closes it properly.

One thing worth flagging though, I've left a comment on #2230 about `isReactable` using a prototype identity check. Anything a function node creates carries a foreign `Object.prototype`, and `cloneMessage` preserves it, so `attachToContext` currently skips rehydration for exactly that data. That matters more once rehydration runs on every write instead of once per deploy, since a clobber from a function node is the likely case.

nodes/store/reactive.js
cstns: That covers the case I was measuring, since the 3,600 point series was the expensive part and arrays dropping out takes care of charts and tables.

Non-array values still keep 5 clones each, which is cheap per widget but not free across a big dashboard, and nothing reads history yet. Your call really, I'd lean to 0 until something consumes it, but the array carve-out handles the part that actually hurt.

PR #2230, opened 2026-09-15
Dashboard data store - Warn when the context store can't hold a live proxy 

n-lark:
## Description

The `global.dashboardStore` store is a Proxy, which is behavior rather than data, so it quietly stops working in two situations. This layer catches both and warns instead of failing silently.

### 1️⃣ Cache-off context store

A persistent context store with `cache: false` keeps no in-memory copy, so every read hands back plain data with the proxy stripped. `isInMemoryBacked()` probes for this up front, and `ui-base` injects the store only when it is safe, otherwise it warns.

### 2️⃣ Replacement

`global.set('dashboardStore', <anything>)` discards every widget's stored value. `attachToContext` remembers the store it injected, per context, so it catches any replacement, including `undefined` and primitives, which a shape check misses. Then re-creates the store, rehydrates any plain data, and calls `onReplaced`. A restart rehydrating plain data is not a replacement, so it stays quiet.

## Related Issue(s)

Relates to https://github.com/FlowFuse/node-red-dashboard/issues/2236

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


nodes/store/reactive.js
```diff
@@ -0,0 +1,152 @@
...
+        this.history = []
+    }
+}
+
+// only plain objects and arrays are deep-watched; Buffers, Dates, class instances etc. are opaque leaves
+function isReactable (v) {
+    if (Array.isArray(v)) return true
+    if (v === null || typeof v !== 'object') return false
+    const proto = Object.getPrototypeOf(v)
+    return proto === Object.prototype || proto === null
```
cstns, line 18: This is the one I'd most want fixed before #2233 and #2239 lean on it.

`Object.getPrototypeOf(v) === Object.prototype` is false for anything a function node creates, because Node-RED runs those in a `vm` context and the objects carry that realm's `Object.prototype`. `RED.util.cloneMessage` preserves the prototype, so the clone on write doesn't normalise it either.

Two things fall out of that, both reproduced:

1. Rehydration silently drops the data. On a live instance, a function node doing `global.set('dashboardStore', { robot: 'PLAIN' })` gets the `onReplaced` warning on the next deploy, but the store comes back `{}`. I instrumented `attachToContext` to be sure: `isReactable=false`, `proto===Object.prototype=false`.
2. A stored function-node object never gets a proxy, so `store.w.nested.temp = 25` fires nothing. Top-level writes still fire, so populating per message is probably fine, it's get-then-mutate that goes quiet.

`reactive.spec.js` does have `observes a function-node get-then-mutate`, but it builds the object in the test realm so it passes without covering this. A `vm.runInContext` fixture would catch it.

`Object.prototype.toString.call(v) === '[object Object]'` is realm-safe and still excludes Date and Buffer, if that helps.

nodes/store/reactive.js
```diff
@@ -0,0 +1,152 @@
...
+function attachToContext (globalContext, opts = {}) {
+    const namespace = opts.namespace || NAMESPACE
+    const existing = globalContext.get(namespace)
+    if (existing && existing[STORE]) {
+        injected.set(globalContext, existing)
+        return existing
+    }
+
+    const previous = injected.get(globalContext)
+    if (previous && existing !== previous) opts.onReplaced?.()
```
cstns, line 141: `attachToContext` only runs from the ui-base constructor, so this check only happens on deploy. Tested it live: clobber the store from a function node and the warning count stays at 0 through the clobber and every write after it, then fires on the next deploy.

So the detection works, but the dead window isn't any shorter, it just stops being silent once you redeploy. The description says it "catches any replacement", and the warning text reads as though it just happened, which could send someone looking at whatever they just deployed rather than the flow that actually did it. Might be worth softening the wording, or finding a cheaper place to re-check, maybe on write once #2233 lands.

nodes/store/reactive.js
```diff
@@ -0,0 +1,152 @@
...
+            const ok = delete target[prop]
+            if (existed) emit(prop, undefined, prop)
+            return ok
+        }
+    }
+    return new Proxy(records, handler)
+}
+
+// A cache-off persistent store serves only async access, so a sync get throws; the reactive proxy needs an in-memory store.
+function isInMemoryBacked (globalContext, namespace = NAMESPACE) {
```
cstns, line 119: This checks whether a sync `get` throws, which isn't quite the same question as whether the store round-trips by reference. It's correct for all three built-in configs (I tried memory, localfilesystem `cache: true` and `cache: false`), so no complaints about the practical coverage.

The gap is a store that serialises without throwing. It passes the probe, loses the proxy anyway, and then trips `onReplaced` on every single deploy, which would blame the user's flows for what's actually a context store config problem. I couldn't name a shipping store that behaves that way, so this is more robustness than a known break. Setting a sentinel, reading it back and comparing identity would test the actual invariant, if you think it's worth the extra lines.

nodes/store/reactive.js
```diff
@@ -0,0 +1,152 @@
...
+    })
+    // registered before walking children so a cycle resolves to this proxy instead of recursing
+    seen.set(value, proxy)
+    for (const k of Object.keys(value)) value[k] = deepReactive(value[k], notify, `${path}.${k}`, clone, seen)
+    return proxy
+}
+
+function createDataStore ({ maxHistory = 5, onChange, clone = deepClone, now = Date.now } = {}) {
+    const cloneValue = (v) => (v && typeof v === 'object') ? clone(v) : v
+    const emit = (prop, rec, path) => { try { onChange?.(prop, rec, path) } catch (err) {} }
```
cstns, line 62: Good call wrapping this. One thought: the empty catch means that once `onChange` is doing socket IO in #2239, a persistent failure is completely invisible. Worth routing it somewhere, even just a debug log, so it's diagnosable?

nodes/store/reactive.js
```diff
@@ -0,0 +1,152 @@
...
+            if (!rec) {
+                rec = new Entry()
+                target[prop] = rec
+            } else {
+                rec.history.push(Object.freeze({ value: cloneValue(rec.value), timestamp: rec.timestamp }))
+                if (rec.history.length > maxHistory) rec.history.shift()
+            }
+            const notify = (path) => {
+                // a replaced value keeps its proxies alive; they must not report into the record any more
+                if (rec.value !== wrapped) return
```
cstns, line 96: Nice guard, though it doesn't cover deletion. `rec` is untouched by `delete`, so a reference held from before still passes `rec.value === wrapped`:

```js
store.k = { n: 1 }
const held = store.k
delete store.k
held.n = 99      // still emits 'k.n' for a key that's gone
```

Same class as the one you just fixed, just via a different route.

nodes/store/reactive.js
```diff
@@ -0,0 +1,152 @@
...
+
+    const handler = {
+        get (target, prop, receiver) {
+            if (typeof prop === 'symbol') return Reflect.get(target, prop, receiver)
+            if (prop === 'toJSON') {
+                return () => Object.fromEntries(Object.entries(target).map(([k, r]) => [k, r.value]))
+            }
+            if (prop.startsWith('$')) {
+                const rec = target[prop.slice(1)]
+                return rec && Object.freeze({ value: rec.value, quality: rec.quality, timestamp: rec.timestamp, history: Object.freeze(rec.history.slice()) })
```
cstns, line 75: Two small things on the projection.

`Object.freeze` means `store.$k.value = 1` now throws a TypeError in strict mode, which cuts against the deliberate note a few lines down about no-op guards returning true so strict callers don't throw. Not obviously wrong, just inconsistent with the stated intent.

And every `$` read allocates a fresh object plus a copy of the history array. Irrelevant unless something ends up reading `$key` per message, which is worth keeping in mind for #2239.

cstns (changes requested):
Reviewed this one on a live Node-RED with both context store configs. 255 tests pass, lint is clean, and the headline feature works: with `cache: false` the warning is clear and actionable, the store is cleanly absent, and the dashboard still serves normally. Cycle guards hold up with both cloners, and `cloneMessage` even preserves the self-reference.

Also worth saying, the commits here landed hours before I posted my review on #2229, so you'd already found the frozen `$` view, the onChange try/catch, cycle guards, `Symbol.for`, the `toJSON` reservation, the array case and the stale-proxy guard independently. That's most of that review already answered, I'll go tidy those threads up.

One thing I'd want fixed before #2233 and #2239 build on this, left inline on `isReactable`: the prototype identity check is false for anything a function node creates, and `cloneMessage` preserves the prototype so the clone doesn't normalise it. That silently skips rehydration (reproduced live, the clobbered data comes back empty) and leaves stored function-node objects unwatched for deep changes. The spec that looks like it covers this builds its object in the test realm, so it passes.

The rest are smaller: replacement only being caught at deploy time, the in-memory probe testing a slightly different property than the one it needs, the empty catch around onChange, and a couple of minor ones.

nodes/store/reactive.js
n-lark: Fixed! I did `proto === null || Object.getPrototypeOf(proto) === null` rather than the `toString` check. Same realm safety, but `Object.prototype.toString.call()` returns `[object Object]` for class instances too, including `http.IncomingMessage` and streams which would be bad since we rely on that for the cycle deal in #2229. Added specs toooo.

nodes/store/reactive.js
n-lark: Updated wording to: 

> global.dashboardStore was replaced by a flow since the last deploy, discarding the stored value of every widget. The store has been re-created. To write your own data, set global.dashboardStore.<key> rather than replacing global.dashboardStore itself.

nodes/store/reactive.js
n-lark: Yahhh valid - so what I did was murder the `isInMemoryBacked` & made the `attachToContext` return `null` when we hit the same condition as `isInMemoryBacked` did. Then also added a check that compares what we just wrote vs what comes back to see if it is serialized, if it was then return`null`. Updated warning text too + specs.

nodes/store/reactive.js
n-lark: Added criteria to #2239 will address in that work. 

> - [ ] Surface subscriber errors from the store rather than swallowing them, now that onChange does socket IO

nodes/store/reactive.js
n-lark: Cleared out value in the `deleteProperty` trap.

nodes/store/reactive.js
n-lark: Yahhh so updated the comment so it only pertains to reserved key guards. I think its okay to keep the behavior for the freeze in strict mode since we'd wanna surface that mistake. Lmk if we want me to do more here but I think its chill. Will keep other in mind for #2239.

n-lark:
Hey @cstns so heads up on the diff for this one - when you rebased the stack of `249-data-store` the rebase wasn't floated up to this branch so the diff was wildly off when you reviewed. I've fixed it now but in the future just harass me and I can merge `main` into the stack since rebasing creates new commits then weird diffs and its just ick. Anyways, I addressed everything you commented on here, ty for review!

PR #2233, opened 2026-09-16
Dashboard data store - Populating the store 

n-lark:
## Description

Value-shaped widgets now write their current clean value into global context at `global.dashboardStore[nodeId]`, and clear the key when the widget is removed. Nothing reads from it yet, so dashboards behave as before.

```js
global.get('dashboardStore.<nodeId>')   // a flow can now read a widget's value
```

| widget | `global.dashboardStore[id]` |
| --- | --- |
| scalar (text, gauge, slider, …) | the value (`42`, `"b"`, `true`) |
| form | field object `{ name, age }` |
| table | the rows array |
| chart | not stored — see Charts in the data store https://github.com/FlowFuse/node-red-dashboard/issues/2238 |
| button, control, event, spacer | nothing |

Client-scoped messages (`msg._client.socketId` on a client-configurable node) are excluded, as they already are from the internal datastore.

The internal datastore still drives page loads, and is unchanged from `main` apart from the one mirror call at the end of `save`.

### Worth knowing

- A message with no `payload` key leaves the mirrored value alone; `payload: undefined` clears it. This matches the merge semantics of the legacy `save`.
- `ui-base` registers its injection options at module scope, so a re-injection after a replacement keeps its `onReplaced`.
- An enabled flag gates the write path, so a cache-off context store doesn't throw on every message.

## Related Issue(s)

Relates to https://github.com/FlowFuse/node-red-dashboard/issues/2236

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


nodes/store/data.js
```diff
@@ -90,6 +124,7 @@ const setters = {
                     ...data[node.id],
                     ...newMsg
                 }
+                writeToStore(node, msg)
```
cstns, line 127: `save` is also what `onSend` calls (`ui_base.js:794`), so client events now land in the store as widget state. Drove this over the real socket against a live dashboard:

```
flow sets the gauge   : {"txt1":42,"gt1":42}
user clicks the tile  : {"txt1":42,"gt1":{"id":"gt1","value":42,"segment":{}}}
```

So the key changes type, from the gauge's value to a click event object. Affects the three widgets that use `widget-send`: tile gauge, file input and ui-template. The table in the description says scalar widgets hold the value, which stops holding as soon as someone clicks.

This is the thing already noted in #2239's acceptance criteria, so I don't think it's news, but it's worth deciding here since this PR is what makes it real in the store rather than just in the internal datastore. ui-file-input is probably the sharpest version, since that payload can be a big base64 blob going into the store and into history.

nodes/store/data.js
```diff
@@ -1,9 +1,43 @@
...
+    storeOptions = { ...opts }
+    storeEnabled = true
+    return getOrCreateStore(globalContext)
+}
+
+function disableStore () {
+    storeEnabled = false
+}
+
+function writeToStore (node, msg) {
```
cstns, line 26: Worth double checking the client-scoping claim in the description. `canSaveInStore` only applies its check when `node.type` is in `base.acceptsClientConfig`, which defaults to `ui-notification` and `ui-control`. So a client-targeted message to a plain widget still writes to the shared store:

```
msg.payload = 'CLIENT_ONLY'; msg._client = { socketId: '...' }   -> ui-text
store: {"txt1":"CLIENT_ONLY"}
```

Same message to a `ui-notification` is correctly excluded. The description's parenthetical ("on a client-configurable node") is accurate, it just reads broader than it is. Presumably this is #2245's territory, so mostly flagging so the two don't drift apart.

nodes/store/data.js
```diff
@@ -1,9 +1,43 @@
...
+    if (!storeEnabled) return
+    if (!msg || typeof msg !== 'object' || !('payload' in msg)) return
+    try {
+        getOrCreateStore(node.context().global)[node.id] = msg.payload
```
cstns, line 30: This is where the `isReactable` thing I left on #2230 actually bites. Same payload shape, two different sources:

```
payload built in a function node -> deep mutation observed: false
payload built by a core node     -> deep mutation observed: true ["widget.nested.deep"]
```

Top-level writes still fire either way, so populating per message is fine as it stands. It matters for #2239, where a flow doing get-then-mutate on a stored value would silently not notify for anything a function node produced.

nodes/store/data.js
```diff
@@ -1,9 +1,43 @@
...
+    } catch (err) {}
```
cstns, line 31: This one and the one in `clearFromStore` make four silent catches in the stack now (these two plus `emit` and the `ui_base` init). If the context store starts throwing mid-run, every widget write no-ops forever with nothing in the log to say so. Probably worth a one-time warn rather than per-message noise?

nodes/store/data.js
```diff
@@ -1,9 +1,43 @@
...
+    return getOrCreateStore(globalContext)
+}
+
+function disableStore () {
+    storeEnabled = false
+}
+
+function writeToStore (node, msg) {
+    if (!storeEnabled) return
+    if (!msg || typeof msg !== 'object' || !('payload' in msg)) return
```
cstns, line 28: Small thing on `payload: undefined`. It reads as cleared, since `toJSON` drops it, but the key is still there:

```js
datastore.save(base, node, { payload: undefined })
// JSON view: {}
// 'w1' in store -> true, $w1 still returns an entry
```

`clearFromStore` does a real `delete`, so the two paths leave the store in different states. Doesn't matter yet, might once #2240 lets widgets bind to keys and needs to tell "no key" from "key with no value".

cstns (commented):
The two acceptance criteria this PR owns both check out on a live instance: the store populates from flow messages, and deleting a widget clears its key. Buttons store nothing and charts stay out via the array branch, as described. 284 tests pass and lint is clean.

Also worth saying, `attachToContext` on every write measures about 0.44us against 0.33us for the store write itself, so the cost of moving clobber recovery onto the write path is basically nothing. That closes the deploy-time-only gap I raised on #2230.

Nothing here is a blocker for a no-op PR, so leaving this as a comment rather than requesting changes. The one I'd most want a decision on before #2239 is client events landing in the store as widget state: a tile gauge click turns its stored value from `42` into a click event object. Already noted in #2239's acceptance criteria, but this PR is where it becomes real.

The rest are smaller: client-scoped messages reaching the shared store for widgets outside `acceptsClientConfig`, the `isReactable` realm issue from #2230 hitting this write path, and a few minor ones.

nodes/store/data.js
```diff
@@ -1,9 +1,43 @@
+const { attachToContext } = require('./reactive.js')
+
 const data = {}
 
 const config = {
     RED: null
 }
 
+let storeOptions = {}
+let storeEnabled = true
```
cstns, line 10: Defaults to enabled before any ui-base has had a chance to call `initStore` or `disableStore`. Probably unreachable in practice since the base constructor runs at deploy, but defaulting to off and letting `initStore` turn it on would make the cache-off path fail closed rather than open.

nodes/store/data.js
n-lark: Yah agreed we need to address - I made another issue for this since its a bug and I'd wanna address outside this epic: #2250. I'll knock it out before I move forward with the other work.

nodes/store/data.js
n-lark: Yahhh with the fix implemented I re-ran your case through save with a real vm realm and both sources notify now:

```
core node    -> deep mutation observed: true ["core.nested.deep"]
function node-> deep mutation observed: true ["fn.nested.deep"]
```

Specs cover it too now.

nodes/store/data.js
n-lark: I added this to #2239 since we'd wanna make sure we are using the same logging for all errors: 

> - [ ]  Store-layer failures surface once rather than silently, covering subscriber errors from onChange and write failures in writeToStore/clearFromStore, so a store that stops accepting writes mid-run is visible in the log without per-message noise

nodes/store/data.js
n-lark: I think we want these as two different behaviors? One is the widget being deleted from the flow, so `clearFromStore `drops the key, and the other is a flow blanking the value of a widget that still exists. Feels like "gone" and "present with no value" are genuinely different states, what you think?

nodes/store/data.js
n-lark: swapped

PR #2245, opened 2026-09-21
[2244] `clientId`-targeted values leak to other clients on reload  

### LINKED ISSUE #2244: `clientId`-targeted values leak to other clients on reload
### Current Behavior

`canSaveInStore` was never updated when #2208 added `clientId` targeting, so the send path and the store path disagree. `isValidConnection` filters on `socketId` and `clientId`; `canSaveInStore` filters on `socketId` only.

A `clientId`-targeted message is therefore delivered to the right client and then saved to the shared store. It escapes on reload: `onLoad` calls `conn.emit('widget-load:' + id, msg, state)` directly, bypassing `canSendTo`, so another client refreshing the page receives a value addressed to someone else.

Two stores have the same gap — `nodes/store/data.js` (the msg) and `nodes/store/state.js` (dynamic `enabled`/`visible`/`class`).

### Expected Behavior

A message scoped to a single `clientId` is never written to the shared store, so no other client can receive it — the same way `socketId`-scoped messages are already handled.

### Steps To Reproduce

Needs a widget type added to the `ui-base` node's **client config** list, since the two defaults can't show it — `ui-control` stores nothing, and `ui-notification` never requests `widget-load`.

1. Add `ui-text` to `acceptsClientConfig` on the `ui-base` node
2. Open the dashboard in two browsers, A and B
3. Send `{ payload: 'for A', _client: { clientId: '<A>' } }` to a `ui-text` widget
4. Only A updates, as expected
5. Refresh B — B shows `for A`

Same with `{ class: 'x', _client: { clientId: '<A>' } }` for the state store.

### Environment

- Dashboard version: 1.31.0
- Node-RED version: 4.1.13
- Node.js version: 24.15.0
- npm version: 11.12.1
- Platform/OS: macOS 26.5.1 (arm64)
- Browser: Chrome

n-lark:
## Description

Scopes `canSaveInStore` on `clientId` as well as `socketId` in both the data and state stores, so a `clientId`-targeted message is no longer saved centrally and replayed to other clients on reload.

See https://github.com/FlowFuse/node-red-dashboard/issues/2244#issue-5529430951

 ### Test plan 

- [x] `data.js` — targeted `payload` reaches A only; before the fix B showed it after a reload, after the fix B keeps its own value
- [x] `data.js` — A's own reload drops the targeted value back to the stored broadcast; later broadcasts still store and replay to both
- [x] `state.js` — targeted `msg.class` applies to A only; before the fix B picked it up after a reload, after the fix it does not
- [x] Unit specs in `test/store/data.spec.js` and `test/store/state.spec.js` (each `clientId` case fails without the fix)

## Related Issue(s)

Resolves https://github.com/FlowFuse/node-red-dashboard/issues/2244

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


nodes/store/data.js
```diff
@@ -19,7 +19,7 @@ function canSaveInStore (base, node, msg) {
 
     if (constrained.includes(node.type)) {
         // core check
-        if (msg._client?.socketId) {
+        if (msg._client?.socketId || msg._client?.clientId) {
```
cstns, line 22: Fix looks right, and matches `isValidConnection` which already checked both.

The thing that made #2244 possible is still here though: "is this message client-scoped?" is answered independently in three places (`data.js`, `state.js`, and `isValidConnection` in `ui_base.js`). #2208 updated one and the other two drifted, which is exactly this bug. A shared `isClientScoped(msg)` in `nodes/utils` that all three call would make the next targeting field impossible to half-apply.

Small related thing if you do that: `isValidConnection` runs `normalizeClientId()` before comparing, while these two test raw truthiness. Doesn't cause a leak for anything realistic, but a shared helper would make the two agree on odd inputs for free.

cstns (approved):
Reproduced the bug and the fix with two browsers following the steps in #2244: before, B shows `for A` after a reload; after, B keeps its own value. Also reverted just the two production lines and confirmed the four new specs go red, so they're genuine regression cover.

Checked the symmetry too, `isValidConnection` already filters on both `socketId` and `clientId`, and there's nowhere else in `nodes/` left with the socketId-only check. Step 4 of the repro (A's own reload drops back to the broadcast) matches your test plan and is consistent with how `socketId` targeting already behaves.

One suggestion left inline about pulling the "is this client-scoped?" test into one shared helper, since the three copies drifting is what caused this in the first place. Not a blocker.

nodes/store/data.js
n-lark: Good call out - added utils with `normalizeClientId()`.

knolleary (approved):

knolleary:
Approved based on the review from @cstns and a review of the delta applied since.

Issue #2239, opened 2026-08-27
Nodes write, presentation reads


n-lark:
### Story 3 — Nodes write, presentation reads
 
> As someone opening a dashboard, I want the page to render from stored state rather than replayed messages, so that what I see matches what everyone else sees.
 
**Closes** [#939](https://github.com/FlowFuse/node-red-dashboard/issues/939), [#428](https://github.com/FlowFuse/node-red-dashboard/issues/428), [#913](https://github.com/FlowFuse/node-red-dashboard/issues/913), [#1251](https://github.com/FlowFuse/node-red-dashboard/issues/1251)
 
- Widget nodes write current values into the layer on input.
- The front end reads from the layer and syncs by delta to a per-client copy; widgets render from that copy.
- This replaces replay-on-connect as the way a page gets its state.
- Existing dashboards keep working with no changes.

cstns:
Something that came up reviewing the stack (#2229, #2230, #2233, #2246) which might be worth folding into the "settle the value model" criterion here.

The same class of thing keeps turning up in different places:

- the `$` view handed out the live `Entry`, so writes bypassed `onChange` and clone-on-write. Fixed for `value` in #2230 with the frozen projection, then came back for `msg` in #2246, because the freeze is shallow.
- there are two write paths into an entry now. `SET_ENTRY` sets value and msg, the plain set path sets only value, so the two drift.
- the store clones `value` itself but trusts the caller to clone `msg`.
- `isReactable` gates both deep-watching and rehydration, and returns false for anything a function node creates.

Each has been an easy fix on its own, so this isn't a complaint about any of them. It's more that the criterion above is currently about which *fields* an entry holds, and every one of these has been about the *rules*: what gets cloned, what's genuinely read-only, what every write path has to set, and what survives serialisation.

Might be worth writing those down in one place before this story starts, so new write paths (and #2240's key binding) have something to check themselves against instead of each one getting caught in review. Only a suggestion, happy to be wrong about how much it's worth.

Issue #2238, opened 2026-09-17
Charts in the data store


n-lark:
# Charts in the data store

## Problem

A chart stores history rather than a value: many messages build the series, bounded by a time window (default one hour) or a point cap (default none).

Once the window is full, each arriving point ages one out, so **trimming runs on every message from then on**. A chart is free for its first hour and pays on every message after, which means it looks fine in any test shorter than the window.

Trimming the legacy datastore is cheap because it only reshuffles pointers:

```js
data[node.id] = currentData.filter(filterFn)
```

The surviving messages are the same objects. Nothing is copied.

The store cannot do that. Its contract is that stored values are insulated from the flows that wrote them, so writing a series has to project the points, deep-clone them, and wrap each one in a Proxy — the same logical operation at 50x the cost.

| per message, 3,600-point series | |
| --- | --- |
| append (`push`) | 0.001ms, 1 event |
| trim — legacy datastore | 0.034ms |
| trim — into the store | 1.73ms |

3,600 points is the default config at 1 msg/sec.

Mutating in place rather than rebuilding is not a way out: `shift()` reindexes a proxied array, so every element hits the set trap — **1.66ms and 1,001 events**.

## Options

Either trim less often and let the series briefly exceed its window, or give the store explicit operations for series.

**1. Batch the trim.** Rebuild every N messages rather than every message. At 1-in-60 that is 0.03ms/msg, level with the legacy path, and needs no new API. Each trim still sends the full series over the socket.

**2. Explicit series ops.** `append` / `trim` / `replace` on the store, each emitting one structured event ("k removed from the front"). Fixes the socket payload as well as the CPU. Adds internal store API.

Neither is a breaking change. `datastore.append` / `save` / `filter` keep their signatures and only their internals change, so chart nodes, third-party widgets and user flows are untouched.

Issue #2237, opened 2026-08-10
Persistence on restart (stretch goal)


n-lark:
### Story 4 — Persistence on restart stretch goal
 
> As someone monitoring production, I want a restart not to wipe what my dashboard was showing, so that a deploy doesn't cost me the trend I was watching.
 
**Open — not designed in refinement.** Everything above lives in memory, so refresh and reconnect work regardless, but a restart clears it. Either Dashboard writes its own snapshot and rehydrates on startup, or it doesn't and restart wipes the store.
 
**Would close** [#1362](https://github.com/FlowFuse/node-red-dashboard/issues/1362), [#1352](https://github.com/FlowFuse/node-red-dashboard/issues/1352), [#753](https://github.com/FlowFuse/node-red-dashboard/issues/753). Without it, those stay open.

cstns:
was the historian feature missed?

n-lark:
> was the historian feature missed?

We should discuss when we refine this @cstns

colinl:
I think there are definitely use cases where re-populating on a node-red restart would be undesirable, so this will need consideration.

PR #2246, opened 2026-09-21
Dashboard data store - storing entire msg 

n-lark:
## Description

We need the whole message in the new datastore, not just `msg.payload`, since multiple nodes read whole messages back off the legacy datastore which will be killed in #2239.

What we are doing here is storing the whole message next to the value, so you can reference it like:

```
$<key>.msg
```

Which will return like:

```
{ payload: 43, topic: 'boiler' }
```

## Related Issue(s)

Relates to #2239 + #2238 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


nodes/store/reactive.js
```diff
@@ -72,7 +104,7 @@ function createDataStore ({ maxHistory = 5, onChange, clone = deepClone, now = D
             }
             if (prop.startsWith('$')) {
                 const rec = target[prop.slice(1)]
-                return rec && Object.freeze({ value: rec.value, quality: rec.quality, timestamp: rec.timestamp, history: Object.freeze(rec.history.slice()) })
+                return rec && Object.freeze({ value: rec.value, msg: rec.msg, quality: rec.quality, timestamp: rec.timestamp, history: Object.freeze(rec.history.slice()) })
```
cstns, line 107: `Object.freeze` is shallow, so `msg` goes out live and the read-only view isn't read-only any more:

```js
s[SET_ENTRY]('w1', 42, { payload: 42, topic: 'boiler' })
s.$w1.msg.payload = 'TAMPERED'
// stored msg is now TAMPERED, onChange fired: []
```

Same shape as the `value` one you fixed in #2230, just via `msg`. It bites harder here because #2239 wants to serve `entry.msg` to widgets, so this is a second write path into the thing that drives rendering. `value` is fine as-is since it's the reactive proxy and writes through it do notify.

nodes/store/reactive.js
```diff
@@ -83,25 +115,7 @@ function createDataStore ({ maxHistory = 5, onChange, clone = deepClone, now = D
...
-                // a replaced value keeps its proxies alive; they must not report into the record any more
-                if (rec.value !== wrapped) return
-                rec.timestamp = now()
-                emit(prop, rec, path)
-            }
-            // clone on write so a flow reusing its own object can't mutate stored state
-            const wrapped = deepReactive(cloneValue(value), notify, prop, cloneValue)
-            rec.value = wrapped
-            rec.timestamp = now()
+            const rec = writeValue(prop, value)
```
cstns, line 118: The plain set path writes value but leaves `msg` untouched, so the two drift:

```
widget write (SET_ENTRY) : value = 42 | msg = {"payload":42,"topic":"boiler"}
flow write  (store.w1=99): value = 99 | msg = {"payload":42,"topic":"boiler"}
```

`global.set('dashboardStore.<id>', 99)` is a reasonable thing for someone to do, and right now it leaves a msg describing the old value. Once widgets render from msg that's a visible contradiction. Clearing `rec.msg` here, or synthesising `{payload: value}`, would at least keep them consistent.

nodes/store/data.js
```diff
@@ -23,11 +23,12 @@ function disableStore () {
...
 }
 
-function writeToStore (node, msg) {
+function writeToStore (node, msg, stored) {
     if (!storeEnabled) return
     if (!msg || typeof msg !== 'object' || !('payload' in msg)) return
     try {
-        getOrCreateStore(node.context().global)[node.id] = msg.payload
+        const clone = config.RED.util.cloneMessage
+        getOrCreateStore(node.context().global)[SET_ENTRY](node.id, msg.payload, clone(stored))
```
cstns, line 31: `stored` is `data[node.id]`, which already contains the payload, and `writeValue` clones `msg.payload` again. With the pre-existing `stripMsg` clone that's three full clones per message. Benchmarked `datastore.save` against #2233:

```
scalar payload          0.0022 -> 0.0026 ms/msg
table payload, 500 rows 1.1977 -> 2.3812 ms/msg
```

Scalars are noise, but the table path roughly doubles. Given #2238 is already sweating 1.73ms on chart trimming, might be worth having the value and the msg share one clone, since the value is a subset of the msg anyway.

nodes/store/reactive.js
```diff
@@ -1,9 +1,11 @@
 const STORE = Symbol.for('@flowfuse/node-red-dashboard/store')
+const SET_ENTRY = Symbol.for('@flowfuse/node-red-dashboard/setEntry')
 const NAMESPACE = 'dashboardStore'
 
 class Entry {
     constructor () {
         this.value = undefined
+        this.msg = undefined
```
cstns, line 8: Worth noting `msg` doesn't survive a restart. `toJSON` flattens to values, and rehydration comes back through the plain set path, so:

```
snapshot a persistent store would write : {"w1":42}
after restart + rehydrate               : value = 42 | msg = undefined
```

Fine today. Relevant to #2237, and to #2239 if msg is what widgets render from, since a restart would leave them with values and no msg.

nodes/store/reactive.js
```diff
@@ -64,6 +66,36 @@ function createDataStore ({ maxHistory = 5, onChange, clone = deepClone, now = D
...
+        // clone on write so a flow reusing its own object can't mutate stored state
+        const wrapped = deepReactive(cloneValue(value), notify, prop, cloneValue)
+        rec.value = wrapped
+        rec.timestamp = now()
+        return rec
+    }
+
+    records[SET_ENTRY] = (prop, value, msg) => {
+        const rec = writeValue(prop, value)
+        rec.msg = msg
```
cstns, line 95: Small asymmetry: the store clones `value` itself but trusts the caller for `msg`. `writeToStore` does clone before calling, so it's correct today, but `SET_ENTRY` is exported through `Symbol.for` so anything can call it, and the store's clone-on-write contract doesn't hold for this half of the entry.

nodes/store/reactive.js
```diff
@@ -64,6 +66,36 @@ function createDataStore ({ maxHistory = 5, onChange, clone = deepClone, now = D
...
 
+    // returns the record without notifying, so a caller can finish writing it before onChange fires
+    const writeValue = (prop, value) => {
+        let rec = records[prop]
+        if (!rec) {
+            rec = new Entry()
+            records[prop] = rec
+        } else if (!Array.isArray(rec.value)) {
+            // snapshotting a whole array on every write (e.g. a table's rows) would blow up memory
+            rec.history.push(Object.freeze({ value: cloneValue(rec.value), timestamp: rec.timestamp }))
```
cstns, line 77: History still captures value only, so `$key.history[0]` and `$key` describe different things now that an entry is value plus msg. Probably fine while nothing reads history, just flagging it so the two don't diverge quietly.

nodes/store/data.js
```diff
@@ -124,7 +125,7 @@ const setters = {
                     ...data[node.id],
                     ...newMsg
                 }
-                writeToStore(node, msg)
+                writeToStore(node, msg, data[node.id])
```
cstns, line 128: Two small things about what ends up in `msg`, both matching the legacy datastore so possibly deliberate.

It's the accumulated merge rather than the message that arrived, so an earlier `topic` sticks around:

```
{payload:1, topic:'boiler'} -> {"payload":1,"topic":"boiler"}
{payload:2}                 -> {"payload":2,"topic":"boiler"}
```

And `_msgid` comes along too. Both are exactly what the legacy store does, which is presumably the point for #2239, but the description's `{ payload: 43, topic: 'boiler' }` reads like a single message. Worth being explicit while the value model is still being settled.

cstns (changes requested):
`$<key>.msg` works as described, the `writeValue`/`SET_ENTRY` split gets you a single emit after both halves are written, and `toJSON` keeps the sidebar clean. 296 tests pass, lint clean.

Requesting changes on two, both because this PR is specifically laying the groundwork for #2239 rather than because anything is user-visible today.

`$key.msg` is writable with no event and no clone, since the frozen projection is shallow. That's the same bug you fixed for `value` in #2230, and it lands on exactly the field #2239 wants to serve to widgets.

And value and msg drift: `SET_ENTRY` writes both, the plain set path writes only value, so `global.set('dashboardStore.<id>', 99)` leaves a msg describing the old value.

Also worth a look: storing the whole msg roughly doubles `save` for large payloads (500-row table goes 1.20 to 2.38 ms/msg) because the payload gets cloned twice, and msg doesn't survive a restart since `toJSON` only flattens values.

nodes/store/reactive.js
n-lark: Valid - added a `deepFreeze` to the history(had same issue) and the msg property with specs.

nodes/store/reactive.js
n-lark: Fixed - went with synthesizing `{payload: value}`

nodes/store/data.js
n-lark: Yahh valid. Reworked `SET_ENTRY` to take the `msg` and pull the value off it, so one clone covers both halves.

Re-benched against #2233: 

```
scalar payload          0.0026 -> 0.0045 ms/msg
table payload, 500 rows 0.5248 -> 0.5329 ms/msg
```

Clones per save:

```
scalar payload          1 -> 2
object payload          2 -> 2   (3 -> 3 once history kicks in)
```

Side effect is the clone moved inside the store, so your point about `SET_ENTRY` trusting its caller is covered by the same change.

nodes/store/reactive.js
n-lark: Covered this under https://github.com/FlowFuse/node-red-dashboard/pull/2246#discussion_r4086958188

nodes/store/reactive.js
n-lark: This improved with the drift fix, since rehydrate comes back through the path that synthesizes msg now:

```
snapshot a persistent store would write : {"w1":42}
after restart + rehydrate               : value = 42 | msg = {"payload":42}
```

So #2239 won't get widgets with a missing `msg`. What still doesn't survive is the rest of the message will leave that to #2237 if we get there.

nodes/store/data.js
n-lark: Yahh so this is deliberate because we need to have the same storage as the legacy data store because we might be keeping replay via `widget-load` as a fallback for third party widgets. Updated the description of the PR.

